PR #2, opened 2018-06-12
Initial migration of code from WF Swarm to SmallRye for the MP OpenAPI impl

EricWittmann:


AndyGee:
Not really possible to review an initial commit of this size, but here's my +1

PR #5, opened 2018-06-18
Updated the version to 1.3.0 to match the MP version

EricWittmann:

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -23,7 +23,7 @@
     <parent>
         <groupId>io.smallrye</groupId>
         <artifactId>smallrye-open-api-parent</artifactId>
-        <version>1.0.0-SNAPSHOT</version>
+        <version>1.3.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>smallrye-open-api-docs</artifactId>
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -23,7 +23,7 @@
     <parent>
         <groupId>io.smallrye</groupId>
         <artifactId>smallrye-open-api-parent</artifactId>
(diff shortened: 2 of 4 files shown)


kenfinnigan:
@EricWittmann, I think I gave you slightly obtuse information the other day!

The version this needs to be is the MP Open API spec version, so I think that would be 1.0.0 for current implementation?

EricWittmann:
OK that actually makes more sense.  The MP+OAI spec is currently at 1.0.1.

EricWittmann:
Well I guess the spec itself is version 1.0.  Is that right, @arthurdm ?

EricWittmann:
My problem with pinning the smallrye version to the spec version that tightly (at the patch level) is that it leaves little room for the impl to evolve.  So imagine if we release smallrye-open-api 1.0.1 and then we find a bug (hard as that might be to believe), what version do we release?  I suppose we can go with a 4th digit:  1.0.1.1 in that case.

Thoughts @kenfinnigan @arthurdm @msavy @AndyGee ?

kenfinnigan:
My initial ideas around version alignment were only at the major and minor level.

Whether a spec is at 1.0.0 or 1.0.1 doesn't impact the version that this impl will be released under.

For instance, 1.0.0 of a spec could be released and 1.0.0 of SmallRye shortly after. Then there could be a few patch releases that fix implementation issues, so we're now at 1.0.3. If 1.0.1 of the spec is then released, it would push the release of the implementation to 1.0.4.

It's only necessary to align impl and spec at the major/minor level, patch is inconsequential from a spec perspective, and required from an impl perspective to provide room for iterations and fixes

EricWittmann:
That sounds reasonable to me. Thanks @kenfinnigan

PR #9, opened 2018-06-25
Added a FilteredIndexView class to filter the contents of a ShrinkWrap IndexView

EricWittmann:


EricWittmann:
Is it really complaining?  Because those classes are definitely referenced.

msavy:
Seems to be a difference between IntelliJ and Eclipse. IntelliJ claims the file/variables are unused (it's not able to know that the file is processed on the bytecode level by Jandex; I guess it thinks it's just kept as a reference and never actually used?).

image

EricWittmann:
Perhaps we should try using the class in a test somewhere just to have a reference to the class and its fields.  Rather than marking it as unused...

PR #16, opened 2018-07-24
Introduce support for @JsonIgnore, @JsonIgnoreProperties, @JsonIgnoreType

### LINKED ISSUE #13: Processing of classes for JSON schema generation should obey @JsonIgnore
When processing a class for the purpose of creating a JSON schema, the class MAY be annotated with `@JsonIgnore`.  We should do our best to support the semantics of that annotation.

msavy:
To facilitate this improvement cleanly, and make future changes and maintenance much easier, I did a substantial refactoring and modularisation of `OpenApiDataObjectScanner`.

`@JsonIgnore`, `@JsonIgnoreProperties`, and `@JsonIgnoreType` should all function now -- hopefully in the same way as in Jackson itself. 

I also took this opportunity to fix some bugs that were outstanding, such as cycle detection for certain nested generic types (cycle detection for nested generics was being skipped previously, but would break if it had been used).

I have a few small aspects to review with Eric before this is finalised.

Fixes #13

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
@@ -15,32 +15,29 @@
  */
 package io.smallrye.openapi.runtime.scanner;
 
-import java.lang.reflect.Modifier;
-import java.util.ArrayDeque;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.Map;
-
-import javax.validation.constraints.NotNull;
-
+import io.smallrye.openapi.api.models.media.SchemaImpl;
+import io.smallrye.openapi.runtime.scanner.dataobject.AnnotationTargetProcessor;
+import io.smallrye.openapi.runtime.scanner.dataobject.DataObjectDeque;
+import io.smallrye.openapi.runtime.scanner.dataobject.IgnoreResolver;
+import io.smallrye.openapi.runtime.scanner.dataobject.TypeResolver;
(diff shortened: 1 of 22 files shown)


implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/WrappedIndexView.java
```diff
@@ -0,0 +1,85 @@
...
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.IndexView;
+import org.jboss.jandex.Type;
+
+import javax.validation.constraints.NotNull;
+import java.util.Collection;
+
+/**
+ * @author Marc Savy {@literal <marc@rhymewithgravy.com>}
```
EricWittmann: Suggest you add some javadoc here to explain the purpose of this class.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/WrappedIndexView.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public class WrappedIndexView implements IndexView {
```
EricWittmann: Perhaps a better name for this class?  Yes it's wrapping an `IndexView` - but **why** is it wrapping it?  I have to open it up and look to know that.  Maybe `EnhancedIndexView`?  That's better though probably not much. :)

msavy:
Thank you for your feedback. I’ll change those bits. 

Naming stuff is hard!

msavy:
@EricWittmann Done. I went with AugmentedIndexView. Not great either, but...

PR #20, opened 2018-08-22
Minimize the use of commons-beanutils

kenfinnigan:
Minimize the use of commons-beanutils and alter one remaining usage to use direct instantiation.


msavy:
Why do you need to minimise the use of `commons-beanutils`? Are you trying to remove it as a dep entirely?

PR #21, opened 2018-08-24
Paths are duplicated if @ApplicationPath contains more than "/"

michbeck100:
With a static file like 

```yaml
openapi: 3.0.1
servers:
- url: http://localhost:8080/api
paths:
  /customers:
```
and a JAXRS Application like 
```java
@ApplicationPath("/api")
public class RestApplication extends Application { 
}
```
The generated document at `/openapi` contains 

```yaml
---
openapi: 3.0.1
servers:
- url: http://localhost:8080/api
paths:
  /customers:
    get:
      ...
  /api/customers:
    get:
      ...
```
This is because the `OpenApiAnnotationScanner` uses the value of `@ApplicationPath` for creating the paths in the OpenAPI document. 
This PR fixes this by ignoring the application path completely during annotation scanning.

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
@@ -134,7 +134,6 @@ public class OpenApiAnnotationScanner {
 
     private OpenAPIImpl oai;
 
-    private String currentAppPath = "";
     private String currentResourcePath = "";
     private String[] currentConsumes;
     private String[] currentProduces;
@@ -203,18 +202,6 @@ public class OpenApiAnnotationScanner {
     private OpenAPIImpl jaxRsApplicationToOpenApi(ClassInfo applicationClass) {
         OpenAPIImpl oai = new OpenAPIImpl();
         oai.setOpenapi(OpenApiConstants.OPEN_API_VERSION);
-
-        // Get the @ApplicationPath info and save it for later (also support @Path which seems nonstandard but common).
-        ////////////////////////////////////////
-        AnnotationInstance appPathAnno = JandexUtil.getClassAnnotation(applicationClass, OpenApiConstants.DOTNAME_APPLICATION_PATH);
-        if (appPathAnno == null) {
-            appPathAnno = JandexUtil.getClassAnnotation(applicationClass, OpenApiConstants.DOTNAME_PATH);
-        }
-        if (appPathAnno != null) {
-            this.currentAppPath = appPathAnno.value().asString();
-        } else {
-            this.currentAppPath = "/";
-        }
 
         // Get the @OpenAPIDefinition annotation and process it.
         ////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


EricWittmann:
If we drop the application path from the annotation scanner, then I think we need to have a way to include that information (in `@ApplicationPath`) in some way.  Your example has it already in the `servers` from the static file.  I think we'll need to add a server entry either during annotation scanning or possibly when serving the `/openapi` file.  

Thoughts @msavy ?

michbeck100:
Maybe the ApplicationPath could be used [here](https://github.com/smallrye/smallrye-open-api/blob/master/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java#L244) as fallback if no server object is found. But then any existing server from the static file might get overwritten during merge.

michbeck100:
Not sure if i understand you correct, but since the annotation model and the static file model are merged to create the complete openapi document, they are not mutually exclusive things.

EricWittmann:
I guess the question boils down to this:  where should the information in `@ApplicationPath` be set when annotation scanning?

That's obviously important information and we can't simply ignore it.  The problem is where to put it.

I think the argument here is that the information should be put in the `servers` section of the document.  And I think I agree.  Unfortunately, that's the **only** part of a server template that can be determined when performing annotation scanning.  The protocol, host, and port cannot be figured out.

So I wonder if what we need to do is perhaps put the info in an extension property so that it can be used later, when some user **asks* for the document using the `/openapi` endpoint.  At that point we should know the rest of the info (protocol, host, port, context path) and we can construct a proper server.

Thoughts?

kenfinnigan:
What's the current status of this?

EricWittmann:
I don't think a consensus was ever reached - so not sure what the best thing to do might be.

MikeEdgar:
My opinion is that at most this change would need to be off by default. I think that there needs to be an understanding by the application developer about how the scanner generally works and then use the static file to fill in the gaps. Specifying the same thing with minor variations will be a problem and I think this is only one example of that.

MikeEdgar:
I was thinking about this issue in terms of supporting scenarios where a service contains multiple `Application` classes. There is currently no way for the scanner to know which resource classes are applicable to which `Application`s since that information is contained in code (`getClasses` and `getSingletons`).

A long-term solution would probably include support in MP-OAI somehow, but suppose this could be supported with an extension property:
`mp.openapi.extensions.application-path.<fully qualified class name, package name, or ALL>=<comma delimited list of paths or NONE>`

This would allow an application to have a resource class be included under one or more `@ApplicationPath`s and also support ignoring an application path altogether when using NONE.

Thoughts?

EricWittmann:
What does it mean to have multiple Application classes in a Microprofile app?  Would that allow mounting different jax-rs apps at different contexts?  If so, I think this makes sense.  However, that seems to imply that the information from `@ApplicationPath` would need to be put in the paths themselves, and **not** in a Server as has been suggested.  Correct?

kenfinnigan:
The spirit of MicroProfile is that there is only ever a single `Application` and `@ApplicationPath` present in a single code base

MikeEdgar:
> What does it mean to have multiple Application classes in a Microprofile app? Would that allow mounting different jax-rs apps at different contexts? If so, I think this makes sense. However, that seems to imply that the information from `@ApplicationPath` would need to be put in the paths themselves, and **not** in a Server as has been suggested. Correct?

Correct - my earlier comment regarding developers not putting conflicting information in the static file and annotations is still applies. The config option I suggested is only tangential to this PR. If such an option were added, my thought is that it could provide a way to ignore `@ApplicationPath`s as requested in this PR. The "normal" use for the option would be to provide a way to map resources to a parent `Application` namespace.

> The spirit of MicroProfile is that there is only ever a single `Application` and `@ApplicationPath` present in a single code base

I absolutely agree with that concept and would suggest a code base with multiple `Application`s should probably be split. That being said, it is something supported by JAX-RS and is almost guaranteed to produce incorrect results from annotation scanning. The fact that it has not yet some up in discussion does show that it is probably not very common for MP users. I checked OpenLiberty's implementation and it basically just logs a warning if more than one `Application` is found.

The bottom line is that there does not seem to be a pressing need to add another custom configuration option or support for it at this time, particularly because it goes against the grain of MicroProfile.

EricWittmann:
Yeah I think it's reasonable to wait and see if this becomes a bigger need in the MP community (multiple apps and how to handle generating OpenAPI for that).  That said, the suggestion of using a config property to provide some level of control is, I think, a good one.  Just not a super high priority IMO.

MikeEdgar:
Closing without merge - this can be revisited in the future if necessary.

PR #22, opened 2018-08-28
THORN-2137: skip module-info.class from scanning

michalszynkiewicz:


msavy:
Quick thought: do we also need `package-info.class`, or is this already ignored?

michalszynkiewicz:
Taking a deeper look at the issue I think it might be more appropriate to fix it by upgrading jandex to 2.0.5.Final in Thorntail.
smallrye-open-api allready uses it and the problem in thorn-2137 is probably caused by the version being overriden by what thorntail uses.
Therefore I'm putting this change on hold.

msavy:
Having a quick look at some recent PRs, they seem to imply that Jandex hides module-info.class from < Java 9, but I guess the issue will come back >=9?

Interested to see how your investigations progress.

https://github.com/wildfly/jandex/pull/33/files

michalszynkiewicz:
@msavy thanks for the link.

I looked through Jandex 2.0.3 code and I didn't find something that would handle package-info.class explicitly but I don't know much about Jandex and package-info.class'es are with us much longer than module-info.class'es so I _think_ they should be okay.

Also, it turns out simply upgrading Jandex to 2.0.5 in Thorntail doesn't fix the problem anyway, the 2.0.3 still sneaks into a user app... 

Taking all this and your comment into account I'd like to proceed with this change.

kenfinnigan:
I updated jandex in thorntail the other day and this issue was fixed. If that’s not the case I’d prefer to fix in thorntail

msavy:
@michalszynkiewicz any idea how the unexpected `2.0.3.Final` Jandex dep is sneaking in? Might be transitively through the test app you're using, perhaps?

michalszynkiewicz:
@kenfinnigan thanks! I checked your changes and everything works fine even without changes here. I'm closing the PR.

michalszynkiewicz:
@msavy I'm not sure, Ken not only changed the dependency version but also added a module for Jandex and it helped.

Issue #27, opened 2018-10-31
Add a toString() to OpenAPIImpl


gastaldi:
I can see the following information in the logs:

```
2018-10-31 14:32:09,326 INFO  [io.smallrye.openapi.api.OpenApiDocument] (ServerService Thread Pool -- 5) OpenAPI document initialized: io.smallrye.openapi.api.models.OpenAPIImpl@15695af8
```

It would be nice if OpenAPIImpl had a toString() to show useful information instead

EricWittmann:
Fixed in https://github.com/smallrye/smallrye-open-api/pull/29

Note: there wasn't much point to the log statement, even with a `toString()` on `OpenAPIDocument`.  It wouldn't have been useful. :)

PR #29, opened 2018-11-01
Added a unit test for simple jax-rs processing

EricWittmann:
Specifically added a test for #28


implementation/src/main/java/io/smallrye/openapi/api/OpenApiDocument.java
```diff
@@ -77,7 +77,6 @@ public OpenAPI get() {
     public void set(OpenAPI model) {
         synchronized (INSTANCE) {
             this.model = model;
-            LOGGER.infov("OpenAPI document set: {0}", model);
```
gastaldi: This change fixes https://github.com/smallrye/smallrye-open-api/issues/27 too

PR #42, opened 2018-11-14
THORN-2190: skip JAX-RS interfaces that don't have concrete impls

michalszynkiewicz:


EricWittmann:
Upon further reflection, I suppose we should really have a unit test for this.  I'll add an issue and address it in the future.

michalszynkiewicz:
maybe it's worth adding it to the TCK?

Issue #44, opened 2018-11-17
Investigate MicroProfile OpenAPI 1.1


jmini:
With the release of [microprofile-open-api](https://github.com/eclipse/microprofile-open-api) planned mid December 2018, I have started to experiment with `1.1-SNAPSHOT` of the API.

You can find the code in the [1.1](https://github.com/jmini/smallrye-open-api/tree/1.1) branch of my fork.
I am not sure it meet all your standard and requirements, but at least it compiles (and the tests are green). 

If you are interested to have this somewhere in this repository, let me know.
Otherwise I can wait until version `1.1` of the API is released and then propose a PR against `master`

EricWittmann:
Hey this is great stuff, thanks!  I was planning on waiting for 1.1 of the spec to be released and then spend a week or so updating SmallRye OpenAPI, so this saves me a ton of work.  Really appreciate the contribution.  Let's go ahead and create a 1.1 branch for SmallRye so we can iterate and collaborate on this stuff.

I'll get that started on Monday...

msavy:
I also had a look at this earlier today and think it looks good — only bit that stood out is the method that’s pending removal, but that should be cleared up soon by upstream. 👍

EricWittmann:
Yeah sorry, by "Let's go ahead" I was referring to myself. :)  I'll create a 1.1 branch tomorrow AM.

jmini:
I have noticed a problem: `ModelConstructionTckTest` is green because none of the tests are run.
I guess this is a TestNG / JUnit issue...

Smallrye is running JUnit Surefire and the TCK Tests are written with TestNG (`@org.testng.annotations.Test`). The tests are ignored.

Have you an idea how to solve this?

---
When this test run, there will be some more work to do.

One thing that was changed with `1.1` and that should fail here: `set/getDefaultValue()` in `APIResponses` is just a convenience for the `APIResponses.DEFAULT` key.

EricWittmann:
In your 1.1 changes you've changed the `ModelContstructionTckTest` so that it no longer extends `BaseTckTest`.  I think this link should work to the relevant portion of the diff:  

https://github.com/smallrye/smallrye-open-api/compare/1.1...jmini:1.1#diff-bda46ad94feb21daec9ce72eb976e373

Try reverting that change.  I was wondering about that change....

EricWittmann:
Also - I have created a `1.1` branch in SmallRye now.

jmini:
[...]

PR #45, opened 2018-11-20
Prepare next version (using MP 1.1-SNAPSHOT)

jmini:
See https://github.com/smallrye/smallrye-open-api/issues/44

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/api/models/ComponentsImpl.java
+++ b/implementation/src/main/java/io/smallrye/openapi/api/models/ComponentsImpl.java
@@ -33,7 +33,7 @@ import org.eclipse.microprofile.openapi.models.security.SecurityScheme;
 /**
  * An implementation of the {@link Components} OpenAPI model interface.
  */
-public class ComponentsImpl extends ExtensibleImpl implements Components, ModelImpl {
+public class ComponentsImpl extends ExtensibleImpl<Components> implements Components, ModelImpl {
 
     private Map<String, Schema> schemas;
     private Map<String, APIResponse> responses;
@@ -83,6 +83,16 @@ public class ComponentsImpl extends ExtensibleImpl implements Components, ModelI
     }
 
     /**
+     * @see org.eclipse.microprofile.openapi.models.Components#removeSchema(java.lang.String)
+     */
+    @Override
+    public void removeSchema(String key) {
+        if (this.schemas != null) {
+            this.schemas.remove(key);
+        }
+    }
+
+    /**
      * @see org.eclipse.microprofile.openapi.models.Components#getResponses()
      */
     @Override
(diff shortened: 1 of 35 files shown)


EricWittmann:
The build is failing with the following errors:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.1:javadoc (default-cli) on project smallrye-open-api: An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - /home/runner/smallrye-open-api/implementation/src/main/java/io/smallrye/openapi/api/models/ExtensibleImpl.java:54: error: unexpected text
[ERROR]      * @see null#removeExtension(java.lang.String)
[ERROR]        ^
[ERROR] /home/runner/smallrye-open-api/implementation/src/main/java/io/smallrye/openapi/api/models/responses/APIResponsesImpl.java:86: error: reference not found
[ERROR]      * @see org.eclipse.microprofile.openapi.models.responses.APIResponses#removeApiResponse(java.lang.String)
[ERROR]             ^
[ERROR] /home/runner/smallrye-open-api/implementation/src/main/java/io/smallrye/openapi/api/models/responses/APIResponsesImpl.java:89: warning - Tag @see: can't find removeApiResponse(java.lang.String) in org.eclipse.microprofile.openapi.models.responses.APIResponses
[ERROR] /home/runner/smallrye-open-api/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/FilteredIndexView.java:47: warning: no description for @param
[ERROR]      * @param delegate
[ERROR]        ^
[ERROR] /home/runner/smallrye-open-api/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/FilteredIndexView.java:48: warning: no description for @param
[ERROR]      * @param config
[ERROR]        ^
[ERROR] 
```

jmini:
Now the error is:

```
[ERROR] Errors: 
[ERROR]   FilterTckTest.initializationError » NullPointer
```

I have it also locally, but I have no idea what changed there and why the test runner can not execute this test.

Can you have a look?

EricWittmann:
Will do!

EricWittmann:
OK I've actually made a number of changes and then merged this PR.  Thanks for this work!  Now we can collaborate on the `1.1` stuff via this branch.  (ping @msavy as an FYI)

Note that the `NullPointer` was due to a problem parsing and modelling the API responses.  The `default` response was being added to the data model even when it was null.  This was putting a `null` value in the map of responses, which was then causing an NPE later in the test.  Fixed by never storing nulls in the API responses map.  I'm not sure what changed on the TCK side to introduce this problem, but it was an obvious fix so I didn't dive deeper.

Also not that I've updated the `pom.xml` files to be more explicit about the dependencies, particularly where we were directly using certain classes but not actually directly depending on the libraries in which they live.  Hopefully I haven't messed anything up (e.g. forgotten a `scope` somewhere).  Happy to have some more eyes looking at this.

Finally, I want @msavy to know that I removed a bunch of `@SuppressWarnings` annotations for unused fields and whatnot from some test classes.  I know we have often have different Eclipse settings for this which has caused one or the other of us to see IDE warnings.  In the future if an IDE complains about unused classes or variables, I suggest we create additional test classes that actually use those variables (but themselves are otherwise not referenced anywhere) rather than suppress the warning.

Issue #46, opened 2018-11-20
Support extensions to JAX-RS


FroMage:
RESTEasy supports additions to JAX-RS, such as:

- `@FormParam`, `@QueryParam` etc… that have an optional parameter name and uses the annotated param/field name
- reactive return types such as RxJava `Single` or `Flowable`

Additionally, I know of applications that do not require `Application` classes to set up jax-rs resources.

It would be great to have extension points in smallrye-open-api to be able to hook into in order to support this behaviour. Swagger has such extension points that I was able to use for that.

FroMage:
For example, see https://github.com/FroMage/redpipe/blob/master/redpipe-engine/src/main/java/net/redpipe/engine/swagger/RxModelConverter.java and https://github.com/FroMage/redpipe/blob/master/redpipe-engine/src/main/java/net/redpipe/engine/core/Server.java#L390

FroMage:
And for custom parameter annotations: https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/ext/OpenAPIExtension.java and https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-jaxrs2/src/main/java/io/swagger/v3/jaxrs2/DefaultParameterExtension.java

FroMage:
I'll try to prototype something and do a PR.

EricWittmann:
Oh that'd be great!

FroMage:
Actually in the case of `@*Param` I can't solve this just with an extension as Jandex does not have the parameter names. I'll see if we can get Jandex to support this.

EricWittmann:
My understanding is that method parameter names are simply not part of the java bytecode.  Admittedly my knowledge in this area is quite stale, so perhaps that has changed.

FroMage:
It is part of the bytecode, with an option to Javac. https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html

MikeEdgar:
@FroMage - are there any other cases not being covered or can this be closed?

MikeEdgar:
Closing - please re-open or open a new issue if anything else needs to be covered.

PR #50, opened 2018-11-30
Added support for CompletionStage and extensions

FroMage:
This is a preview PR for #46

Don't merge it until https://github.com/wildfly/jandex/pull/43 is merged and jandex is released. However, this should support:

- the JAX-RS 2.1 return type of `CompletionStage`
- the ability to register extensions to the scanning process for:
    - custom types such as RESTEasy `@*Param` that use the method parameter name
    - custom async return types such as RESTEasy support for RxJava 1&2 `Single`

Please review so we can get this merged ASAP when Jandex is released. Thanks.

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
@@ -1,0 +1,55 @@
+package io.smallrye.openapi.runtime.scanner;
+
+import org.eclipse.microprofile.openapi.models.parameters.Parameter.In;
+import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.MethodParameterInfo;
+import org.jboss.jandex.Type;
+
+import io.smallrye.openapi.runtime.util.JandexUtil.JaxRsParameterInfo;
+
+/**
+ * Extension point for supporting extensions to JAX-RS.
+ */
+public class AnnotationScannerExtension {
+
+    /**
+     * Determines where an @Parameter can be found (examples include Query,
+     * Path, Header, Cookie, etc).
+     * 
+     * @param paramInfo
+     * @return the parameter location, or null if unknown
+     */
+    public In parameterIn(MethodParameterInfo paramInfo) {
+        return null;
+    }
+
+    /**
+     * Returns jax-rs info about the parameter at the given index. If the index
+     * is invalid or does not refer to a jax-rs parameter then this should
+     * return null. Otherwise it will return a {@link JaxRsParameterInfo} object
+     * with the name and type of the param.
+     * 
+     * @param method
+     *            MethodInfo
+     * @param idx
+     *            index of parameter
+     * @return JaxRsParameterInfo or null if unknown.
+     */
+    public JaxRsParameterInfo getMethodParameterJaxRsInfo(MethodInfo method, int idx) {
+        return null;
+    }
+
+    /**
+     * Unwraps an asynchronous type such as
+     * <code>CompletionStage&lt;X&gt;</code> into its resolved type
+     * <code>X</code>
+     * 
+     * @param type
+     *            the type to unwrap if it is a supported async type
+     * @return the resolved type or null if not supported
+     */
+    public Type resolveAsyncType(Type type) {
+        return null;
+    }
+
+}
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
@@ -1,6 +1,9 @@
 package io.smallrye.openapi.runtime.scanner;
 
+import java.util.Collection;
+
 import org.eclipse.microprofile.openapi.models.parameters.Parameter.In;
+import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.MethodParameterInfo;
 import org.jboss.jandex.Type;
@@ -50,6 +53,14 @@ public class AnnotationScannerExtension {
      */
     public Type resolveAsyncType(Type type) {
         return null;
+    }
+
+    /**
+     * Gives a chance to extensions to process the set of jax-rs application classes.
+     * @param scanner the scanner used for application scanning
+     * @param applications the set of jax-rs application classes
+     */
+    public void processJaxRsApplications(OpenApiAnnotationScanner scanner, Collection<ClassInfo> applications) {
     }
 
 }
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
@@ -13,7 +13,7 @@ import io.smallrye.openapi.runtime.util.JandexUtil.JaxRsParameterInfo;
 /**
  * Extension point for supporting extensions to JAX-RS.
  */
-public class AnnotationScannerExtension {
+public interface AnnotationScannerExtension {
 
     /**
      * Determines where an @Parameter can be found (examples include Query,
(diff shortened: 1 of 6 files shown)


implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -128,6 +131,8 @@
  */
 public class OpenApiAnnotationScanner {
 
+    private static final DotName COMPLETION_STAGE_NAME = DotName.createSimple(CompletionStage.class.getName());
```
EricWittmann: Maybe put this in https://github.com/smallrye/smallrye-open-api/blob/master/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConstants.java

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
```diff
@@ -0,0 +1,55 @@
...
+import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.MethodParameterInfo;
+import org.jboss.jandex.Type;
+
+import io.smallrye.openapi.runtime.util.JandexUtil.JaxRsParameterInfo;
+
+/**
+ * Extension point for supporting extensions to JAX-RS.
+ */
+public class AnnotationScannerExtension {
```
EricWittmann: This is a class rather than an interface?

EricWittmann:
This looks interesting.  Am I right in assuming that it's up to the platform to load the extensions in whatever way makes sense?  So e.g. Thorntail would be responsible for loading the extensions in some Thorntaily way?  And other platforms would do the same.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
FroMage: Done

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
FroMage: Yes, so that extensions don't have to implement every method.

FroMage:
ATM my usage doesn't need this to be a service. It could make sense to also allow for this to be service-loaded. In general I try to not add features before there's a user for them.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
EricWittmann: I wonder if an interface with default methods might better capture the intent.

EricWittmann:
This change looks great to me - with potential for additional extension points in the future.  So a big +1 from me on this.  Just let me know when it's ready to be merged (when the new version of jandex is available).

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
FroMage: Well, yes that would work too. I guess it only depends on your minimum Java version, but if it's fine by you I can change that.

msavy: +1 to interface. AFAIK, (cc @kenfinnigan), Smallrye expects Java >=8.

kenfinnigan: Correct, right now it's Java 8+

EricWittmann: Great, then I think the consensus is interface with default methods.  Otherwise I think this looks good and we just need to wait for jandex before merging.  Thanks for the contribution!

FroMage:
I rebased, added an extension point, depend on latest released jandex and turned the class into an interface, but that makes the implementation really ugly because you can't use `super` to invoke the super class to avoid figuring out what is the right default return value, you need to call `return return AnnotationScannerExtension.super.resolveAsyncType(type);` which IMO is much worse than if we were using classes as initially.

FroMage:
It'd be really great if we could get this reviewed and released ASAP because we need it in another project :)

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Determines where an @Parameter can be found (examples include Query,
+     * Path, Header, Cookie, etc).
+     * 
+     * @param paramInfo
+     * @return the parameter location, or null if unknown
+     */
+    public default In parameterIn(MethodParameterInfo paramInfo) {
+        return null;
```
msavy: I might be missing something, but why are you providing default/null implementations in this interface instead of just a non-default definition and requiring the user to provide a null impl if they really want that.

This might relate to your comment https://github.com/smallrye/smallrye-open-api/pull/50#issuecomment-450897205 in a way I'm not grokking.

FroMage:
@kenfinnigan it's up to you guys. If it weren't for that `super` annoyance, I would not have minded either way, but that difference probably means most extension implementors will never delegate to the superinterface because they won't know how to invoke that super method. I knew it was possible only because I followed the implementation of default methods in Java at the time they were introduced, which is a pretty high bar…

EricWittmann:
If the usage is ugly perhaps an interface without default implementations and instead include a default/abstract/no-op implementation class?  I honestly don't have a strong opinion on this, so I'm fine with whatever the consensus might be. :)

kenfinnigan:
@FroMage if you want to simplify the implementation usage, then I'm good with whatever change that means

FroMage:
OK I just pushed a change to have an interface and default base class implementing that. I still don't see why that's better than a single class, but I don't care at all, for implementors it's still extending a single class and no `super` weirdness.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
FroMage: Because providing a default impl makes it instantly obvious to users what a valid default return value is. This shows that returning `null` is acceptable and will allow other extensions to participate. Otherwise you have to rely on javadoc and have no code to prove it.

FroMage:
Thanks! When do you think we can expect a release? Even a Beta would be good :)

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
msavy: Okay, thanks, I wasn't sure whether an implementor should generally be expected to provide an implementation to all these methods for it to be sensible.

kenfinnigan:
In terms of a release, we may have a slight problem.

Current master has been updated to work against 1.1-RC1 of the spec, and the auto release process we currently have doesn't support releasing anything containing a "Beta" type designation. I think you have admin access to Semaphore so you can take a look and see what I mean

EricWittmann:
@FroMage What sort of timeframe do you need for a release?  We should be releasing MP OpenAPI 1.1 in the next week or so.  Do you need this released sooner than, say, two weeks from now?

FroMage:
Two weeks should be good, but if you can't release with Jandex Beta that's another problem to solve…

kenfinnigan:
It would be better if it wasn't Beta, but not sure what Jason's plans around that are

EricWittmann:
@FroMage and @kenfinnigan - Version 1.1.1 of MP+OAI has been released and I have updated SmallRye OpenAPI to support it.  Everything looks good from my perspective.  I'd be happy if a release were done asap.  Any chance jandex has a Final release out that can be used?

FroMage:
I think @n1hility and @stuartwdouglas can make jandex releases.

FroMage:
Yay, thanks!

kenfinnigan:
Sounds good @EricWittmann. Do you need me to pull trigger or you can?

EricWittmann:
I don't know how.  But if you point me in the right direction I'm happy to do it. :)

kenfinnigan:
Pulled the trigger, so will build shortly

kenfinnigan:
@FroMage, @EricWittmann - https://search.maven.org/search?q=smallrye-open-api

1.1.0 is now in Central

FroMage:
Great, thanks!

PR #61, opened 2018-12-18
Run TCK on random free port.

### LINKED ISSUE #57: Run junit TCK tests on non-8080 port
The tests start up a server on port 8080.  Instead, it should start the server on a random port and then use that.  Otherwise it's very possible to collide with something else running on the machine.

msavy:
Use builder-helper-maven-plugin's reserve-network-port to reserve a
random free port and export to a Maven Property. Maven-surefire-plugin
exports this to the Java System Property `smallrye.openapi.server.port`.

If `smallrye.openapi.server.port` is unset or empty, the test port will
default to 8080.

Fixes #57

Signed-off-by: Marc Savy <marc@rhymewithgravy.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,6 +35,7 @@
 
     <properties>
         <version.asciidoctor.plugin>1.5.7.1</version.asciidoctor.plugin>
+        <version.buildhelper.plugin>3.0.0</version.buildhelper.plugin>
         <version.com.fasterxml.jackson>2.9.8</version.com.fasterxml.jackson>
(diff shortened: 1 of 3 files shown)


msavy:
@EricWittmann should I do this PR against the other branch as well?

EricWittmann:
No that's OK - I think just against `master` is fine.

msavy:
Okay, cool. It would be easy to do, so if you change your mind LMK.

EricWittmann:
Yeah we can always cherry pick the commit, but we'll need to merge `2.x` onto `master` at some point, so I think we should only double up if it's really necessary.

Issue #56, opened 2018-12-17
Investigate MicroProfile OpenAPI 2.0


jmini:
Now that we have a first milestone release of MP-OpenAPI 2.0 ([2.0-MR1](https://github.com/eclipse/microprofile-open-api/releases/tag/mp-openapi-2.0-MR1)), I think it would make sense to create a `2.0` branch in this repo to work on an implementation that is compatible with this version.

I have started to do it to validate some other stuff I am working on. This is interesting because it allows to find some issue/limitation in the API.

Can a `2.0` branch be created in this repo?

This would be similar to #44 (for `2.0` instead of `1.1`)

EricWittmann:
Branch `2.0` created.

kenfinnigan:
There's already a `1.1` branch, would it be better to just move `master` to be 2.x?

EricWittmann:
That's a really great point, considering we likely won't be making changes to the 1.0.x line anymore.

EricWittmann:
Although maybe it makes more sense to create a `2.0` branch and move the `1.1` branch to master...

What do you guys think @jmini and @msavy ?

kenfinnigan:
I don't think we want to maintain multiple branches of active development, so if there's going to be an MP OpenAPI 2.0 soon, then that should be what is developed on master.

EricWittmann:
Side question, @kenfinnigan : Can we do releases of smallrye off any branch?  If yes, then having `master` === 2.0 probably makes sense.

EricWittmann:
Ah OK - so here's the state of things.  We're going to be releasing version `1.1` of the specification this month.  Once that happens, we'll need to update SmallRye to support that version of the spec.  That is the purpose of the `1.1` branch here in SmallRye.  When we do that, `smallrye-open-api` will go from version `1.0.4` to `1.1.0`.

Far off in the distant future of 2019, we will be releasing version `2.0` of the specification (with more extensive changes).  At that time, we'll need to update SmallRye to implement the new spec, and presumably change `smallrye-open-api` to version `2.0.0`.

@jmini has requested a `2.0` branch here in SmallRye so that he can start making some changes to align with what's going on in spec 2.0, and I assume to evaluate those changes in a practical way.  So the 2.0 work is very preliminary, and based off a very early milestone release of the spec.  On the other hand, the 1.1 work is close to release.

kenfinnigan:
[...]

PR #62, opened 2018-12-18
[2.0] initial contribution

jmini:
See #56

---

On my `2.0` branch, I did not manage to update completely to `2.0-MR1`. 

8736af2: Two tests are broken at TCK level, I have commended them out and opened issues to get them fixed.

52a152e: An other big issue is the `MergeUtil` that is now broken, since getters for List and Maps returns immutable collections.
I did not investigate it yet, but now that `Paths`, `Callback`, `Content`, `APIResponses`, `Scopes`, `SecurityRequirement` and  `ServerVariables` are no longer forced to extend maps, maybe this code can be rewritten differently (merging 2 immutable maps is like creating a new one and calling `putAll()` twice - or I need better understanding of what the `MergeUtil` is doing).

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/api/models/ComponentsImpl.java
+++ b/implementation/src/main/java/io/smallrye/openapi/api/models/ComponentsImpl.java
@@ -16,6 +16,7 @@
 
 package io.smallrye.openapi.api.models;
 
+import java.util.Collections;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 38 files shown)


EricWittmann:
Thanks for this PR - I'll get to it as soon as I can.  I can have a look at `MergeUtil` and the broken tests after I review this PR and merge it...

PR #85, opened 2019-04-05
chore: remove dependency on commons-collections and commons-io

gastaldi:


gastaldi:
I removed the formatting changes, it should contain only the change now

Issue #90, opened 2019-04-29
Parameter with @Valid are not take account in schema


Dufgui:
When a JSR303 annotation is used on the RequestBody of REST operation, the paths.{operation}.requestBody is not generated in the swagger documentation

This code was tested with OpenJDK Runtime Environment (build 1.8.0_161-b14) and Quarkus 0.14.0.

```java
@Path("/fruits")
@Produces(MediaType.APPLICATION_JSON)
@Consumes(MediaType.APPLICATION_JSON)
public class FruitResource {

    private Set<Fruit> fruits = Collections.newSetFromMap(Collections.synchronizedMap(new LinkedHashMap<>()));

    public FruitResource() {
        fruits.add(new Fruit("Apple", "Winter fruit"));
        fruits.add(new Fruit("Pineapple", "Tropical fruit"));
    }

    @GET
    public Set<Fruit> list() {
        return fruits;
    }

    @POST
    public Set<Fruit> add(@Valid /* JSR Bean validation*/ Fruit fruit) { 
        fruits.add(fruit);
        return fruits;
    }

    @DELETE
    public Set<Fruit> delete(Fruit fruit) { // no JSR validations
        fruits.remove(fruit);
        return fruits;
    }
}
```
This is the generated OpenAPI spec

---
```yaml
openapi: 3.0.1
info:
  title: Generated API
  version: "1.0"
paths:
  /fruits:
    get:
      responses:
        200:
          description: OK
          content:
            application/json: {}
    post: 
      responses: #request body not generated for post
        200:
          description: OK
          content:
            application/json: {}
    delete:
      requestBody: #request body generated for delete
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Fruit'
      responses:
        200:
          description: OK
          content:
            application/json: {}
components:
  schemas:
    Fruit:
      properties:
        description:
          type: string
        name:
          type: string
```

I see it, it come from 
OpenApiAnnotationScanner.processJaxRsMethod in the following block:
```
        // If the request body is null, figure it out from the parameters.  Only if the
        // method declares that it @Consumes data
        if (operation.getRequestBody() == null && currentConsumes != null) {
            Type requestBodyType = JandexUtil.getRequestBodyParameterClassType(method);
            if (requestBodyType != null) {
                Schema schema = typeToSchema(requestBodyType);
                if (schema != null) {
                    RequestBody requestBody = new RequestBodyImpl();
[...]

PR #98, opened 2019-05-04
Ensure JAX-RS Request Body is properly handled when non JAX-RS annotations present

geoand:
Prior to this fix, the implementation would disregard a proper JAX-RS
request body parameter if it was annotated with any annotation, like for
example `javax.validation.Valid`

Relates to: https://github.com/quarkusio/quarkus/issues/2339

cc @kenfinnigan

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/util/JandexUtil.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/util/JandexUtil.java
@@ -44,6 +44,8 @@ import static java.util.stream.Collectors.toList;
  */
 public class JandexUtil {
 
+    private static final String JAXRS_PACKAGE = "javax.ws.rs";
+
     /**
      * Simple enum to indicate the type of a $ref being read/written.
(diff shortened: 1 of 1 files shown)


kenfinnigan:
Does this also fix #90 ?

geoand:
I hadn't seen that issue at all! It seems exactly the same as the Quarkus one which I tested against with this patch, so I would say yes, it does.

geoand:
Thanks! 

Would it be possible to get a new release so we can include it in Quarkus and close a couple related issues that people are having?

Thanks again

EricWittmann:
We should make sure to get https://github.com/smallrye/smallrye-open-api/pull/97 merged as well - before doing a release.

EricWittmann:
@kenfinnigan I've merged both outstanding PRs - if you can kick off a release.  Should it be a standard release or a snapshot release?  Not sure what we support here.  Doesn't matter to me! :)

kenfinnigan:
I've kicked off a release build

PR #99, opened 2019-05-08
Use @Schema 'name' for props and schema refs; Fix #25

### LINKED ISSUE #25: @Parameter with missing name produces duplicate
Name is missing in the `@Parameter`:

```java
public Response deleteTask(
@Parameter(
            description = "The id of the task",
            example = "e1cb23d0-6cbe-4a29",
            schema = @Schema(
                    type = SchemaType.STRING))
 @PathParam("taskId")
String taskId) {
    /* .. */
}
```
produces:
```yaml
      parameters:
      - in: path
        description: The id of the task
        schema:
          type: string
        example: e1cb23d0-6cbe-4a29
      - name: taskId
        in: path
        required: true
        schema:
          type: string
```

=> Only one parameter should be generated.

---
An other related problem, name is present:

```java
public Response deleteTask(
@Parameter(
            description = "The id of the task",
            name = "taskId",
            example = "e1cb23d0-6cbe-4a29",
            schema = @Schema(
                    type = SchemaType.STRING))
 @PathParam("taskId")
String taskId) {
    /* .. */
}
```

produces:
```yaml
      parameters:
      - name: taskId
        in: path
        description: The id of the task
        schema:
          type: string
        example: e1cb23d0-6cbe-4a29
```

This is not valid because `required : true` is missing. This is mandatory for path parameter in OpenAPI

MikeEdgar:
This is what I have for the change to use the @Schema name (and issue #25). Most of the real change is in the schema registry, but I also consolidated some duplicate functions and moved some that seemed better placed in the SchemaFactory than in the main scanner. As noted in the code comments, this does not handle merging Schema information found in @Components with information found directly on the models. 

I think we might need to have a bigger discussion around how best to merge/resolve things within the scanning phase (not to mention the higher-level phases) and whether or not warnings should be logged if the user codes annotations that are not valid or have undefined behavior. For example, a @Schema with a ref that also includes other properties - a good case to issue warning messages? I like the idea of being told that something in an annotation will be ignored/discarded or is otherwise not valid.

* Consolidate duplicate Schema utility functions into SchemaFactory
* Construct SchemaRegistry with the IndexView
* Use custom key to store schemas in registry for added control of the equals/hashCode methods
* Update JAX-RS annotation scanning to address issue #25

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -119,6 +120,7 @@ import io.smallrye.openapi.runtime.util.JandexUtil;
 import io.smallrye.openapi.runtime.util.JandexUtil.JaxRsParameterInfo;
 import io.smallrye.openapi.runtime.util.JandexUtil.RefType;
 import io.smallrye.openapi.runtime.util.ModelUtil;
+import io.smallrye.openapi.runtime.util.SchemaFactory;
 
 /**
  * Scans a deployment (using the archive and jandex annotation index) for JAX-RS and
@@ -143,8 +145,6 @@ public class OpenApiAnnotationScanner {
     private String currentResourcePath = "";
     private String[] currentConsumes;
     private String[] currentProduces;
-
-    private SchemaRegistry schemaRegistry;
 
     private List<AnnotationScannerExtension> extensions;
 
@@ -180,7 +180,10 @@ public class OpenApiAnnotationScanner {
         // Initialize a new OAI document.  Even if nothing is found, this will be returned.
         oai = new OpenAPIImpl();
         oai.setOpenapi(OpenApiConstants.OPEN_API_VERSION);
-        schemaRegistry = SchemaRegistry.newInstance(config, oai);
+
+        @SuppressWarnings("unused")
+        // Creating a new instance of a registry which will be set on the thread context.
+        SchemaRegistry schemaRegistry = SchemaRegistry.newInstance(config, oai, index);
 
         // Get all jax-rs applications and convert them to OAI models (and merge them into a single one)
         Collection<ClassInfo> applications = this.index.getAllKnownSubclasses(DotName.createSimple(Application.class.getName()));
@@ -516,7 +519,12 @@ public class OpenApiAnnotationScanner {
         // Process @Parameter annotations
         /////////////////////////////////////////
         List<AnnotationInstance> parameterAnnotations = JandexUtil.getRepeatableAnnotation(method,
-                OpenApiConstants.DOTNAME_PARAMETER, OpenApiConstants.DOTNAME_PARAMETERS);
+                                                                                           OpenApiConstants.DOTNAME_PARAMETER,
+                                                                                           OpenApiConstants.DOTNAME_PARAMETERS);
+
+        // Cross reference from the JAX-RS *Param name to the name specified by the Schema
+        Map<String, String> parameterNameOverrides = new HashMap<>();
+
         for (AnnotationInstance annotation : parameterAnnotations) {
             Parameter parameter = readParameter(annotation);
             if (parameter == null) {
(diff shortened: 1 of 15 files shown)


MikeEdgar:
Looks like my last minute edit of a JavaDoc is causing an issue. I'll get that fixed up and update the request.

MikeEdgar:
Those are some good points regarding validation and warnings. I'd be interested in following along with those discussions for additional spec capabilities. It would be ideal if there are some defined scenarios that are invalid, such as the example of using ref along with other properties.

EricWittmann:
Here is a link to some discussion on the topic:  https://github.com/eclipse/microprofile-open-api/issues/339

EricWittmann:
Ha ha ha - according to GH, my comment was made "4 hours from now".  No one ask me for my flux capacitor, I'm not sharing!

(also this is going to be a confusing thread to anyone who looks at it  in the future)

To be clear, [this comment](https://github.com/smallrye/smallrye-open-api/pull/99#issuecomment-490457536) should be the first one in the thread.

MikeEdgar:
> No one ask me for my flux capacitor, I'm not sharing!

I did notice that myself as well :-) I pushed the updated JavaDoc and it looks to be OK now.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -557,13 +581,17 @@ private void processJaxRsMethod(OpenAPIImpl openApi, ClassInfo resource, MethodI
...
                 Type paramType = parameters.get(idx);
                 Parameter parameter = new ParameterImpl();
                 parameter.setName(paramInfo.name);
                 parameter.setIn(paramInfo.in);
-                parameter.setRequired(true);
-                Schema schema = typeToSchema(paramType);
+
+                if (paramInfo.in == In.PATH) {
+                    parameter.setRequired(true);
+                }
```
EricWittmann: Nice addition.

implementation/src/main/java/io/smallrye/openapi/runtime/util/SchemaFactory.java
```diff
@@ -68,59 +111,157 @@ public static Schema readSchema(IndexView index,
         schema.setExternalDocs(readExternalDocs(annotation.value(OpenApiConstants.PROP_EXTERNAL_DOCS)));
         schema.setDeprecated((Boolean) overrides.getOrDefault(OpenApiConstants.PROP_DEPRECATED, JandexUtil.booleanValue(annotation, OpenApiConstants.PROP_DEPRECATED)));
         schema.setType((Schema.SchemaType) overrides.getOrDefault(OpenApiConstants.PROP_TYPE, JandexUtil.enumValue(annotation, OpenApiConstants.PROP_TYPE, Schema.SchemaType.class)));
-        schema.setEnumeration((List<Object>) overrides.getOrDefault(OpenApiConstants.PROP_ENUM, JandexUtil.stringListValue(annotation, OpenApiConstants.PROP_ENUM)));
+        schema.setEnumeration((List<Object>) overrides.getOrDefault(OpenApiConstants.PROP_ENUMERATION, JandexUtil.stringListValue(annotation, OpenApiConstants.PROP_ENUMERATION)));
```
EricWittmann: Good catch - the OpenAPI data model uses `enum` for this, which is probably why this was wrong.  I guess there is/was no test coverage for this. :(

implementation/src/main/java/io/smallrye/openapi/runtime/util/SchemaFactory.java
```diff
@@ -68,59 +111,157 @@ public static Schema readSchema(IndexView index,
...
+     */
+    private static Discriminator readDiscriminatorMappings(AnnotationValue value) {
+        if (value == null) {
+            return null;
+        }
+        LOG.debug("Processing a list of @DiscriminatorMapping annotations.");
+        Discriminator discriminator = new DiscriminatorImpl();
+        AnnotationInstance[] nestedArray = value.asNestedArray();
+        for (@SuppressWarnings("unused") AnnotationInstance nested : nestedArray) {
+            // TODO iterate the discriminator mappings and do something sensible with them! :(
```
EricWittmann: I know, right?

EricWittmann:
@msavy or @kenfinnigan - any comments before I merge this?

implementation/src/main/java/io/smallrye/openapi/runtime/util/SchemaFactory.java
MikeEdgar: This actually caused the TCK to fail with the consolidated version of this method. I think you're right that none of the tests within the implementation project hit this.

PR #112, opened 2019-05-20
Use Bean Validation constraints for schema parameters if not specified

### LINKED ISSUE #103: Supplement schema restrictions using Bean Validation annotations, if present
Add constraint information to field-level schemas when Bean Validation annotations are available and when the information is not already specified in the `@Schema` annotation.

MikeEdgar:
Changes for #103. 

Regarding the unit tests, I've been meaning to ask whether it's intended for the Surefire tests to be effectively disabled based on the 'includes' configuration.

*EDIT* Note that this functionality is enabled by default (no configuration).

*EDIT 2* Fixes #103


MikeEdgar:
Pushed again re-based on master.

PR #120, opened 2019-05-30
Do not set null/empty enumeration from annotation on schema

### LINKED ISSUE #110: Using @Schema disables enum value generation
Annotating an enum field with for example `@Schema(required = true)` stops auto generation of enum value info.
This unfotunately cannot be restored by specifying `@Schema(implementation = MyEnum.class, required = true)`.
There should be a way. Best would be to not having to specify an implementation at all. I think the implementation should in fact be only overriden if you explicitly specify one. If not, the default generation should be used.
My proposal is to change `Class<?> implementation() default Void.class;` to `Class<?> implementation() default Infer.class;`, where `Infer.class` is just an internal control class. You will then have to specify `Void.class` yourself to disable auto generation.

MikeEdgar:
This set of changes prevents the generated/scanned `Enum` value list from being overwritten by an undefined or empty `enumeration` in an `@Schema` annotation.

Fixes #110

### DIFF
--- a/implementation/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/EnumRequiredContainer.java
+++ b/implementation/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/EnumRequiredContainer.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2018 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package test.io.smallrye.openapi.runtime.scanner.entities;
+
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+/**
+ * @author Marc Savy {@literal <marc@rhymewithgravy.com>}
+ */
+public class EnumRequiredContainer {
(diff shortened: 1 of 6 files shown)


implementation/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/EnumRequiredContainer.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package test.io.smallrye.openapi.runtime.scanner.entities;
+
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+/**
+ * @author Marc Savy {@literal <marc@rhymewithgravy.com>}
```
EricWittmann: Nice of you to give away the credit.... or is it the blame?? ;)

MikeEdgar: :-) I noticed this after I already pushed the branch.

MikeEdgar:
> I'm good with this unless @MikeEdgar wants to update anything before I merge?

Unless @msavy does not want to the credit, no further changes.

EricWittmann:
OK great!  Thanks @msavy for this excellent contribution!  :)

MikeEdgar:
Next time I copy/paste I'll be more careful :-)

EricWittmann:
Note for those interested - the semaphoneci build failed on this, but I checked out the changes locally and the build was fine.  The ci build failed with this:

```
Could not resolve dependencies for project io.smallrye:smallrye-open-api-1.1:jar:1.0.0-SNAPSHOT: Could not transfer artifact org.codehaus.groovy:groovy:jar:2.5.6 from/to central (https://repo.maven.apache.org/maven2): GET request of: org/codehaus/groovy/groovy/2.5.6/groovy-2.5.6.jar from central failed: Premature end of Content-Length delimited message body (expected: 5466446; received: 1144848
```

I'm assuming that's just a transient error.

PR #122, opened 2019-06-05
Use type params in generated schema names; process enum as map value

### LINKED ISSUE #121: Include type parameters in generated schema names
There was some discussion on generated schema names in #97. The current implementation in the `SchemaRegistry` uses a "name + 1" approach if a name collision occurs. The collision could occur in two scenarios - 1) if the same class name exists in multiple packages or 2) if the class name is a parameterized type. 

This issue is addressing the second scenario. Consider the unit test `ExpectationWithRefsTests#testComplexNestedGenericsWithRefs` which is rooted on a field of type `Fuzz<KustomPair<Fuzz<String, Date>, ?>, Double>`. The current expected is as follows, note the sequential references to different `Fuzz` types.

```javascript
{
  "components": {
    "schemas": {
      "KustomPair": {
        "required": [
          "bar",
          "foo"
        ],
        "type": "object",
        "properties": {
          "bar": {
            "type": "object"
          },
          "foo": {
            "$ref": "#/components/schemas/Fuzz"
          }
        }
      },
      "Fuzz": {
        "maxLength": 123456,
        "type": "object",
        "properties": {
          "qAgain": {
            "format": "date",
            "type": "string"
          },
          "qAgain3": {
            "format": "date",
            "type": "string"
          },
          "qValue": {
            "format": "date",
            "description": "Ah, Q, my favourite variable!",
            "type": "string"
          },
          "tAgain2": {
            "type": "string"
          },
          "tAgain4": {
            "type": "string"
          },
          "tValue": {
            "type": "string"
          }
        }
      },
      "Fuzz1": {
        "type": "object",
        "properties": {
          "qAgain": {
            "format": "double",
            "type": "number"
          },
          "qAgain3": {
            "format": "double",
            "type": "number"
          },
          "qValue": {
            "format": "double",
            "description": "Ah, Q, my favourite variable!",
            "type": "number"
          },
          "tAgain2": {
            "$ref": "#/components/schemas/KustomPair"
          },
          "tAgain4": {
            "$ref": "#/components/schemas/KustomPair"
          },
          "tValue": {
            "$ref": "#/components/schemas/KustomPair"
          }
        }
      }
    }
  }
}
```

Using type parameters would produce a more predictable result.

```javascript
{
  "components" : {
    "schemas" : {
      "FuzzKustomPairDouble" : {
        "type" : "object",
        "properties" : {
          "qAgain" : {
            "format" : "double",
            "type" : "number"
          },
          "qAgain3" : {
            "format" : "double",
            "type" : "number"
          },
          "qValue" : {
            "format" : "double",
            "description" : "Ah, Q, my favourite variable!",
            "type" : "number"
          },
          "tAgain2" : {
            "$ref" : "#/components/schemas/KustomPairFuzzObject"
          },
          "tAgain4" : {
            "$ref" : "#/components/schemas/KustomPairFuzzObject"
          },
          "tValue" : {
            "$ref" : "#/components/schemas/KustomPairFuzzObject"
          }
        }
      },
      "KustomPairFuzzObject" : {
        "required" : [ "bar", "foo" ],
        "type" : "object",
        "properties" : {
          "bar" : {
            "type" : "object"
          },
          "foo" : {
            "$ref" : "#/components/schemas/FuzzStringDate"
          }
        }
      },
      "FuzzStringDate" : {
        "maxLength" : 123456,
        "type" : "object",
        "properties" : {
          "qAgain" : {
            "format" : "date",
            "type" : "string"
          },
          "qAgain3" : {
            "format" : "date",
            "type" : "string"
          },
          "qValue" : {
            "format" : "date",
            "description" : "Ah, Q, my favourite variable!",
            "type" : "string"
          },
          "tAgain2" : {
            "type" : "string"
          },
          "tAgain4" : {
            "type" : "string"
          },
          "tValue" : {
            "type" : "string"
          }
        }
      }
    }
  }
}
```

This is probably a less common scenario in real world applications, but it would help with unit tests in this project :-)

MikeEdgar:
Fixes #121 

Implementation of default schema naming using type parameters (and wildcard bounds). This also addresses an problem discovered where enumerated values are not generated when an `Enum` is the value of a `Map`.

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/SchemaRegistry.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/SchemaRegistry.java
@@ -14,6 +14,7 @@ import org.eclipse.microprofile.openapi.models.OpenAPI;
 import org.eclipse.microprofile.openapi.models.media.Schema;
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationTarget;
+import org.jboss.jandex.DotName;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.ParameterizedType;
 import org.jboss.jandex.Type;
@@ -232,7 +233,7 @@ public class SchemaRegistry {
             schemaName = JandexUtil.stringValue(schemaAnnotation, OpenApiConstants.PROP_NAME);
         }
 
-        String nameBase = schemaName != null ? schemaName : key.type.name().local();
+        String nameBase = schemaName != null ? schemaName : key.defaultName();
         String name = nameBase;
         int idx = 1;
         while (this.names.contains(name)) {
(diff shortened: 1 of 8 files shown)


EricWittmann:
This is neat - how deep can this naming get?

MikeEdgar:
This naming uses the immediate parameters of a type, so it would only be one level deep. That may be cause for concern as @msavy mentioned in #121. In those cases it should fall back to the sequential naming, but I don't know if there's a test explicitly hitting that scenario (I'll add one when I get a chance).

MikeEdgar:
Re-based on master after merge of #112

MikeEdgar:
Added several test cases and a bit of clean-up in name generation code.

msavy:
Mike, what do you think about filing some issues so we can track the
problem cases and fill in those gaps later

MikeEdgar:
> Mike, what do you think about filing some issues so we can track the problem cases and fill in those gaps later

@msavy - I will definitely do so as they arise. The cases I was addressing initially were only dealing with the unit tests and the fact they were breaking based on order of classes scanned.

Issue #125, opened 2019-06-10
Allow registering custom schema types


michael-schnell:
It should be possible to register your own schema types that will then be used during the annotation scanning process. 

This would allow to provide schema information for external types. Sometimes you don't control types used in your components. This means you cannot add annotations for those types. In such cases it would be nice to use above feature.

Let's assume you want to provide your own schema definition for the UUID type.
```yaml
components:
  schemas:
    UUID:
      type: string
      format: uuid
      pattern: ^[a-f0-9]{8}-?[a-f0-9]{4}-?[1-5][a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}$
      title: UUID
      description: Universally Unique Identifier
      example: de8681db-b4d6-4c47-a428-4b959c1c8e9a
```
Then you will use a reference everywhere for the UUID:
```yaml
MyCommand:
  type: object
  required:
    - command-id
  properties:
    command-id:
      $ref: "#/components/schemas/UUID"
```
To achieve this you could do something like this in your Java code:
```java
Type uuidType = Type.create(DotName.createSimple(UUID.class.getName()), Kind.CLASS);
Schema schema = new SchemaImpl();
schema.setType(Schema.SchemaType.STRING);
schema.setFormat("uuid");
schema.setPattern("^[a-f0-9]{8}-?[a-f0-9]{4}-?[1-5][a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}$");
schema.setTitle("UUID");
schema.setDescription("Universally Unique Identifier");
schema.setExample("de8681db-b4d6-4c47-a428-4b959c1c8e9a");
SchemaRegistry.currentInstance().register(uuidType, schema);
```

Unfortunately this does not work, because the [OpenApiAnnotationScanner](https://github.com/smallrye/smallrye-open-api/blob/master/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java) currently always creates a new [SchemaRegistry](https://github.com/smallrye/smallrye-open-api/blob/master/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/SchemaRegistry.java) instance in the `scan()` method.

There should be some kind of enhancement mechanism that allows setting your own types for the registry used by the scanner, that will then always be used. "Always" means that the definition will never be overwritten by the OpenApiAnnotationScanner.

[...]

PR #126, opened 2019-06-10
Allow registering custom schema types (Issue #125)

michael-schnell:


implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -1872,6 +1875,21 @@ private Object parseExtensionValue(String value) {
...
     }
 
+    
+    private CustomSchemaRegistry getCustomSchemaRegistry() {
+        if (config == null || config.customSchemaRegistryClass() == null) {
+            // Provide default implementation that does nothing
+            return (type) -> {};
+        } else {
+            try {
+                return (CustomSchemaRegistry) Class.forName(config.customSchemaRegistryClass()).newInstance();
```
EricWittmann: I think this is the only thing I'm a bit concerned about.  I think `Class.forName` isn't going to work well in all cases and we're going to need to be able to pass the classloader in (notice how it's done for filter and modelreader).  We'll probably need to pass the CL in similarly.

michael-schnell: OK, will check it

EricWittmann: Context classloader is much better, for sure.  I'm inclined to merge this and if consumers of this library need something more fine grained, we can address it later.

PR #134, opened 2019-06-12
Only add same-doc component paths to keys matching required pattern

### LINKED ISSUE #124: @ExampleObject with http/https URL should not lead to "#/components/" reference
Currently the annotation 
```java
@ExampleObject(ref = "https://bitbucket.org/myaccount/whatever.json")`
```
generates an entry like 
```yaml
$ref: '#/components/examples/https://bitbucket.org/myaccount/whatever.json'
```
A full URL starting with "http://" or "https://" should stay unchanged in the generated open api specification.

MikeEdgar:
Fixes #124

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/util/JandexUtil.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/util/JandexUtil.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import io.smallrye.openapi.api.OpenApiConstants;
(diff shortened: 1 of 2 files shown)


implementation/src/main/java/io/smallrye/openapi/runtime/util/JandexUtil.java
```diff
@@ -45,13 +46,28 @@
...
+        Schema("schemas"),
+        SecurityScheme("securitySchemes"),
+        Callback("callbacks"),
+        Link("links"),
+        Response("responses"),
+        Parameter("parameters"),
+        Example("examples"),
+        RequestBody("requestBodies");
+
+        String componentPath;
```
EricWittmann: Nice change.

PR #138, opened 2019-06-17
[in-progress] Expand parameter scan processing

### LINKED ISSUE #123: PathItemImpl should consider multiple methods for same path
Currently a second method on the same path simply overrides a previously parsed operation. It should be allowed to have the same path multiple times for different content types. Instead of simply setting the new operation, it should be checked wether it is a different content type.

Consider the following (valid) OpenAPI specification with two different content types ("json" and "x-www-form-urlencoded"):

```yaml
  /offices/create:
    post:
      summary: Creates a new office.
      operationId: createOffice
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/CreateOfficeCommand'
          application/x-www-form-urlencoded:
            schema:
              type: object
              required:
                - id
                - name
              properties:
                id:
                  type: string
                name:
                  type: string
```

This will be implemented in the same Java resource with two methods:

The first one allows a 'application/json' body:
```java
@POST
@Path("/offices/create")
@Consumes({ MediaType.APPLICATION_JSON })
@Operation(operationId = "createOffice")

public void createOffice(@RequestBody(required = true, 
content = @Content(schema = @Schema(implementation = CreateOfficeCommand.class)))
final CreateOfficeCommand cmd) { ... }
```

The second one allows 'application/x-www-form-urlencoded' parameters:
```Java
@POST
@Path("/offices/create")
@Consumes("application/x-www-form-urlencoded")
@Operation(operationId = "createOffice")

public void createOffice(@FormParam("id") String id, @FormParam("name") String name) { ... }
```

Parsing the resource using the [OpenApiAnnotationScanner](https://github.com/smallrye/smallrye-open-api/blob/master/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java) currently the second "x-www-form-urlencoded" method definition overrides the first 'application/json' one.

Instead the second method for the same path ("/offices/create") should simply be added as another media type for the same operation.

### LINKED ISSUE #129: Expand parameter scanning beyond resource method arguments
The list below is the full set of locations where we might find applicable parameter annotations to scan. The current implementation covers the first possibility.

Possible locations of `@Parameter` and JAX-RS parameter annotations
1) Resource method arguments
2) Resource method (`@Parameter` only)
3) Resource class fields and bean property methods (setters)
4) Fields and bean property methods of `@BeanParam` objects in locations 1 and 3.
5) Support for `@FormParam` annotation by generating a wrapper requestBody/schema (relates to #123)

It's not clear in the MP OpenAPI spec whether `@Parameter` should be supported on resource class fields and bean properties. It seems reasonable to support them wherever JAX-RS parameter annotations can be found, but that may be debatable.

### LINKED ISSUE #135: @BeanParam parameters are apparently ignored
We just had this report in the Quarkus bugtracker:

**Describe the bug**
The openapi does not support @BeanParam.
If the rest service method uses the @BeanParam this attribute is ignored in the generated openapi.yaml file.

**Expected behavior**
The openapi.yaml contains also the information from the @BeanParam parameter object.

**Actual behavior**
The @BeanParam is ignored and not generated in the openapi.yaml

**To Reproduce**
1. Create rest method with a @BeanParam parameter
2. Download openapi http://localhost:8080/openapi

Could you take a look?

Once it's fixed, we need to upgrade SmallRye OpenAPI in Quarkus: https://github.com/quarkusio/quarkus/issues/2819

MikeEdgar:
This is what I have so far for updated parameter processing. This code probably needs some additional test cases, but I want to submit now to get some discussion going on the approach. Generally, parameters are processed in two phases, class (path params) and method (operation params).

Fixes #123 
Fixes #129 
Fixes #135 

* Support for FormParam
* Support for MatrixParam
* Support for path-level and operation-level parameters

### DIFF
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -101,6 +101,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-core-spi</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.eclipse.microprofile.openapi</groupId>
             <artifactId>microprofile-openapi-tck</artifactId>
             <scope>test</scope>
--- a/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConstants.java
+++ b/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConstants.java
@@ -16,6 +16,10 @@
 
 package io.smallrye.openapi.api;
 
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.concurrent.CompletionStage;
 
 import javax.ws.rs.ApplicationPath;
@@ -23,6 +27,7 @@ import javax.ws.rs.BeanParam;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.CookieParam;
 import javax.ws.rs.DELETE;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.FormParam;
 import javax.ws.rs.GET;
 import javax.ws.rs.HEAD;
@@ -232,6 +237,7 @@ public final class OpenApiConstants {
     public static final DotName DOTNAME_PATH = DotName.createSimple(Path.class.getName());
     public static final DotName DOTNAME_PRODUCES = DotName.createSimple(Produces.class.getName());
     public static final DotName DOTNAME_CONSUMES = DotName.createSimple(Consumes.class.getName());
+
     public static final DotName DOTNAME_QUERY_PARAM = DotName.createSimple(QueryParam.class.getName());
     public static final DotName DOTNAME_FORM_PARAM = DotName.createSimple(FormParam.class.getName());
     public static final DotName DOTNAME_COOKIE_PARAM = DotName.createSimple(CookieParam.class.getName());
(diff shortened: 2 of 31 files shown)


MikeEdgar:
After reviewing the [Quarkus extensions to parameter processing](https://github.com/quarkusio/quarkus/blob/master/extensions/smallrye-openapi/deployment/src/main/java/io/quarkus/smallrye/openapi/deployment/RESTEasyExtension.java), I think it would be a good idea to include run-time support for the [Resteasy parameter annotations](https://github.com/resteasy/Resteasy/tree/master/resteasy-core-spi/src/main/java/org/jboss/resteasy/annotations/jaxrs). With the changes in this PR, some of the methods in `AnnotationScannerExtension` become a bit more awkward and may need to be revisited, if they are necessary at all. In particular, `parameterIn` and `getMethodParameterJaxRsInfo`.

Thoughts?

+@FroMage

FroMage:
You mean directly supporting the RESTEasy param annotations in smallrye-open-api? I'd be OK with that, but then you'd depend on it.

MikeEdgar:
> You mean directly supporting the RESTEasy param annotations in smallrye-open-api? I'd be OK with that, but then you'd depend on it.

The dependency can be avoided using `DotName` constants.

FroMage:
Oh, right. Well that would certainly make our life easier. But we have another reason for those extension points, and that is to make smallrye-open-api aware of the pluggable support for async types in RESTEasy. So either you would have to take that over too, or we'd still need to keep some extension methods for that.

MikeEdgar:
Understood. I think for now only the extension points related to the parameters would be affected. If it turns out they are needed for another reason in the future, we should discuss a new method in the extension interface to align with the parameter changes in this PR.

MikeEdgar:
Squashed in several updates and bug fixes, plus a re-base on master. If there are any test scenarios that it looks like are missing, please let me know and I'll add them.

* Parameter list merge logic, Resteasy annotations
* Remove unused Parameter methods in JandexUtil
* Support references for async types
* Resolved several compiler warnings and Sonarcloud bug

FroMage:
I checked quickly and it looks good, you're testing both without and with the param name value.

PR #146, opened 2019-06-27
Exclude enum fields from values

### LINKED ISSUE #43: Unit Test: JAX-RS interfaces without a concrete impl
See this PR:  https://github.com/smallrye/smallrye-open-api/pull/42

We should add a unit test for this logic.

### LINKED ISSUE #106: Enum fields are included in schema
I am trying to generate schema for an Enum type:
```
public enum JobState {
    DONE,
    FAILED,
    PROCESSING(FAILED, DONE),
    QUEUED(PROCESSING);

    private final JobState[] targetStates;

    JobState(final JobState... targetStates) {
        this.targetStates = targetStates;
    }
    ...
}
```

Unfortunately, currently this results in the `targetStates` field being included in the enumeration of the schema:

```
              state:
                enum:
                - DONE
                - FAILED
                - PROCESSING
                - QUEUED
                - targetStates
                type: string
```

MikeEdgar:
Fixes #106 

* Ignore static enum fields also
* Update affected unit tests
* Sort enum constant names for predictable results
* Add test for JandexUtil#getJaxRsResourceClasses (fixes #43)

With this PR, the results of scanning the enum used in several tests has changed. In the example below, only `Gold` and `Green` will be output in the list of enumerated values.

```java
public enum BazEnum {
    Green,
    Gold;

    int z = 0;

    static BazEnum Orange;
}
```

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeProcessor.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeProcessor.java
@@ -213,27 +213,37 @@ public class TypeProcessor {
                 readEnumeration(valueType, propsSchema);
             } else {
                 propsSchema.type(Schema.SchemaType.OBJECT);
+                pushToStack(valueType, propsSchema);
             }
 
-            pushToStack(valueType, propsSchema);
             propsSchema = SchemaRegistry.checkRegistration(valueType, typeResolver, propsSchema);
         }
 
         return propsSchema;
     }
 
+    /**
+     * Add each Enum constant name to the list of the given schema's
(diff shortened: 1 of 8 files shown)


MikeEdgar:
> 
> 
> Have you double-checked this doesn't break in the eclipse compiler? I remember having problems with that because their representation of enums is slightly different to OpenJDK.

I'm using Eclipse for development and I haven't seen any issues. That's what I'm using to run the unit tests in addition to the command line. From what I can tell, checking for `0x4000` in the flags is a standard approach. See `java.lang.reflect.Field#isEnumConstant()` and https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.5-200-A.1.

PR #148, opened 2019-07-01
Add support for RestEasy multipart request content types

MikeEdgar:
Fixes #144 ; feedback welcome.

* Add JaCoCo aggregation to TCK and parent for full coverage report in SonarCloud
* Treat RestEasy multipart interfaces as object types in scan
* Always retrieve types for `CollectionStandin` and `MapStandin`, even when not indexed by the app/calling code
* Clean up several "smells" reported in SonarCloud
* Update resteasy-core-spi

PR #150, opened 2019-07-12
Support annotated methods for schemas and user-defined prop order

### LINKED ISSUE #87: Don't automatically sort object fields
I wondered why my OpenAPI file under Thorntail didn't look like I expected it to.
Classes under `components.schemas` always have sorted fields instead of their original order in the Java class.
I think it's because you are using ClassInfo from Jandex internally to get [the field info](https://github.com/smallrye/smallrye-open-api/blob/70fe6cb3734082d4d220914b035e88081085732d/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java#L88). Unfortunately ClassInfo does automatic field sorting: https://github.com/wildfly/jandex/blob/61d1115e0a0f15dcfb2b0bbd082100f847963529/src/main/java/org/jboss/jandex/ClassInfo.java#L557

There seems to be no option to disable this behavior. Neither in this implementation, nor in Jandex ClassInfo.
I suggest not to alter the property order or at least make it optional (default: false).
Thanks :)

### LINKED ISSUE #149: OpenApiDataObjectScanner doesn't handle methods
There is already a relevant TODO in the source code, wondering if this is planned to be implemented at some point. I consider it to be a common use case to have to add OpenAPI schema annotation on a getter method (e.g. when dealing with interfaces) and swagger's scanner seems to support it as well.
 
https://github.com/smallrye/smallrye-open-api/blob/a87b278e7abe93eada7028a56c494c848f8496de/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java#L229-L230

MikeEdgar:
Fixes #149
Works around #87

This may require some additional work/test cases/documentation, but I'm opening this request to get some early feedback and discussion. Note this will at least need to be re-based after #148.

- Property order determined by 3rd party standard or widely-used annotations

@neon-dev - if you don't agree that the use of 3rd party annotations addresses your issue #87, I'll remove from this comment that the issue is fixed by this PR.

*Edit*: updated comment for issue #87

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
@@ -15,19 +15,36 @@
  */
 package io.smallrye.openapi.runtime.scanner.dataobject;
 
+import java.lang.reflect.Modifier;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.Deque;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.PriorityQueue;
+import java.util.Queue;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationTarget;
+import org.jboss.jandex.AnnotationTarget.Kind;
+import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
 import org.jboss.jandex.FieldInfo;
+import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.ParameterizedType;
 import org.jboss.jandex.Type;
 import org.jboss.jandex.TypeVariable;
 import org.jboss.logging.Logger;
 
+import io.smallrye.openapi.api.OpenApiConstants;
 import io.smallrye.openapi.runtime.util.TypeUtil;
 
 /**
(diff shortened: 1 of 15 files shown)


neon-dev:
@MikeEdgar thanks. As you already stated it's a workaround and I also see it like that. Since it doesn't fully solve #87 I'd argue to not include it in the comment. It's a nice feature that can help in the meantime, but it's unpractical and redundant if we'd always have to manually specify property definition order on every class.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
```diff
@@ -76,37 +203,201 @@ public Type getResolvedType(Type fieldType) {
...
+            properties.put(propertyName, resolver);
+        }
+    }
+
+    private static void scanMethod(Map<String, TypeResolver> properties, MethodInfo method, Deque<Map<String, Type>> stack) {
+        String methodName = method.name();
+        Type returnType = method.returnType();
+        Type propertyType;
+
+        if (methodName.startsWith("get") &&
```
atzoum: The javabeans convention for booleans is different so better include this as well, e.g.
`if (methodName.startsWith("get") || methodName.startsWith("is"))`
The substring(4) trick will no longer work as well :)

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
```diff
@@ -76,37 +203,201 @@ public Type getResolvedType(Type fieldType) {
...
+            updateTypeResolvers(properties, stack, method, propertyType);
+        }
+    }
+
+    private static void updateTypeResolvers(Map<String, TypeResolver> properties,
+            Deque<Map<String, Type>> stack,
+            MethodInfo method,
+            Type propertyType) {
+        String methodName = method.name();
+        String propertyName = Character.toLowerCase(methodName.charAt(3)) + methodName.substring(4);
```
atzoum: property names can be customized through `javax.json.bind.annotation.JsonbProperty` or `com.fasterxml.jackson.annotation.JsonProperty` annotations

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
```diff
@@ -76,37 +203,201 @@ public Type getResolvedType(Type fieldType) {
...
+        do {
+            chain.put(klazz, type);
+        } while ((type = klazz.superClassType()) != null && (klazz = index.getClass(type)) != null);
+
+        return chain;
+    }
 
-            currentType = currentClass.superClassType();
+    private static void scanField(Map<String, TypeResolver> properties, FieldInfo field, Deque<Map<String, Type>> stack) {
+        String propertyName = field.name();
```
atzoum: same here, a property's name can be customized through an annotation

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
MikeEdgar: Good catch, I overlooked that.

MikeEdgar:
>     2. Property names can be customized if a method or field is annotated with a `JsonProperty` or `JsonbProperty` annotation

I believe the names given in the property order should correspond to the property name before any customization. That is the case for `JsonbPropertyOrder` and `XmlType#propOrder`. Do you know if Jackson behaves differently? I haven't done any experimentation with it.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
neon-dev: Also `javax.xml.bind.annotation.XmlAttribute` / `javax.xml.bind.annotation.XmlElement` because Jackson also recognizes those annotations. Or just one of them, not sure right now.

atzoum:
Jackson works both ways, it is smart enough to recognise both default and custom names. See the test below 

```
package jackson;

import static org.junit.jupiter.api.Assertions.assertEquals;

import org.junit.jupiter.api.Test;

import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.annotation.JsonPropertyOrder;

public class JacksonPropertyOrderTest {

	@Test
	public void test() throws Exception {
		
		TestSchema1 testSchema1 = new TestSchema1();
		testSchema1.name = "name";
		testSchema1.comment = "comment";
		String json1 = new ObjectMapper().writeValueAsString(testSchema1);
		assertEquals("{\"theName\":\"name\",\"comment\":\"comment\"}", json1);
		
		TestSchema2 testSchema2 = new TestSchema2();
		testSchema2.name = "name";
		testSchema2.comment = "comment";
		String json2 = new ObjectMapper().writeValueAsString(testSchema2);
		assertEquals("{\"theName\":\"name\",\"comment\":\"comment\"}", json2);
		
		
	}
	
	@JsonPropertyOrder({"name", "comment"})
	public static class TestSchema1 {
		
		@JsonProperty("theName")
		String name;
		String comment;
		
		public String getComment() {
			return comment;
		}
		
		public String getName() {
			return name;
		}
	}
	
	@JsonPropertyOrder({"theName", "comment"})
	public static class TestSchema2 {
		
		@JsonProperty("theName")
		String name;
		String comment;
		
		public String getComment() {
			return comment;
		}
		
		public String getName() {
			return name;
		}
	}
}

```

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
MikeEdgar: Added support for boolean methods name `isXXX`.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
MikeEdgar: Added support for ordering by customized names and using customized names via JAXB element and attribute annotations. Note that ordering by Jsonb and JAXB customized names may not be how those frameworks serialize objects.

MikeEdgar:
Several updates have been pushed to this PR. It was also missing logic to compare primitives and wrapper types when looking for methods/fields for the same property.

EricWittmann:
@kenfinnigan I think the semaphoreci failure is transient - can you kick off a rebuild (or give me access to do it)?

kenfinnigan:
Do you not have access to Semaphore?

If not, create an account and let me know your ID

EricWittmann:
> Do you not have access to Semaphore? If not, create an account and let me know your ID

My semaphore ID:  https://semaphoreci.com/ericwittmann

kenfinnigan:
I've retriggered the PR job, and added you as an admin to CI

MikeEdgar:
Re-based. I think this one is ready to go (or for additional feedback).

EricWittmann:
All checks pass, and it looks pretty good to me (a lot of work in here!).  Wondering if @msavy has any thoughts - if not I'll merge.

EricWittmann:
Thanks!

PR #152, opened 2019-07-17
Enable customization of generated responses

### LINKED ISSUE #100: Enable APIResponse auto-generation with annotations present
Methods returning a specific object type (with the exception of JAX-RS `Response`) do not need `@APIResponse` because of the auto detection, which is very handy.
Unfortunately you often cannot use that feature for some (most?) JAX-RS implementations. Because if you return `null`, RESTeasy for example returns 204 No Content and this cannot be declared without needing to also declare the `@APIResponse` for 200.

This is because the current logic does not auto generate anything if there is any explicitly defined `@APIResponse` annotation present:
https://github.com/smallrye/smallrye-open-api/blob/8e05eeb79e800312da9ef49b1a6821424bb069a4/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java#L627

There are multiple ways this could be changed:
1. Auto generate default responses if the respective `responseCode` (200 for non-void methods, 201/204 for void methods) has not been explicitly defined. Although you then cannot disable them you can still override them.
2. Auto generate default responses if an empty `@APIResponse` annotation is present.
3. Introduce a new annotation or option to recognize that a methods return value is nullable. Then the auto generation can add a 204 response addtional to the 200.

MikeEdgar:
~~Addresses~~ Fixes #100 

Work in progress (missing test cases) - for discussion only. Sorry about the white space changes in the JavaDocs. I'm not sure if it's my IDE or the code formatting in the Maven process.

*Edit 8/7*: update 'fixes' labeling.

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConstants.java
+++ b/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConstants.java
@@ -42,6 +42,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
+import javax.ws.rs.container.AsyncResponse;
 import javax.ws.rs.core.Application;
 import javax.ws.rs.core.Response;
 
@@ -247,6 +248,7 @@ public final class OpenApiConstants {
     public static final DotName DOTNAME_HEADER_PARAM = DotName.createSimple(HeaderParam.class.getName());
     public static final DotName DOTNAME_MATRIX_PARAM = DotName.createSimple(MatrixParam.class.getName());
     public static final DotName DOTNAME_BEAN_PARAM = DotName.createSimple(BeanParam.class.getName());
+    public static final DotName DOTNAME_ASYNC_RESPONSE = DotName.createSimple(AsyncResponse.class.getName());
 
     // RestEasy parameter extension annotations
     public static final DotName DOTNAME_RESTEASY_QUERY_PARAM = DotName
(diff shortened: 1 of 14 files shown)


neon-dev:
This looks very promising.
I just had an idea regarding disabling auto-generation. Instead of an empty `@APIResponses` I suggest a boolean `disableContentGeneration` property in this annotation.
Otherwise you would need to enable declaring  `@APIResponse`s and `@APIResponses` on the same method which is currently not possible. So with the property instead you can disable auto-generation but will have to declare `@APIResponse`s inside `@ApiResponses` instead of outside it.

Secondly, looking at the current code it doesn't look like the current PR could detect an empty `@APIResponses` (if there are separate `APIResponse`s) anyway. Because I believe Jandex unwraps the `@APIResponses` annotation and merges all contained items into a list of `APIResponse` instances. So the `responses` variable should not be empty once there is a `@APIResponse` defined outside of the wrapper.

neon-dev:
Offtopic but maybe relevant for you @MikeEdgar, I had the same issue regarding trailing whitespaces with IntelliJ on an Eclipse formatted project. Don't completely remember but I guess it was fixed by setting [this option](http://prntscr.com/ogvesk) from "All" to "Modified lines".

MikeEdgar:
>  it doesn't look like the current PR could detect an empty `@APIResponses`

You are right - I'll get that fixed.

EricWittmann:
This looks like a great start.

MikeEdgar:
Pushed updates, this should work now as discussed in #100. This will still need to be squashed/re-based after #148 is merged. There is some overlap with the Resteasy annotation constants and the POMs.

MikeEdgar:
I believe this one is ready for final review - the functionality should be in alignment with the discussion over in #100. Nothing significant has really changed in a few weeks, mainly just re-basing on master. This also includes a bump in the jackson-databind component to address the security vulnerability referenced in #153.

@neon-dev, I think it will be up to you to decide if this qualifies as a full fix for issue #100 (or comment what else is missing).

EricWittmann:
I'll give this until the end of the week for further comment before merging.

neon-dev:
Looking at the tests it looks like it does everything required to fix #100 and more. Think it's a very elegant and handy solution, thanks :)

MikeEdgar:
After this and the other bot changes are merged, would it be a good time for a new release?

kenfinnigan:
We're currently changing the artifact/version scheme back to what we had previously, so I'm cascading my way through releasing everything.

Over the next day or so OpenAPI will likely get released as part of that.

EricWittmann:
@kenfinnigan Can you merge any of the dependabot PRs you think are appropriate as part of that work?

EricWittmann:
I would recommend not bumping Jackson, however.  I don't think the `pr1` build is stable.  I made that mistake in Apicurio and some users suffered for it.

kenfinnigan:
Yep, agree

kenfinnigan:
So this is good to be merged?

PR #167, opened 2019-08-17
Use RolesAllowed/DeclareRoles values as OAuth2/OIDC security scopes

### LINKED ISSUE #15: Process @RolesAllowed while generating OpenAPI doc from annotations
Hello,
I raised a discussion whether this would be possible to support. Would be interesting to investigate any future support or enhancement of this kind. Let me know if I am out of line though :)

https://github.com/eclipse/microprofile-open-api/issues/244

BR

David

MikeEdgar:
Fixes #15 

Differences from the discussion in #15 -
1. Presence of a `@PermitAll` annotation on a method will prevent a resource's `@RolesAllowed` from applying to an `Operation`.
2. Presence of a `@DenyAll` annotation on a method will generate an empty scope array on an `Operation`.

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
@@ -458,6 +458,15 @@ public class OpenApiAnnotationScanner {
         }
     }
 
+    /**
+     * Adds the array of roles as scopes to each of the OAuth2 flows stored previously.
+     * The flows are those declared by the application in components/securitySchemes
+     * using annotations where the scopes were not defined. The description of the scope
+     * will be set to the role name plus the string " role".
+     *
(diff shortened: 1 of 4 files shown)


implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -384,6 +458,19 @@ private void processJaxRsResourceClass(OpenAPIImpl openApi, ClassInfo resourceCl
         }
     }
 
+    void addScopes(String[] roles) {
```
EricWittmann: I think this could use a little bit of javadoc.

MikeEdgar: @EricWittmann - added the method JavaDoc. Please let me know if there is anything else.

PR #169, opened 2019-08-28
Optimize the java bean scanner when running in Quarkus

EricWittmann:
When running in Quarkus, it seems that the data type scanner (which is basically responsible for converting a java bean into a JSON schema) was being overly aggressive.  For example, it was treating the standard `getClass()` method as a property of a bean.  I have improved the filtering of the fields and methods to ensure that doesn't happen.  This cuts **way** down on the amount of scanning being done, and also seems to fix the "stream closed" problem reported in https://github.com/quarkusio/quarkus/issues/3728.  

I'm not 100% sure what is causing the "stream closed" error in Jandex, but I do know that it's way down the rabbit hole (trying to scan multiple java.* classes) by the time it fails.

I'm also not sure why the scanning seems to be more aggressive in Quarkus than elsewhere, but I'm certain there's a good explanation for that. :)

I also added a slight delay after shutting down the junit HTTP server to give the port time to be reclaimed.  I was seeing a few "port already in use" errors during the maven build.

### DIFF
--- a/tck/src/test/java/io/smallrye/openapi/tck/extra/complex/ComplexResource.java
+++ b/tck/src/test/java/io/smallrye/openapi/tck/extra/complex/ComplexResource.java
@@ -37,7 +37,6 @@ public class ComplexResource {
     public CollectionWrapper postArrayOfInts(Integer[] ints) {
         CollectionWrapper result = new CollectionWrapper();
         result.getValues().addAll(Arrays.asList(ints));
-        result.setData(new byte[] { 1, 2, 3, 4, 5 });
         return result;
     }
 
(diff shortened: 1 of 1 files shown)


MikeEdgar:
> I'm also not sure why the scanning seems to be more aggressive in Quarkus than elsewhere

@EricWittmann - do you know if Quarkus is including the core Java classes in the Jandex index? That would explain the `getClass` method being scanned.

MikeEdgar:
I'm only guessing, but perhaps only the `beanArchiveIndexBuildItem` should be used here:

https://github.com/quarkusio/quarkus/blob/9aff3d16231ade5b9386d5eedff40d0a1c199b67/extensions/smallrye-openapi/deployment/src/main/java/io/quarkus/smallrye/openapi/deployment/SmallRyeOpenApiProcessor.java#L108

EricWittmann:
That's a good theory - I can def. confirm that the core java classes are being included in the jandex index.

PR #183, opened 2019-09-28
Update jackson-databind and jackson-dataformat-yaml to 2.10.0

MikeEdgar:
Quote numeric YAML keys in unit test expectation. This should fix the issue with PR #182


EricWittmann:
Rats - I didn't see that you fixed this already in a separate PR.  I just pushed a commit here:  

https://github.com/smallrye/smallrye-open-api/pull/182

@MikeEdgar - Let me know what you think about unifying the jackson versions.  I'm not sure why we have different properties for each of the jackson modules.  I can't imagine we'll want different versions of them.

MikeEdgar:
@EricWittmann, I think unifying the versions is a good idea. I believe they diverged to update for the security vulnerabilities that occurred recently. I do not think it had anything to do with functionality. 

Closing this PR.

PR #188, opened 2019-10-04
Ensure java.lang not scanned, add jackson-annotations, enum scan fix

### LINKED ISSUE #185: java.lang.OutOfMemoryError: Java heap space when OpenAPI is enabled and enums are used in a JAX-RS endpoint
As reported in https://github.com/quarkusio/quarkus/issues/4329, an OOME is thrown when the quarkus-smallrye-openapi extension is used and a JAX-RS endpoint has an enum as a query parameter. It works if smallrye-openapi 1.1.3 is used

### LINKED ISSUE #186: Enum types as parameters and responses not correctly scanned
When enums are used as resource parameters or responses, the generated schema is a plain object rather than the expected string with a list of the enum values.

### LINKED ISSUE #187: Regression: exclusion of jackson-annotations breaks Quarkus
Quarkus is relying on Maven dependency resolution to include jackson-annotations in the final jar file. The dependencyManagement entry added in #179 needs to be reverted before the next release.

MikeEdgar:
Fixes #185 - always make top-level index a ```FilteredIndexView``` and always exclude ```java.lang``` from scanning
Fixes #186 - handle enum types in ```OpenApiDataObjectScanner```, move some common code into utility classes
Fixes #187 - remove ```dependencyManagement``` for jackson-annotations

Also refactored ```JandexUtil#getJaxRsResourceClasses``` to more easily remove duplicates with a ```Stream```. The Quarkus Jandex index tree contains duplicate entries which results in resources being scanned multiple times.

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
@@ -164,7 +164,13 @@ public class OpenApiAnnotationScanner {
      */
     public OpenApiAnnotationScanner(OpenApiConfig config, IndexView index, List<AnnotationScannerExtension> extensions) {
         this.config = config;
-        this.index = index;
+
+        if (index instanceof FilteredIndexView) {
+            this.index = index;
+        } else {
+            this.index = new FilteredIndexView(index, config);
+        }
+
         this.extensions = extensions;
     }
 
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -29,16 +29,6 @@
 
     <name>SmallRye: MicroProfile OpenAPI Implementation</name>
 
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>com.fasterxml.jackson.core</groupId>
-                <artifactId>jackson-annotations</artifactId>
-                <scope>test</scope>
-            </dependency>
-        </dependencies>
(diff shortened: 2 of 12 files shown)


implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -164,7 +164,13 @@ public OpenApiAnnotationScanner(OpenApiConfig config, IndexView index) {
      */
     public OpenApiAnnotationScanner(OpenApiConfig config, IndexView index, List<AnnotationScannerExtension> extensions) {
         this.config = config;
-        this.index = index;
+
+        if (index instanceof FilteredIndexView) {
+            this.index = index;
+        } else {
+            this.index = new FilteredIndexView(index, config);
+        }
```
EricWittmann: Nice.  Make sure it's always filtered.  +1

EricWittmann: Also I'm excited to have used multi-line commenting for the first time.  Way to go, GitHub. :)

MikeEdgar: > 
> 
> Also I'm excited to have used multi-line commenting for the first time. Way to go, GitHub. :)

It's pretty nice - helps with context!

Issue #189, opened 2019-10-05
Add scanning of JAX-RS sub-resources


MikeEdgar:
The annotation scanner does not currently support scanning [sub-resources](https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/jaxrs-resources.html#d0e2542). This has impacts on the main resource class/method scanning as well as parameter processing.

Quarkus: quarkusio/quarkus#3918

MikeEdgar:
Fixed by #190

PR #190, opened 2019-10-05
implemented scanning for subresources

balta3:
Implemented support for scanning and analyzing subresource locators in JAX-RS.


MikeEdgar:
@balta3 , thanks for the PR. I left a few comments in the code. Please also add several unit tests to help demonstrate the cases this will support and help ensure there are no obvious issues like null pointers, etc.

MikeEdgar:
Fixes #189

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/ParameterProcessor.java
```diff
@@ -1230,6 +1232,22 @@ static boolean isBeanPropertyParam(AnnotationInstance annotation) {
         return relevant;
     }
 
+    /**
+     * Determines if the give method is a JAX-RS sub-resource locator method
```
EricWittmann: "given"? :)

MikeEdgar: Next time... ;-)

Issue #191, opened 2019-10-08
Class loading for reflection breaks Quarkus ORM enhancement


MikeEdgar:
Original issue: quarkusio/quarkus#2961
Regression caused by fix for quarkusio/quarkus#4329 and #185

MikeEdgar:
@FroMage in #185 :
> Apparently, if we are able to pass you a ClassLoader instead of you using Class.forName (which will always use the CL of smallrye-openapi), we can make this work. Or you can use the TCCL.
>
>I can make a PR for either solution if you want.

Using the TCCL in `TypeUtil.getClass(String name)` instead of `Class.forName` probably will be the easiest and quickest at this point. @EricWittmann, any concerns?

EricWittmann:
No concerns I can think of at the moment. :)

MikeEdgar:
I'm thinking it might be a good idea to have an internal index in `TypeUtil` with the JDK classes we are known to be looking for, specifically `java.lang.Enum` and the collection types. That should remove the need for reflection. It would be something like the small internal index that is used in `OpenApiDataObjectScanner` for the stand-in types. By doing that, we won't impact class field scanning, but still be able to determine the relevant class hierarchies using Jandex only.

I know a fix for this is very likely to be time sensitive so that change can happen later.

kenfinnigan:
@MikeEdgar is this something you're actively working on a fix for?

MikeEdgar:
@kenfinnigan - not yet. If @FroMage has not already started I can pick it up, but I won't be able to do that until this evening.

kenfinnigan:
Ok, I can take a look now.

FroMage:
Thanks @kenfinnigan

Issue #193, opened 2019-10-10
UUID Support


waggy55:
At present, it seems the UUID type is not recognised. For projects that define the API in yaml, this is not a problem, because the type can be defined in the yaml file. However, for projects where the API is created by parsing the code, the UUID type is represented as:
```
{
   "leastSigBits": 0,
    "mostSigBits": 0,
    "leastSignificantBits": 0,
    "mostSignificantBits": 0
}
```
This is quite unusable (at least not quick to use). For microservices environments, the API often contains UUIDs, so support for this would be much appreciated.

Many thanks,
Dan

MikeEdgar:
@waggy55 - are you using Quarkus?

waggy55:
@MikeEdgar yes I am

MikeEdgar:
Thank you. I'm assuming you're expecting to see the UUID type represented as a string?

We are going to look at other ways to account for the fact that the class index being passed in contains JDK classes that the unit tests assume are not there.

waggy55:
Yes, to see the UUID as a string would be great! Thank you for adding it to "To do".

MikeEdgar:
@EricWittmann , @msavy - can either of you think of a reason that `java.util.UUID` should not be added to the `TypeUtil` `TYPE_MAP` with a value of `STRING_FORMAT`? I think this is a pretty common type and it is supported by JAX-RS out of the box via its static `fromString` method for parsing.

EricWittmann:
I can't think of a reason.  The only thing I would say is that it might be nice to map UUID to a more interesting schema type that perhaps validates the format of a typical Java UUID.  In other words, rather than just map it to string, maybe we can do better and generate a re-usable type that is a string with a regex pattern or some such.

Not a big deal for me - just a thought.  I'm more than happy mapping it to string.

MikeEdgar:
>  In other words, rather than just map it to string, maybe we can do better and generate a re-usable type that is a string with a regex pattern or some such.

I really like this idea. This has been implemented in #195. I think there may be a number of classes that fall into this category and adjusting the approach for eclipse/microprofile-open-api#364 may be a good idea as well.

PR #194, opened 2019-10-14
Reduce the occurrence of reflection for many known cases

MikeEdgar:
Relates to #191 and also the fix for quarkusio/quarkus#2961.

The idea with these changes is to mostly eliminate the fallback to reflection during the "is a" tests used by class/annotation scanning by including a small internal index of JDK classes most commonly used. Ideally, this would eliminate the need for any downstream work-around like in [the fix for quarkus 2961](https://github.com/quarkusio/quarkus/commit/cc654c6f0fef16f0f37554ad4d22c7a826ced210#diff-25a0f7545a16a5e105652956c74d3736).

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/util/TypeUtil.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/util/TypeUtil.java
@@ -15,6 +15,9 @@
  */
 package io.smallrye.openapi.runtime.util;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
 import java.math.BigDecimal;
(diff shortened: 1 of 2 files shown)


MikeEdgar:
@kenfinnigan , I tested this with the repo referenced in quarkusio/quarkus#2961 using Hibernate+Panache and Quarkus 0.24.0. I can run some additional tests with master as well.

MikeEdgar:
> However, given the Quarkus model, this approach now likely makes more sense than before.

Thanks @msavy for the feedback. I think going down this path will allow the annotation scanner to further skip introspection of some of the JDK packages to prevent unexpected results like seen in #193.

PR #198, opened 2019-10-19
Use property name from mutator methods, skip annotated getters

### LINKED ISSUE #197: RESTEasy param annotations on fields or properties causing mayem
I've added RESTEasy param annotations on fields and properties of my class:

```java
@Path("/test")
public class TestResource {

    @GET
    public String getTest() {
        return "TEST";
    }

    // make sure these don't break the build when fields
    @org.jboss.resteasy.annotations.jaxrs.PathParam
    String pathField;
    @org.jboss.resteasy.annotations.jaxrs.FormParam
    String formField;
    @org.jboss.resteasy.annotations.jaxrs.CookieParam
    String cookieField;
    @org.jboss.resteasy.annotations.jaxrs.HeaderParam
    String headerField;
    @org.jboss.resteasy.annotations.jaxrs.MatrixParam
    String matrixField;
    @org.jboss.resteasy.annotations.jaxrs.QueryParam
    String queryField;

    // make sure these don't break the build when properties
    public String getPathProperty() {
        return null;
    }

    @org.jboss.resteasy.annotations.jaxrs.PathParam
    public void setPathProperty(String p) {
    }

    public String getFormProperty() {
        return null;
    }

    @org.jboss.resteasy.annotations.jaxrs.FormParam
    public void setFormProperty(String p) {
    }

    public String getCookieProperty() {
        return null;
    }

    @org.jboss.resteasy.annotations.jaxrs.CookieParam
    public void setCookieProperty(String p) {
    }

    public String getHeaderProperty() {
        return null;
    }

    @org.jboss.resteasy.annotations.jaxrs.HeaderParam
    public void setHeaderProperty(String p) {
    }

    public String getMatrixProperty() {
        return null;
    }

    @org.jboss.resteasy.annotations.jaxrs.MatrixParam
    public void setMatrixProperty(String p) {
    }

    public String getQueryProperty() {
        return null;
    }

    @org.jboss.resteasy.annotations.jaxrs.QueryParam
    public void setQueryProperty(String p) {
    }

}
```

And the generated documentation became a bit crazy:

```
  "/test{testMatrix}": {
    "get": {
      "responses": {
        "200": {
          "description": "OK",
          "content": {
            "*/*": {
              "schema": {
                "type": "string"
              }
            }
          }
        }
      }
    },
    "parameters": [
      {
        "name": "p",
        "in": "path",
        "required": true,
        "schema": {
          "type": "string"
        }
      },
      {
        "name": "pathField",
        "in": "path",
        "required": true,
        "schema": {
          "type": "string"
        }
      },
      {
        "name": "testMatrix",
        "in": "path",
        "required": true,
        "schema": {
          "type": "object",
          "properties": {
            "p": {
              "type": "string"
            },
            "matrixField": {
              "type": "string"
            }
          }
        },
        "style": "matrix"
      },
      {
        "name": "p",
        "in": "query",
        "schema": {
          "type": "string"
        }
      },
      {
        "name": "queryField",
        "in": "query",
        "schema": {
          "type": "string"
        }
      },
      {
        "name": "headerField",
        "in": "header",
        "schema": {
          "type": "string"
        }
      },
      {
        "name": "p",
        "in": "header",
        "schema": {
          "type": "string"
        }
      },
      {
        "name": "cookieField",
        "in": "cookie",
        "schema": {
          "type": "string"
        }
      },
      {
        "name": "p",
        "in": "cookie",
        "schema": {
          "type": "string"
        }
      }
    ]
  },
```

Several issues here:

- It's using the setter property's parameter name instead of the property name: `p` should be `cookieProperty`
- If I put the annotation on the getter by mistake, I get an NPE:
```
Caused by: java.lang.NullPointerException
	at io.smallrye.openapi.runtime.util.SchemaFactory.typeToSchema(SchemaFactory.java:230)
	at io.smallrye.openapi.runtime.scanner.ParameterProcessor.setSchemaProperties(ParameterProcessor.java:612)
	at io.smallrye.openapi.runtime.scanner.ParameterProcessor.getParameters(ParameterProcessor.java:486)
	at io.smallrye.openapi.runtime.scanner.ParameterProcessor.process(ParameterProcessor.java:352)
	at io.smallrye.openapi.runtime.scanner.OpenApiAnnotationScanner.processJaxRsMethod(OpenApiAnnotationScanner.java:584)
	at io.smallrye.openapi.runtime.scanner.OpenApiAnnotationScanner.processJaxRsResourceClass(OpenApiAnnotationScanner.java:437)
	at io.smallrye.openapi.runtime.scanner.OpenApiAnnotationScanner.scan(OpenApiAnnotationScanner.java:214)
```
- What's with the `testMatrix` name? Where did it come from? It should be either `matrixField` or `matrixProperty`, and what is it doing in the URI?
- Why do I get `"parameters"` sometimes inside `get` and sometimes after it as a sibling? This is allowed?

MikeEdgar:
Fixes #197


FroMage:
Looks good, but you still have `testMatrix` in there, so it doesn't completely fix #197 yet.

MikeEdgar:
@FroMage , I still think the name of the matrix needs to be present in the path (`in` = `path` for `style` = `matrix`). I did post on #197 that I thought that was incorrect, but later removed that comment when I realized I was reading the section of the OAI spec dealing with param serialization.

There are not many examples of this other than the [Stack Overflow answer](https://stackoverflow.com/a/44697098/738343) I mentioned in #197. The fact that the format used there works with the Swagger UI tooling is good evidence that it is correct for matrix style path parameters.

Please note that the name of the parameter can be customized using a `@Parameter` annotation.

MikeEdgar:
Pushed an update to default the value of `explode` to `true` for generated matrix `@Parameters` which was previously overlooked.

@EricWittmann , what are your thoughts on the matrix parameter representation? My line of thinking is that the JAX-RS `@MatrixParam`s represent the properties of a named matrix that coincides with the closest path segment. If the name isn't provided, the code is currently generating one using the path segment value. Unfortunately there is nothing in the TCK for this style of parameter and there are not many examples of the OAI output.

FroMage:
What I meant is that the name `testMatrix` is invalid: it doesn't come from the annotated element. If it's present, it should be `matrixField` and `matrixProperty`. And it should be consistent with a `@MatrixParam` as method parameter, which ATM do not appear in the path IIRC.

FroMage:
Oh, I see, it's a composite OpenApi structure that points to the real matrix params. Can you add a test that verifies that a method parameter `@MatrixParam` behaves the same?

FroMage:
> My line of thinking is that the JAX-RS @MatrixParams represent the properties of a named matrix that coincides with the closest path segment

The JAX-RS spec does not actually specify this, as crazy as it sounds. So, perhaps the TCK enforces a behaviour but it's not specified.

MikeEdgar:
> Can you add a test that verifies that a method parameter `@MatrixParam` behaves the same?

There are a few scenarios in [ParameterScanTests.java](https://github.com/smallrye/smallrye-open-api/blob/20e7842c620c39ea46b8cd3822b54fc064b80e23/implementation/src/test/java/io/smallrye/openapi/runtime/scanner/ParameterScanTests.java) from the original implementation that include those. Let me know if you're thinking something else and I'll add them.

As far as the naming of the matrix in the path itself, you're right that it's not specified by JAX-RS. That is more of an OpenAPI construct. JAX-RS simply associates the parameters with the path segment. OpenAPI needs the name added to the path so that the parameter object (containing the matrix param primitives) can be referenced.

FroMage:
> As far as the naming of the matrix in the path itself, you're right that it's not specified by JAX-RS

No, I meant that it does not specify which segment the matrix params apply to.

MikeEdgar:
Take a look at the JavaDoc for `javax.ws.rs.MatrixParam`

> Note that the @MatrixParam annotation value refers to a name of a matrix parameter that resides in the last matched path segment of the Path-annotated Java structure that injects the value of the matrix parameter.

I read that as meaning that `@MatrixParam`s in the arguments of a resource method apply to the `@Path` of the method (if present) and that `@MatrixParam`s injected as properties of the resource class apply to the `@Path` on the resource class itself.

FroMage:
You're right, I only checked the spec and not the javadoc. It would indeed apply to the last path segment.

EricWittmann:
Just needs some conflict resolution.

MikeEdgar:
I need to resolve the conflicts this has from the UUID PR, but @FroMage , please let me know if you are otherwise in agreement that the representation of the matrix parameter is correct.

MikeEdgar:
I re-based on master which dismissed your review/approval @EricWittmann. The diff is identical but please give it another look over when you have some time.

EricWittmann:
DOUBLE APPROVAL

PR #204, opened 2019-11-05
Schema scan enhancements

### LINKED ISSUE #201: Is it possible to flatten nested POJOs
See eclipse/microprofile-open-api#380 for original discussion.

MikeEdgar:
This PR includes some significant updates to the schema scanning process. Feedback/discussion/criticism is welcome.

- *** Support schema composition on property schemas using `allOf`. This allows for the global schema of the field's type (which is separately registered in `components`) to be combined with locally-declared schema data such as `description` or constraints.
- *** Change default value of `mp.openapi.extensions.schema-references.enable` to `true`
- Fixes #201 when `implementation` is specified.
- Fix class schema registration, ensures the class's `@Schema` information is the data used in `components` top-level entries
- Do not process using the field type if `@Schema` `implementation` is specified
- Add support for `@Schema` on `enum` classes. As pointed out by @pe-st in his ["Code First or Design First?" presentation](https://github.com/pe-st/apidocs), this was not previously supported
- Do not scan for or generate `properties` for non-object schemas, also pointed out by @pe-st
- Prune deprecated `AnnotationScannerExtension` methods and define defaults in interface
- Deprecate `DefaultAnnotationScannerExtension` as default methods are defined in interface
- Fix a few minor Sonar issues

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
+++ b/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
@@ -162,8 +162,8 @@ public class OpenApiConfigImpl implements OpenApiConfig {
      */
     @Override
     public Set<String> pathServers(String path) {
-        String servers = getConfig().getOptionalValue(OASConfig.SERVERS_PATH_PREFIX + path, String.class).orElse(null);
-        return asCsvSet(servers);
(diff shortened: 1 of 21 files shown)


MikeEdgar:
I'll leave this open for a few more days and rebase/merge if there is no feedback or objection.

EricWittmann:
It was +1 from me - fine to merge it now I think.

PR #206, opened 2019-11-05
Archive path might start with . (dot) breaking inclusion/exclusion

### LINKED ISSUE #205: ArchiveUtil doesn't correctly include and exclude classes and packages
According to the documentation [MicroProfile OpenAPI Specification > 3.1.1. Core configurations](https://download.eclipse.org/microprofile/microprofile-open-api-1.0/microprofile-openapi-spec.html#_core_configurations) inclusion and exclusion of packages and classes follows the dot-notation, e.g. ```mp.openapi.scan.exclude.classes=com.xyz.MyClassC,com.xyz.MyClassD```.

However, ArchiveUtil's ```#acceptClassForScanning``` determines the fqcn wrongly so that no inclusion/exclusion is possible. This mainly happens when using ShrinkWrap archives where the archive path starts with a leading slash.

In order to successfully exclude/include packages and classes, one has to specify them with a leading (dot):
```mp.openapi.scan.exclude.classes=.com.xyz.MyClassC,.com.xyz.MyClassD```

Please find a test case here: https://gist.github.com/user667/7be8f83e405e7c31fe273b0f458e3414

I would expect exclusion/inclusion to work as described in the documentation.

I experienced this behavior while working with Thorntail 2.5.0.Final and ```io.thorntail:microprofile-openapi```.

### PR DESCRIPTION
**Pull request by user667**
Fixes #205

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/api/util/ArchiveUtil.java
+++ b/implementation/src/main/java/io/smallrye/openapi/api/util/ArchiveUtil.java
@@ -221,6 +221,9 @@ public class ArchiveUtil {
         }
         String fqcn = archivePath.replaceAll("/", ".").substring(0, archivePath.lastIndexOf(OpenApiConstants.CLASS_SUFFIX));
         String packageName = "";
+        if (fqcn.startsWith(".")) {
+            fqcn = fqcn.substring(1);
+        }
         if (fqcn.contains(".")) {
             int idx = fqcn.lastIndexOf(".");
             packageName = fqcn.substring(0, idx);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by user667** [comment]
Updated PR: Changed fix so fqcn is sanitized before last . (dot) is removed and also applied formatting changes from ```formatter-maven-plugin```.

File: implementation/src/main/java/io/smallrye/openapi/api/util/ArchiveUtil.java [hunk 1]
```diff
@@ -221,6 +221,9 @@ private static boolean acceptClassForScanning(OpenApiConfig config, String archi
         }
         String fqcn = archivePath.replaceAll("/", ".").substring(0, archivePath.lastIndexOf(OpenApiConstants.CLASS_SUFFIX));
         String packageName = "";
+        if (fqcn.startsWith(".")) {
+            fqcn = fqcn.substring(1);
+        }
```
MikeEdgar: This will work, but I wonder if it would be cleaner to do a substring(1) up on line 158 and then remove the leading `/` from `OpenApiConstants.WEB_ARCHIVE_CLASS_PREFIX`. Thoughts?

PR #227, opened 2019-12-06
Auto scan annotated exception mapper instances

### LINKED ISSUE #220: Explore scanning of annotated ExceptionMapper instances
@lamtrhieu, @EricWittmann - idea for consideration.

Suppose an application defines the following classes. It might be interesting for the scanner to find instances of `ExceptionMapper` that map declared checked exceptions or (undeclared?) runtime exceptions, adding any discovered `@APIResponse` annotations from the mapper to the path/method being scanned. This would be a convenient way to have all operations include a 404 or 500 response without declaring it with annotations throughout an application.

```java
@Provider
public class AppServerExceptionMapper implements ExceptionMapper<AppServerException> {
  @APIResponse(responseCode = "500", ...)
  Response toResponse(AppServerException exception) {
    ...
  }
}

@Path("/app/{id}")
public class AppResource {
  @GET
  public String getSomething(@PathParam("id") String id) 
      throws AppServerException {
    ...
  }

  @PUT
  public String updateSomething(@PathParam("id") String id, String newVersion) 
      throws AppServerException {
    ...
  }
}
```

lamtrhieu:
Fixes #220

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
+++ b/implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
@@ -20,22 +20,13 @@ import java.beans.PropertyDescriptor;
 import java.lang.reflect.Method;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.ws.rs.core.Application;
+import javax.ws.rs.ext.ExceptionMapper;
 
 import org.eclipse.microprofile.openapi.annotations.enums.Explode;
 import org.eclipse.microprofile.openapi.models.Components;
@@ -220,6 +211,8 @@ public class OpenApiAnnotationScanner {
         for (ClassInfo resourceClass : resourceClasses) {
             processJaxRsResourceClass(oai, resourceClass, null);
         }
+
+
 
         if (oai != null) {
             // Sort the tags unless the application has defined the order in OpenAPIDefinition annotation(s)
@@ -532,6 +525,10 @@ public class OpenApiAnnotationScanner {
             List<Parameter> locatorPathParameters) {
         LOG.debugf("Processing jax-rs method: {0}", method.toString());
 
+
+
+
+
         final Operation operation;
 
         // Process any @Operation annotation
@@ -677,6 +674,33 @@ public class OpenApiAnnotationScanner {
         if (apiResponses == null || !JandexUtil.isEmpty(apiResponses)) {
             createResponseFromJaxRsMethod(method, operation);
         }
+
+        //Add common response using exception mapper
+        //Collection<ClassInfo> exceptionMappers = this.index.getAllKnownImplementors(DotName.createSimple(ExceptionMapper.class.getName()));
+
+        Collection<ClassInfo> exceptionMappers = this.index.getKnownDirectImplementors(DotName.createSimple(ExceptionMapper.class.getName()));
+        List<AnnotationInstance> collect = exceptionMappers.stream()
+                .map(classInfo -> classInfo.annotations().get(DotName.createSimple(org.eclipse.microprofile.openapi.annotations.responses.APIResponse.class.getName())))
+                .flatMap(List::stream).collect(Collectors.toList());
+
+        method.exceptions();
+
+
+        ClassInfo classInfo = exceptionMappers.stream().findFirst().get();
+
+        Map<DotName, List<AnnotationInstance>> mapperToResponse = classInfo.annotations();
+        List<AnnotationInstance> apiResponseAnnotationInstances = mapperToResponse.get(DotName.createSimple(APIResponse.class.getName()));
+        AnnotationInstance test = (AnnotationInstance) mapperToResponse.get(DotName.createSimple(org.eclipse.microprofile.openapi.annotations.responses.APIResponse.class.getName())).get(1);
+
+        String responseCode = JandexUtil.stringValue(test, OpenApiConstants.PROP_RESPONSE_CODE);
+        if (responseCode == null) {
+            responseCode = APIResponses.DEFAULT;
+        }
+        APIResponse response = readResponse(test);
+        responses = ModelUtil.responses(operation);
+        responses.addAPIResponse(responseCode, response);
+
+
 
         // Process @SecurityRequirement annotations
         ///////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


lamtrhieu:
@EricWittmann @MikeEdgar 
Can you please help to review this one. I try to generate more response using the exceptions in the jax-rs methods.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -428,6 +420,30 @@ private void processJaxRsResourceClass(OpenAPIImpl openApi, ClassInfo resourceCl
...
+    /**
+     * Build a map between exception class name and its corresponding @ApiResponse annotation in the jax-rs exception mapper
+     *
+     */
+    private Map<DotName, AnnotationInstance> processExceptionMappers() {
+        Map<DotName, AnnotationInstance> exceptionHandlerMap = new HashMap<>();
+        Collection<ClassInfo> exceptionMappers = this.index.getKnownDirectImplementors(DotName.createSimple(ExceptionMapper.class.getName()));
+
+        for (ClassInfo classInfo : exceptionMappers) {
+            DotName API_RESPONSE = DotName.createSimple(org.eclipse.microprofile.openapi.annotations.responses.APIResponse.class.getName());
```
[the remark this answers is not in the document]
lamtrhieu: Thanks @MikeEdgar I will try to reuse that. I will spend some times this weekend to fix this. Sorry for delay.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -678,6 +692,17 @@ private void processJaxRsMethod(OpenAPIImpl openApi, ClassInfo resourceClass, Me
             createResponseFromJaxRsMethod(method, operation);
         }
 
+        //Add api response using list of exceptions in the methods and exception mappers
+        List<Type> methodExceptions = method.exceptions();
+
+        for (Type type : methodExceptions) {
+            DotName exceptionDotName = type.name();
+            if (exceptionAnnotationMap.keySet().contains(exceptionDotName)) {
```
lamtrhieu: @MikeEdgar What do you think about this one. Do we need to check if the method exception is subclasses of the exception declared in the mapper class?
I can also handle that situation by going up the hierarchy of the method exception to find any parent that match the mapper exception.

MikeEdgar: @lamtrhieu - possibly, but you might pull in too many results. For examples, if a method throws ExceptionA (which extends ExceptionB) and there is a mapper for both, it might be incorrect to include the ExceptionB response information on that particular method. That could always be added in later if it proves to be a desirable feature.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -821,6 +855,23 @@ private void processJaxRsMethod(OpenAPIImpl openApi, ClassInfo resourceClass, Me
...
+     * Add api response to api responses using the annotation information
+     *
+     * @param apiResponseAnnotation The api response annotation
+     * @param operation the method operation
+     */
+    private void addApiReponseFromAnnotation(AnnotationInstance apiResponseAnnotation, Operation operation) {
+        String responseCode = JandexUtil.stringValue(apiResponseAnnotation, OpenApiConstants.PROP_RESPONSE_CODE);
+        if (responseCode == null) {
+            responseCode = APIResponses.DEFAULT;
+        }
```
MikeEdgar: I know this method is now called from two locations, but I'm wondering if the exception responses should ever be set with "default". It might also be a good idea to skip the exception annotation if the operation already has the response code declared locally on the resource method.

Maybe in the case of thrown exceptions, it should be required that the `responseCode` be set on the `@APIResponse`, else it is ignored. Alternatively, if the thrown exception is one of the known JAX-RS exceptions that specifically relate to a response code (e.g. `javax.ws.rs.NotFoundException`), then a missing `responseCode` can be inferred. That could be a subsequent PR though.

lamtrhieu: @MikeEdgar Thanks for raising that. Let me try to come up with a solution.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -449,6 +456,31 @@ private void processJaxRsResourceClass(OpenAPIImpl openApi, ClassInfo resourceCl
...
+     * Build a map between exception class name and its corresponding @ApiResponse annotation in the jax-rs exception mapper
+     *
+     */
+    private Map<DotName, AnnotationInstance> processExceptionMappers() {
+        Map<DotName, AnnotationInstance> exceptionHandlerMap = new HashMap<>();
+        Collection<ClassInfo> exceptionMappers = this.index
+                .getKnownDirectImplementors(OpenApiConstants.DOTNAME_EXCEPTION_MAPPER);
+
+        for (ClassInfo classInfo : exceptionMappers) {
+            DotName exceptionDotName = classInfo.interfaceTypes().get(0).asParameterizedType().arguments().get(0).name();
```
EricWittmann: Is the interface type you want always going to be the one at index 0?

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -449,6 +456,31 @@ private void processJaxRsResourceClass(OpenAPIImpl openApi, ClassInfo resourceCl
...
+
+            MethodInfo toResponseMethod = classInfo.method("toResponse", Type.create(exceptionDotName, Type.Kind.CLASS));
```
EricWittmann: This is neat - users can annotate the `toResponse` method with `@ApiResponse`.  I like it. :)

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
lamtrhieu: Hi @MikeEdgar Can you please help to review my changes based on your comment ?

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
MikeEdgar: @lamtrhieu - you can ensure the correct interface by doing something like this:

`classInfo.interfaceTypes().stream().filter(it -> it.name().equals(OpenApiConstants.DOTNAME_EXCEPTION_MAPPER)).findFirst().get()`

That will avoid issues where the exception mapper implements two or more interfaces and the first one is not the exception mapper.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
MikeEdgar: @lamtrhieu - the approach you outlined sounds reasonable.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
EricWittmann: Nice - this is exactly what I was getting at.  +1

lamtrhieu: @MikeEdgar I have modify the code with your approach. Please help to review again. 
Thanks

MikeEdgar:
@lamtrhieu - I think this is looking pretty good now. @EricWittmann, do you have any additional feedback?

EricWittmann:
LGTM - nice job @lamtrhieu

PR #249, opened 2020-01-30
Move ArchiveUtil to TCK

### LINKED ISSUE #246: Move ArchiveUtils class out of src/main and move to src/test
The `ArchiveUtil` class is found here:

https://github.com/smallrye/smallrye-open-api/blob/master/implementation/src/main/java/io/smallrye/openapi/api/util/ArchiveUtil.java

This class lives in `src/main/java` but is only used by test classes.  We should move it to `src/main/test` so that we no longer have a non-test dependency on ShrinkWrap.  It's possible/likely it should be removed from **implementation** and into **tck** but I didn't check that.

gastaldi:
This gets rid of ShrinkWrap as a hard dependency on the smallrye-open-api implementation

Fixes #246


EricWittmann:
So I think the issue with this is that Thorntail actually does use the ArchiveUtil.  See the conversation here:  https://github.com/smallrye/smallrye-open-api/issues/246

MikeEdgar:
Hi @gastaldi - we were thinking the most straight-forward resolution would be to add `<optional>true</optional>` to the `shrinkwrap-api` dependency in implementation/pom.xml while keeping the `ArchiveUtil` and associated test class in the same location.

I think once this is done, we may be at a good place for a tag/release. There have been a few changes recently requested by downstream projects, including this.

gastaldi:
@MikeEdgar thanks! I'll close this PR and send another one with that change then.

PR #250, opened 2020-01-31
Add Extension parsing method to AnnotationScannerExtension

### LINKED ISSUE #245: Decouple OpenApiAnnotationScanner from Jackson
We are layering on SmallRye's OpenAPI implementation as much as we can but want to use a parser _other than_ Jackson.

We do our own parsing of any static file to create an `OpenAPI` instance, rather than using `OpenApiProcessor#modelFromStaticFile`, and we do our own serializing.

But `OpenApiAnnotationScanner` depends on Jackson's `ObjectMapper`...I think in two places inside `#parseExtension`. 

I don't see an easy way to continue to use `OpenApiAnnotationScanner` as-is without also relying on Jackson. 

Questions:
1. Am I missing an easy solution?
2. Is there any chance you could decouple `OpenApiAnnotationScanner` from Jackson's `ObjectMapper` and allow the caller to provide a way to accomplish the equivalent?
3. I know there are some other references to Jackson classes from elsewhere in SmallRye - e.g., `IgnoreResolver`, `JsonUtil`, and a few others - but I'm not sure they are in the code paths we'd continue to rely on.

Thanks.

MikeEdgar:
Fixes #245 

- Add tests to require that Jackson library is not necessary for
annotation scanning when providing custom implementation of
`AnnotationScannerExtension#parseExtension`
- Refactor `ExceptionMapper` search to eliminate Sonar "bugs"
- Replace TestNG `@Test` annotation with JUnit annotation in `ExceptionMapperScanTests`

### DIFF
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -117,7 +117,6 @@
             <scope>test</scope>
             <type>pom</type>
         </dependency>
-
     </dependencies>
 
     <build>
@@ -131,13 +130,36 @@
                         <SMALLRYE_MP_CONFIG_PROP>1234</SMALLRYE_MP_CONFIG_PROP>
                     </environmentVariables>
                     <redirectTestOutputToFile>true</redirectTestOutputToFile>
-                    <!--
-                    TODO: Confirm if these are necessary
-                    <enableAssertions>true</enableAssertions>
-                    <argLine>-Xmx512m</argLine>
-                    <forkMode>once</forkMode>
-                    -->
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/CustomExtensionParsingTests.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>jackson-exclusion-tests</id>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <includes>
+                                <include>**/CustomExtensionParsingTests.java</include>
+                            </includes>
+                            <systemProperties>
+                                <classpath.jackson.excluded>true</classpath.jackson.excluded>
+                            </systemProperties>
+                            <classpathDependencyExcludes>
+                                <classpathDependencyExclude>com.fasterxml.jackson.core:jackson-core</classpathDependencyExclude>
+                                <classpathDependencyExclude>com.fasterxml.jackson.core:jackson-databind</classpathDependencyExclude>
+                                <classpathDependencyExclude>com.fasterxml.jackson.dataformat:jackson-dataformat-yaml</classpathDependencyExclude>
+                            </classpathDependencyExcludes>
+                        </configuration>
+                    </execution>
+                </executions>
             </plugin>
         </plugins>
     </build>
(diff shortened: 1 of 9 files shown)


implementation/pom.xml
```diff
@@ -131,13 +130,36 @@
...
+                                <classpath.jackson.excluded>true</classpath.jackson.excluded>
+                            </systemProperties>
+                            <classpathDependencyExcludes>
+                                <classpathDependencyExclude>com.fasterxml.jackson.core:jackson-core</classpathDependencyExclude>
+                                <classpathDependencyExclude>com.fasterxml.jackson.core:jackson-databind</classpathDependencyExclude>
+                                <classpathDependencyExclude>com.fasterxml.jackson.dataformat:jackson-dataformat-yaml</classpathDependencyExclude>
+                            </classpathDependencyExcludes>
+                        </configuration>
+                    </execution>
+                </executions>
```
EricWittmann: TIL the surefire plugin is cooler than I realized.

MikeEdgar: Same - I have not used it in this way before. TIL about TIL, also :-D

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
```diff
@@ -47,4 +47,19 @@ default boolean isJaxRsAnnotationExtension(AnnotationInstance instance) {
...
+     * - JSON object - starts with '{'
+     * - JSON array - starts with '['
+     * - number
+     * - boolean
+     * - string
+     *
+     * @param key the name of the extension property
+     * @param value the string value of the extension
+     */
+    default Object parseExtension(String key, String value) {
```
MikeEdgar: @tjquinno, please let me know if this method definition will cover what you need to parse extensions.

tjquinno: Thanks @MikeEdgar . I think the answer is "yes." I had to shift gears to some other things briefly so I'm not in a position right now to try this (probably first thing next week). But I think it looks as if it'll work.

tjquinno:
@MikeEdgar Appreciate the _very_ fast turnaround on this. Is there somewhere I can look to see about the plans and timing for the next release?

MikeEdgar:
@tjquinno - minor releases are typically as needed.

@kenfinnigan, @EricWittmann, @msavy - I think we are at a good point to tag and release. Any objections?

msavy:
Makes sense to me -- quite a bit of stuff gone in recent days alone!

EricWittmann:
No objections.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/AnnotationScannerExtension.java
tjquinno: @MikeEdgar Just FYI, I'm using a local build of the latest SmallRye master (with this PR) and I see my `AnnotationScannerExtension` `parseExtension` method being invoked as expected. Thanks again.

MikeEdgar: @tjquinno - very good, glad to hear that. I expect there will be a release in the next few days.

MikeEdgar:
@kenfinnigan - please tag/release smallrye-open-api when you get a chance. Thanks

kenfinnigan:
As we're using jakarta dependencies now, does it require a version bump to 1.3.x?

MikeEdgar:
I am OK with bumping the minor version, but I think it should be 1.2.x. Does it matter if the major/minor differs from the upstream MP version?

kenfinnigan:
Sorry, I did mean 1.2.x.

I saw 1.1.20 and somehow read 1.2.x instead

PR #252, opened 2020-02-03
Added an option to disable processing of @ApplicationPath.

EricWittmann:


implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
```diff
@@ -350,7 +350,7 @@ private OpenAPIImpl jaxRsApplicationToOpenApi(ClassInfo applicationClass) {
         ////////////////////////////////////////
         AnnotationInstance appPathAnno = JandexUtil.getClassAnnotation(applicationClass,
                 OpenApiConstants.DOTNAME_APPLICATION_PATH);
-        if (appPathAnno == null) {
+        if (appPathAnno == null || config.applicationPathDisable()) {
             appPathAnno = JandexUtil.getClassAnnotation(applicationClass, OpenApiConstants.DOTNAME_PATH);
         }
```
MikeEdgar: Is the intent that `@Path` on an `Application` instance would not be impacted by the new configuration?

MikeEdgar:
@EricWittmann - this looks good. I'll wait for your feedback on the one comment before closing the review. I noticed that the two new test classes have 2018 as the copyright year. I'm not sure if that's important so I just wanted to mention it.

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
EricWittmann: Yeah I wasn't sure about that.  But that's the decision I ended up going with.  What do you think?

EricWittmann:
I fixed the copyright issues.  Damn you copy/paste. :)

implementation/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
MikeEdgar: I think that's good. It keeps the configuration name in alignment with the annotation.

Issue #258, opened 2020-02-14
Using a Custom String Converter that does nothing makes startup fail


agentgonzo:
By adding a custom `Converter<String>` that does nothing (see https://github.com/agentgonzo/quarkus-quickstarts/commit/38d838cae650ad4014a3da5f35694539040b8c12) will cause startup to fail.

This is caused by the call here https://github.com/smallrye/smallrye-open-api/blob/359b8e3493ebff4662bafe7a54decc3730f3ecb4/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java#L217  which in turn calls https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java#L90 which passes the empty string `""` to the converter. The converter returns the empty string `""` and then `smallrye-open-api` tries to load the class with the name `""`.

This does _not_ happen if there are no custom converters as `null` is returned (and handled correctly)

The full stack trace on launch of https://github.com/agentgonzo/quarkus-quickstarts/commit/38d838cae650ad4014a3da5f35694539040b8c12 is:
```
10:25:20,755 INFO  [org.jbo.threads] JBoss Threads version 3.0.0.Final
10:25:21,145 ERROR [io.qua.dev.DevModeMain] Failed to start Quarkus: java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
        [error]: Build step io.quarkus.smallrye.openapi.deployment.SmallRyeOpenApiProcessor#build threw an exception: java.lang.RuntimeException: Failed to create instance of custom schema registry:
        at io.smallrye.openapi.runtime.scanner.OpenApiAnnotationScanner.getCustomSchemaRegistry(OpenApiAnnotationScanner.java:2160)
        at io.smallrye.openapi.runtime.scanner.OpenApiAnnotationScanner.scan(OpenApiAnnotationScanner.java:198)
        at io.quarkus.smallrye.openapi.deployment.SmallRyeOpenApiProcessor.generateAnnotationModel(SmallRyeOpenApiProcessor.java:300)
        at io.quarkus.smallrye.openapi.deployment.SmallRyeOpenApiProcessor.build(SmallRyeOpenApiProcessor.java:251)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #259, opened 2020-02-14
Handle custom `Converter`s returning "".

agentgonzo:
Fix for #258 

Custom `Converter<String>`s registered by users can cause errors when starting.
https://github.com/smallrye/smallrye-config/blob/1.6.1/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java#L90 _can_ pass in an empty string if `OpenApiConstants.CUSTOM_SCHEMA_REGISTRY_CLASS` is not specified. In this case, the Converter (correctly) returns null back to the caller and the caller will then try to load a class named `""`.
We should validate the empty string on return and treat it as `null`

### DIFF
--- a/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
+++ b/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
@@ -230,7 +230,7 @@ public class OpenApiConfigImpl implements OpenApiConfig {
      * getConfig().getOptionalValue(key) can return "" if optional {@link Converter}s are used. Enforce a null value if
      * we get an empty string back.
      */
-    private String getStringConfigValue(String key) {
+    String getStringConfigValue(String key) {
         return getConfig().getOptionalValue(key, String.class).map(v -> "".equals(v.trim()) ? null : v).orElse(null);
     }
 
--- a/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
+++ b/implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
@@ -231,7 +231,7 @@ public class OpenApiConfigImpl implements OpenApiConfig {
      * we get an empty string back.
      */
     private String getStringConfigValue(String key) {
-        return getConfig().getOptionalValue(key, String.class).map(v -> "".equals(v) ? null : v).orElse(null);
+        return getConfig().getOptionalValue(key, String.class).map(v -> "".equals(v.trim()) ? null : v).orElse(null);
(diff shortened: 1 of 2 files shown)


implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
```diff
@@ -228,6 +226,14 @@ public boolean applicationPathDisable() {
         return applicationPathDisable;
     }
 
+    /**
+     * getConfig().getOptionalValue(key) can return "" if optional {@link Converter}s are used. Enforce a null value if
+     * we get an empty string back.
+     */
+    private String getStringConfigValue(String key) {
+        return getConfig().getOptionalValue(key, String.class).map(v -> "".equals(v) ? null : v).orElse(null);
```
MikeEdgar: @agentgonzo - maybe also call `trim` on `v`? I think eliminating whitespace-only values makes sense here also.

MikeEdgar:
Thanks for the PR, I think this is a good idea, even in the cases there a `Converter` is not used and there is a white space property value.

EricWittmann:
I agree that a test would be great for this.  Seems like something that could be pretty easily tested. :)  If @agentgonzo (great github id) doesn't have time for that, no worries we can add it later.  I only have one question on the code that I will add as a comment.

implementation/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
```diff
@@ -228,6 +226,14 @@ public boolean applicationPathDisable() {
         return applicationPathDisable;
     }
 
+    /**
+     * getConfig().getOptionalValue(key) can return "" if optional {@link Converter}s are used. Enforce a null value if
+     * we get an empty string back.
+     */
+    private String getStringConfigValue(String key) {
+        return getConfig().getOptionalValue(key, String.class).map(v -> "".equals(v.trim()) ? null : v).orElse(null);
```
EricWittmann: This is my own ignorance here - but I assume that `map()` is only called if the value is non-null.  Otherwise the call to `trim()` will result in an NPE.

agentgonzo:
Yes, map only gets called if the optional is present. If it's empty, then
the map call is skipped and it just returns the orElse section.

I'm away for a few days. The test will be simple to implement, but I've not
worked in this repo before to don't know where to put it or how to set up
the test, given that I found the bug with a bunch of CDI injection.  Some
help in that area would be great

MikeEdgar:
@agentgonzo - I pushed a few test cases to your PR.

PR #281, opened 2020-04-02
Implement scan of APIResponseSchema and RequestBodySchema annotations

### LINKED ISSUE #76: Introduce a simpler version of ApiResponse annotation
OpenApi has an [ApiResponse annotation](https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/org/eclipse/microprofile/openapi/annotations/responses/APIResponse.java) which, via its nested Content, can provide the information about the JAX-RS Response entity class.

While this annotation covers well for a case where Response dynamically sets different entity types per the current response media type, it is too complex for a typical case of Response returning a known single entity type only, with @Produces already typing a response media type.

One proposal is to have something like this:
```
@Produces("application/json")
@ApiResponseType(MyType.class)
@GET
public Response get() { return Response.ok(new MyType()).build(); }   
```
but other options should also be considered

MikeEdgar:
Fixes #76 

This will not build until eclipse/microprofile-open-api#410 is merged, but I wanted to open the PR to make sure this is in alignment with the new structure and get any feedback. Tests all pass using my local build of the MP OAI TCK in that PR. I plan on going back and adding JavaDoc comments where they are missing on new methods.

This puts the Spring dependencies as "provided" to access the `HttpStatus` class to obtain the reason phrase. It seems like a lot of baggage just for that, but no other obvious solution occurred to me. What are everyone's thoughts on that?

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -82,7 +82,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
-                <version>3.2.0</version>
                 <executions>
                     <execution>
                         <goals>
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/requestbody/RequestBodyConstant.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/requestbody/RequestBodyConstant.java
@@ -1,6 +1,7 @@
 package io.smallrye.openapi.runtime.io.requestbody;
 
 import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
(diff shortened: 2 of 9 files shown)


MikeEdgar:
Thanks @phillip-kruger . I'll push the JavaDocs once the upstream changes are merged and available as a snapshot build.

phillip-kruger:
@MikeEdgar - let us know once you have added the javadoc and we can merge ! Thanks

MikeEdgar:
@phillip-kruger - absolutely. I am waiting on one more review in the spec change PR.

MikeEdgar:
Re-based on `master` and added a commit with the documentation.

PR #282, opened 2020-04-02
Set up GH Actions for Build and Release.

radcortez:


radcortez:
My pleasure. There is also some setup in the settings that needs to be done. If you don't mind I can take care of it. (setup secrets, branch protections, etc).

phillip-kruger:
Cool Thanks !

Issue #274, opened 2020-03-16
Additional properties generated in a header ref


michaeljmcd:
When I include a ref in a header, like the one below:

      headers = { @Header(ref = "my-header"), },

The resulting YAML generation looks like:

            my-header:
              $ref: my-header
              style: simple

The extra `style: simple` property then causes validation with Spectral to fail, with the error:

           error  no-$ref-siblings             $ref cannot be placed next to any other properties

Which at least broadly aligns with the OpenAPI spec (https://swagger.io/specification/#referenceObject) which says that "This object cannot be extended with additional properties and any properties added SHALL be ignored."

Looking through the code, I think the root cause is that `HeaderImpl.java` defaults the style to `SIMPLE` and `OpenApiSerializer.writeHeader()` doesn't exclude properties when `$ref` is set. It looks like adding a bit of logic to the serializer to skip over other properties when a ref is present would fix the issue. I'd be happy to open a PR if there is some agreement that there shouldn't be any siblings to `$ref`.

MikeEdgar:
+1

The JavaDoc for the `@Header` annotation's `ref` property states that if it is used in addition to other properties, the result is undefined. My opinion is that if the serializer encounters a `ref` as well as other properties, only the `ref` should be serialized. I think that works within the bounds of the specification. 

There are TCK tests that check for the default `simple` value, so I think that default needs to remain in place in `HeaderImpl`.

```java
    /**
     * Reference value to a Header object.
     * <p>
     * This property provides a reference to an object defined elsewhere. This property and
     * all other properties are mutually exclusive. If other properties are defined in addition
     * to the ref property then the result is undefined.
     * 
     * @return reference to a header
     **/
    String ref() default "";
```

michaeljmcd:
That's a good callout on the text in the javadoc description. A quick `grep` shows it to exist in MicroProfile OpenAPI in these classes:

    api/src/main/java/org/eclipse/microprofile/openapi/annotations/callbacks/Callback.java
    api/src/main/java/org/eclipse/microprofile/openapi/annotations/headers/Header.java
    api/src/main/java/org/eclipse/microprofile/openapi/annotations/links/Link.java
    api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/ExampleObject.java
    api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Schema.java
    api/src/main/java/org/eclipse/microprofile/openapi/annotations/parameters/Parameter.java
[...]

PR #286, opened 2020-04-06
Ref exclusiveness

michaeljmcd:
This is based on the discussion on #274 where `$ref` properties are generated alongside other properties, which can cause failures in other tools that consume OpenAPI. The idea is simple, if a `$ref` property is present, then no other properties are serialized.

At the time of writing, I have done this for two entities, `Header` and `Callback`. 

Is there interest in this change? If so, I will clean up the merge conflicts and do the remaining entities.


implementation/pom.xml
```diff
@@ -60,6 +60,11 @@
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-yaml</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.9</version>
+        </dependency>
```
phillip-kruger, line 67: Can we do this without pulling in this dep ?

implementation/src/main/java/io/smallrye/openapi/runtime/io/OpenApiSerializer.java
```diff
@@ -918,19 +919,23 @@ private void writeHeader(ObjectNode parent, Header model, String name) {
...
-        JsonUtil.booleanProperty(node, OpenApiConstants.PROP_ALLOW_EMPTY_VALUE, model.getAllowEmptyValue());
-        JsonUtil.enumProperty(node, OpenApiConstants.PROP_STYLE, model.getStyle());
-        JsonUtil.booleanProperty(node, OpenApiConstants.PROP_EXPLODE, model.getExplode());
-        writeSchema(node, model.getSchema(), OpenApiConstants.PROP_SCHEMA);
-        writeObject(node, OpenApiConstants.PROP_EXAMPLE, model.getExample());
-        writeExamples(node, model.getExamples());
-        writeContent(node, model.getContent());
-        writeExtensions(node, model);
 
+        if (StringUtils.isNotEmpty(model.getRef())) {
```
phillip-kruger, line 923: So something like model.getRef()!=null && !model.getRef().isEmpty() ?

implementation/pom.xml
michaeljmcd: Fair point, I've pushed an update that replaces the Apache Commons call with a quick and dirty helper method.

implementation/src/main/java/io/smallrye/openapi/runtime/io/OpenApiSerializer.java
michaeljmcd: Fair point, I've pushed changes to use a utility method rather than the Commons dependency.

michaeljmcd:
> 
> 
> Thanks @michaeljmcd ! LGTM. There is however some conflicts now :( Can you look at that, then we merge.

Sounds good. I've rebased against master and applied the same change to the other writer classes.

phillip-kruger:
Hi @michaeljmcd . Great ! Thanks. We have test failures now : https://github.com/smallrye/smallrye-open-api/pull/286/checks?check_run_id=571240433

Can you have a look ?

michaeljmcd:
Absolutely. Is there some setup I need to get the TCK tests to run? When I do a fresh clone of `smallrye-open-api` and do a `mvn clean install` against master, I get these same errors in the `PetStoreAppTest` suite.

MikeEdgar:
This is because my PR #281 is needed as of yesterday. Give me until tomorrow morning to finish the JavaDocs and push the update. Sorry

michaeljmcd:
I've merged master into this branch, pulling in the results of PR #281 and the checks are passing now.

PR #295, opened 2020-04-14
Spring pet store test case

### LINKED ISSUE #217: Allow spring-web extension REST endpoints to be published in openapi
Reported in quarkusio/quarkus#4693

Quarkus uses SmallRye OpenAPI to render API endpoints. As reported, if you create a Spring Web REST endpoint and add openapi, the Spring endpoint does not show up.

The OpenApiAnnotationScanner (or an extension) will need to discover Spring Rest endpoints.

ebullient:
This adds a test case based on the generated server classes for the Swagger Pet Store example.

This is not ready for integration yet, as the result is missing all kinds of information (and it does a few things slightly differently than other tools, which isn't a problem, but will require a custom "good" result json for testing once it all works.

Resolves #217

### DIFF
--- a/extension-spring/pom.xml
+++ b/extension-spring/pom.xml
@@ -57,9 +57,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>javax.validation</groupId>
-            <artifactId>validation-api</artifactId>
-            <version>2.0.0.Final</version>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+            <version>2.0.2</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/extension-spring/pom.xml
+++ b/extension-spring/pom.xml
@@ -56,7 +56,18 @@
             <version>2.5</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>javax.validation</groupId>
+            <artifactId>validation-api</artifactId>
+            <version>2.0.0.Final</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.github.joschi.jackson</groupId>
+            <artifactId>jackson-datatype-threetenbp</artifactId>
+            <version>2.6.4</version>
+            <scope>test</scope>
+        </dependency>
 
         <!-- Depend on core tests -->
         <dependency>
--- a/extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
+++ b/extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
@@ -20,6 +20,7 @@ import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.Type;
+import org.jboss.jandex.Type.Kind;
 import org.jboss.logging.Logger;
 import org.springframework.http.HttpStatus;
 
@@ -67,7 +68,7 @@ public class SpringAnnotationScanner implements AnnotationScanner {
 
     @Override
     public boolean isScannerInternalResponse(Type returnType) {
-        return returnType.name().equals(SpringConstants.RESPONSE_ENTITY);
+        return returnType.name().equals(SpringConstants.RESPONSE_ENTITY) && returnType.kind() != Kind.PARAMETERIZED_TYPE;
     }
 
     @Override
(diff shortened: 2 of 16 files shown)


extension-spring/pom.xml
```diff
@@ -56,7 +56,18 @@
             <version>2.5</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>javax.validation</groupId>
+            <artifactId>validation-api</artifactId>
+            <version>2.0.0.Final</version>
+            <scope>test</scope>
+        </dependency>
```
MikeEdgar, line 64: Please update this to use jakarta.validation:jakarta.validation-api:2.0.2

extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
```diff
@@ -67,7 +68,7 @@ public boolean isPostMethod(final MethodInfo method) {
 
     @Override
     public boolean isScannerInternalResponse(Type returnType) {
-        return returnType.name().equals(SpringConstants.RESPONSE_ENTITY);
+        return returnType.name().equals(SpringConstants.RESPONSE_ENTITY) && returnType.kind() != Kind.PARAMETERIZED_TYPE;
```
MikeEdgar, line 71: I think you might want to check for an instance of `HttpEntity` where the type argument is a wildcard, e.g. `HttpEntity<?>`. `ResponseEntity` extends `HttpEntity` so you will cover a few more scenarios.

See `OpenApiDataObjectScanner`'s static init block for an example of how to get the `ClassInfo` of types that cannot be assumed to be in the application's `IndexView` and also how to check if a type inherits from another using `TypeUtil#isA`.

Related to the type of response, you may also look at creating an implementation of `AnnotationScannerExtension` that is used in the Spring scanner with the `resolveAsyncType` overridden. You would want logic similar to that in `SchemaFactory#resolveAsyncType` to pull the type argument from the `HttpEntity<T>` or `ResponseEntity<T>`. The overridden method would then be called from `SchemaFactory#resolveAsyncType`.

extension-spring/src/test/java/io/smallrye/openapi/runtime/scanner/SpringAnnotationScannerTest.java
```diff
@@ -33,4 +33,23 @@ public void testBasicSpringDefinitionScanning() throws IOException, JSONExceptio
...
+        index(indexer, "test/io/smallrye/openapi/runtime/scanner/entities/Order.class");
+        index(indexer, "test/io/smallrye/openapi/runtime/scanner/entities/Pet.class");
+        index(indexer, "test/io/smallrye/openapi/runtime/scanner/entities/Tag.class");
+        index(indexer, "test/io/smallrye/openapi/runtime/scanner/entities/User.class");
+        OpenApiAnnotationScanner scanner = new OpenApiAnnotationScanner(emptyConfig(), indexer.complete());
+
+        OpenAPI result = scanner.scan();
+
+        printToConsole(result);
+        assertJsonEquals("resource.testBasicSpringDefinitionScanning.json", result);
```
MikeEdgar, line 52: Is there a different test JSON file that should be used here?

ebullient: I didn't bother to make one yet. ;)
I said this was a draft. What was being produced is still unique to smallrye (in terms of which default strings are used). I will make a json result when what we're producing is close enough that I can define what the correct result is.

extension-spring/pom.xml
ebullient: Sure. I was pulling from other projects.. not an issue.

extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
ebullient: I am not sure I understand what that "isScannerInternalResponse" is even supposed to do for Spring. JAX-RS Response is truly useless when it comes to figuring out response types. I'm tempted to always return false, to be honest.

MikeEdgar: You're right, but if the parameter is a wildcard, that's no better than the JAX-RS `Response`.

phillip-kruger: Yea I think for now return false, if we later figure out there is some class that is like Response then we update it.

phillip-kruger:
Hi @ebullient - should we retry this with the changes I made ? Let me know.

phillip-kruger:
Hi @ebullient - I want to see if this generates a better schema now, but I can not update the branch (here by clicking the button) . Do you want to look at this again ?

phillip-kruger:
No feedback. Closing.

PR #309, opened 2020-04-22
Baby steps: Adding support for ResponseEntity in Spring

phillip-kruger:

As discussed with @ebullient - baby steps to add Spring use cases. 
This PR only add support for ResponseEntity as a response type.

You will see there is a JAX-RS test case that mimic the Spring test case for comparison. The result of the JaxRs and Spring is exactly the same (so the json file we test against)

@ebullient @MikeEdgar - let me know what you think. After this we can do support for 
`@RequestMapping(method = RequestMethod.XXX)` as at the moment we only do `@XXXMapping`


ebullient:
LGTM. Thank you!

PR #311, opened 2020-04-26
Baby steps: Adding support for RequestMapping in Spring

phillip-kruger:
`@RequestMapping` working on GET as an alternative to `@GetMapping`
Basic `@PostMapping` and alternative `@RequestMapping` working

@ebullient @MikeEdgar - let me know what you think


extension-spring/src/main/java/io/smallrye/openapi/spring/ParameterProcessor.java
```diff
@@ -1036,6 +1034,11 @@ static String pathOf(AnnotationTarget target) {
                     path = methodAnnotation;
                 }
             }
+            // Also support @RequestMapping
+            AnnotationInstance methodAnnotation = target.asMethod().annotation(SpringConstants.REQUEST_MAPPING);
+            if (methodAnnotation != null && methodAnnotation.value() != null) {
+                path = methodAnnotation;
+            }
```
MikeEdgar, line 1041: Should this code also apply to classes with the `@RequestMapping` annotation?

ebullient: Y. @RequestMapping is a valid class-level annotation. Not sure that needs to be on this pass..

phillip-kruger: Yes maybe. We should create a test case for that then ...

phillip-kruger:
Merging this one, will create new tests for the class level case.

PR #317, opened 2020-04-30
Fixes #310, adds Mutiny types

### LINKED ISSUE #310: Add support for Mutiny types
It would be good to have OOTB support for `Uni` and `Multi` types from Mutiny.

See https://github.com/quarkusio/quarkus/issues/8811

kenfinnigan:


phillip-kruger:
Hi @kenfinnigan . I am pulling v2 of smallrye into Quarkus, and there is a test in IT that is failing that has got to do with this I think. see https://github.com/quarkusio/quarkus/pull/9695/checks?check_run_id=723450702#step:14:255

Can you point me in the right direction ?

kenfinnigan:
Probably log the JSON to see if the type is present.

I'd run this against Quarkus and it worked fine, so not sure what changed.

phillip-kruger:
Did you run both master and the 1.x branch against Quarkus ? This is testing the master (V2)

kenfinnigan:
I think I only did it with v1 because master was against MP spec SNAPSHOT at the time

phillip-kruger:
I think the test is actually "wrong" and it has been fixed by https://github.com/smallrye/smallrye-open-api/pull/286/files.

When you reference something, there should not be any other fields. I am "fixing" the test.

Issue #318, opened 2020-05-01
Implement i18n logging and messages


kenfinnigan:
Follow the procedures defined in https://github.com/smallrye/smallrye-parent/wiki/I18N-Messages-and-Logging as part of implementing https://github.com/smallrye/smallrye-parent/issues/102 for all SmallRye projects

An example usage can be found in https://github.com/smallrye/smallrye-common

mcserra:
Can i pick this one?

kenfinnigan:
Sure!

radcortez:
Nice to see Portuguese people around here :) Welcome!

mcserra:
I have some doubts on this requirements: 

> All interfaces should be package-private and not public

In the `runtime.io` package there are logging per sub-package which follow the same pattern, I'm creating one interface per package but I'm wondering if it wouldn't be better to create one public class in `runtime.io` and pass one parameter. One example:

`Processing a @Contact annotation.` and `Processing a @Components annotation.`

kenfinnigan:
In general it should be `package-private`, but if there's lots of sub packages you can use a single interface to prevent burning through groups of numbers for a couple of classes.

The main benefit is to prevent it from being usable as a "public API" of the project

PR #319, opened 2020-05-02
Add i18n logging and messaging [#318]

mcserra:


MikeEdgar:
Would it make sense for the interfaces and the `BasicLogger` contained therein to be named the same in each package? It might be nice to look in a package and always expect to see a `Messages` interface without much thought. Similarly all of the logging calls would look like `Messages.log.particularLogStatement(...)` where "Messages.log" is the same everywhere.

mcserra:
Makes sense. If being different from the other implementations is not a problem, I'll do it. Just let me know.

MikeEdgar:
It's really just a preference. We can wait and see if anyone else has thoughts on it. Also, to your point, it would be different from the other implementations.

MikeEdgar:
@EricWittmann, @phillip-kruger - one last question on this one. Is there anything that needs to be considered here regarding non-RH consumers of this library that have been discussed recently?

phillip-kruger:
Hi @MikeEdgar - nothing that I know of. Let's merge this, we can always do more PRs to add/change things.

PR #326, opened 2020-05-15
Add deep package scan capability with regular expression support

### LINKED ISSUE #289: Package deep scan logic changed since 1.1.3
I noticed there is a change how packages are scanned after upgrading from v1.1.3 to v1.2.1. 

1.1.3 would also scan all subpackages while 1.2.1 does not. I haven't found it in the release notes or closed issues, I am not sure which behaviour is correct, whether it is a bugfix or regression so I am raising the issue. I did not bisect the commits to find out exactly which version introduced the change.

MikeEdgar:
Fixes #289 

This is my attempt to implement what was discussed in eclipse/microprofile-open-api#422. Without using regex, it adds support for package matching using `startsWith`, which aligns SR more closely with other implementations. Beyond that, it gives some additional control using regex as suggested in the hangout on 4/30. If anything could use some additional comments I'm happy to add them to make this as clear as possible for future research/debugging.

If this looks good, it might be a good idea to apply these changes to `1.x` as well for immediate downstream use.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
+++ b/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
@@ -2,6 +2,7 @@ package io.smallrye.openapi.api;
 
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 /**
  * Accessor to OpenAPI configuration options.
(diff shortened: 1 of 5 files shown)


MikeEdgar:
Thanks @EricWittmann - merging!

PR #329, opened 2020-05-25
Baby steps: Adding support for PUT and DELETE in Spring

phillip-kruger:
This adds Tests cases for PUT and DELETE for Spring

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
@@ -36,6 +36,7 @@ import org.jboss.jandex.AnnotationTarget;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.ParameterizedType;
 import org.jboss.jandex.Type;
 
 import io.smallrye.openapi.api.constants.OpenApiConstants;
@@ -84,6 +85,8 @@ public interface AnnotationScanner {
     public boolean isAsyncResponse(final MethodInfo method);
 
     public boolean isPostMethod(final MethodInfo method);
+
+    public boolean isDeleteMethod(final MethodInfo method);
 
     public boolean isScannerInternalResponse(Type returnType);
 
@@ -347,7 +350,7 @@ public interface AnnotationScanner {
         final String code = String.valueOf(status);
         final String description = getReasonPhrase(status);
 
-        if (returnType.kind() == Type.Kind.VOID) {
+        if (isVoidResponse(method)) {
             if (generateResponse(code, operation)) {
                 response = new APIResponseImpl().description(description);
             }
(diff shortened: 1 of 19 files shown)


phillip-kruger:
@ebullient please also have a look. After this we can maybe try the pet store again ?

extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
```diff
@@ -127,6 +139,15 @@ public OpenAPI scan(final AnnotationScannerContext context, OpenAPI openApi) {
         return openApi;
     }
 
+    private boolean hasRequestMappingMethod(final MethodInfo method, final RequestMethod requestMethod) {
+        if (method.hasAnnotation(SpringConstants.REQUEST_MAPPING)) {
+            AnnotationInstance annotation = method.annotation(SpringConstants.REQUEST_MAPPING);
+            AnnotationValue value = annotation.value("method");
+            return value != null && value.asString().equals(requestMethod.name());
```
MikeEdgar, line 146: I think that `value.asEnumArray()` (which returns `String[]`) needs to be used for the `method` property here.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
```diff
@@ -417,9 +420,11 @@ default void createResponseFromRestMethod(final AnnotationScannerContext context
     default int getDefaultStatus(final MethodInfo method) {
         final int status;
 
-        if (method.returnType().kind() == Type.Kind.VOID) {
+        if (isVoidResponse(method)) {
             if (isPostMethod(method)) {
                 status = 201; // Created
+            } else if (isDeleteMethod(method)) {
+                status = 204; // No Content (Maybe this should be 202 Accepted ?)
```
MikeEdgar, line 427: There is wording in the spec about this now ([link](https://github.com/eclipse/microprofile-open-api/blob/54822cd7014a4c01a781b7f281527f9c85d028c3/api/src/main/java/org/eclipse/microprofile/openapi/annotations/responses/APIResponseSchema.java#L96-L105)), I'm not sure if checking for DELETE here might have an impact. It makes sense, and maybe we can get DELETE added to the MP spec as well.

extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/JaxRsAnnotationScannerBasicTest.java
```diff
@@ -52,4 +52,41 @@ public void testBasicJaxRsPostDefinitionScanning() throws IOException, JSONExcep
...
+     * 
+     * @throws IOException
+     * @throws JSONException
+     */
+    @Test
+    public void testBasicJaxRsPutDefinitionScanning() throws IOException, JSONException {
+        Indexer indexer = new Indexer();
+        index(indexer, "test/io/smallrye/openapi/runtime/scanner/resources/GreetingPutResource.class");
+        index(indexer, "test/io/smallrye/openapi/runtime/scanner/entities/Greeting.class");
+        OpenApiAnnotationScanner scanner = new OpenApiAnnotationScanner(emptyConfig(), indexer.complete());
```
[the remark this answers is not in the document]
phillip-kruger: Yes that is much better ! Thanks :)

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
phillip-kruger: Yea I can remove this as it will fall through to the default that is the same, but I think 202 might be a better default for delete ? I'll bring it up in the spec meeting.

extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
phillip-kruger: Cool ! Thanks :)

phillip-kruger:
Thanks @MikeEdgar ! Please check now, I made the changes as suggested. Let me know! Thanks

PR #334, opened 2020-05-27
Added UI (Swagger UI) module

phillip-kruger:


phillip-kruger:
Not really, but they can use it. It can be used by developers directly or by Wildfly and Quarkus. I might pull it into Quarkus and replace the webjars version of swagger ui

EricWittmann:
Hm - maybe I'll submit an option to enable ReDoc instead of Swagger UI. :)  ReDoc doesn't have the interactivity features of Swagger UI, but it looks a lot better.

phillip-kruger:
We can make the ui module a pom module and add both... ? w.d.y.t ?

PR #338, opened 2020-05-28
Add Support for MP OpenAPI 2.0 Schema Config to 1.x

### LINKED ISSUE #335: Add support for mp.openapi.schema.FQCN to 1.x
Add support for mp.openapi.schema.FQCN to 1.x

This should help with situations like what is described in quarkusio/quarkus#8810

MikeEdgar:
Fixes #335 

Allows users to provide custom schema for a fully-qualified class name using Config.


implementation/src/main/java/io/smallrye/openapi/runtime/io/OpenApiParser.java
```diff
@@ -483,7 +502,7 @@ private Components readComponents(JsonNode node) {
      * 
      * @param node
      */
-    private Schema readSchema(JsonNode node) {
+    private SchemaImpl readSchema(JsonNode node) {
```
phillip-kruger, line 505: Do you have to use the impl ? No way we can just use the interface ?

phillip-kruger: OK because you have to set the name ?

MikeEdgar: I did that in the past for handy access to `getName`.

PR #341, opened 2020-05-29
Feature/parse schema type attribute

### LINKED ISSUE #300: Example object in Schema is always of String type
I have a very simple User object where I have `id` field of type integer. I added example with specifying the type
``` java
public class User {

    @Schema(type = SchemaType.INTEGER, example = "1000")
    private Integer id;
... constructor, getters, and setters omitted
}
```
This generates a schema with example as a string
```
    User:
      type: object
      properties:
        id:
          format: int32
          type: integer
          example: "1000"
```
Because of this in e.g Swagger UI when I'm prompted to fill in the example I have string type, users can think that it's mandatory to wrap integer into double-quotes.
While debugging I've found that smallrye-open-api reads value from annotation but doesn't do anything with it. The line below `readAttr()` outputs a String object, which should be of type Integer 
https://github.com/smallrye/smallrye-open-api/blob/143d798943cdf3319790eed1e68d327d64728484/core/src/main/java/io/smallrye/openapi/runtime/io/schema/SchemaFactory.java#L196

Screen Shot 2020-04-17 at 12 15 51

dmi3coder:
Fixes #300 
Parses Example value to the identified type if Schema is not String, returns String object otherwise


phillip-kruger:
Hi @dmi3coder - Thanks for this, unfortunately the build fails. You might have to compile the full app on your PC before commit, so that the formatting can happen. (I think)

phillip-kruger:
(Also maybe update branch)

phillip-kruger:
:) Ok you done it

dmi3coder:
@phillip-kruger yes, I accidentally pushed without one space :)

core/src/main/java/io/smallrye/openapi/runtime/io/schema/SchemaFactory.java
```diff
@@ -274,6 +276,31 @@ public static Schema readSchema(IndexView index,
...
 
+    /**
+     * Reads the attribute named by propertyName from annotation, and parses it to identified type. If no value was specified,
+     * an optional default value is retrieved from the defaults map using the propertyName as
+     * they key. Array-typed annotation values will be converted to List.
+     *
+     * @param <T> the type of the annotation attribute value
+     * @param annotation the annotation to read
+     * @param propertyName the name of the attribute to read
+     * @param defaults map of default values
```
dmi3coder, line 287: Forgot to add `schemaType` param here, will fix it shortly

MikeEdgar:
@dmi3coder - this looks good, thank you. I have a couple of comments -
- Can you re-base your feature on the current master branch? It makes for a cleaner merge and commit history.
- If you have time, would you mind modifying one of the existing test scenarios to include a non-string example?

Let us know, thanks

dmi3coder:
@MikeEdgar yes sure, I'll do a rebase and add the test.
Can you maybe suggest a good place to put additional test cases for this feature? I'll have a look tomorrow otherwise for a place to put them.

phillip-kruger:
Hi @dmi3coder @MikeEdgar  I am merging this, we can add tests in a follow up PR. Thanks :)

PR #352, opened 2020-06-01
Fix NPE w/interface inheritance, Improve handling of java.util.Set types

### LINKED ISSUE #349: Annotation scanner NPE when processing JAX-RS endpoints with params of type SortedSet
Hi, I've noticed that some RESTEasy tests failing against Wildfly when declaring REST resources methods accepting parameters of type `SortedSet`. 
Maybe I am missing something but looks like this is allowed by JAX-RS spec: https://docs.oracle.com/javaee/7/api/javax/ws/rs/FormParam.html 

The following stack trace represents the output of a deployment failure when an attempt to deploy a WAR with such a resource:

```
Caused by: java.lang.NullPointerException
	at io.smallrye.openapi.runtime.scanner.ParameterProcessor.setSchemaProperties(ParameterProcessor.java:637)
	at io.smallrye.openapi.runtime.scanner.ParameterProcessor.getFormBodyContent(ParameterProcessor.java:605)
	at io.smallrye.openapi.runtime.scanner.ParameterProcessor.process(ParameterProcessor.java:404)
	at io.smallrye.openapi.runtime.scanner.OpenApiAnnotationScanner.processJaxRsMethod(OpenApiAnnotationScanner.java:577)
	at io.smallrye.openapi.runtime.scanner.OpenApiAnnotationScanner.lambda$processJaxRsResourceClass$0(OpenApiAnnotationScanner.java:422)
...
```

I've isolated a reproducer [1] which has three methods to demonstrate the failure: the first and the second ones (namely `formAsListOfString` and `formAsSetOfString`) take a `List` and a `Set` and are working fine, while the last one - `formAsSortedSetOfString` -, taking a `SortedSet`, fails. 
The methods are using `@FormParam` because it was the first case that I spotted but I've seen failures when using the same data type with `@QueryParam` and `@MatrixParam` as well.

The reproducer should fail by default. For a successful deploy the `formAsSortedSetOfString` method should be commented.

[1]
https://github.com/fabiobrz/microprofile-openapi-example/tree/sorted-set-repro

MikeEdgar:
Final of three PRs for #349
Fixes #349

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
@@ -71,6 +71,9 @@ public class OpenApiDataObjectScanner {
     // Map
     public static final DotName MAP_INTERFACE_NAME = DotName.createSimple(Map.class.getName());
     public static final Type MAP_TYPE = Type.create(MAP_INTERFACE_NAME, Type.Kind.CLASS);
+    // Set
+    public static final DotName SET_INTERFACE_NAME = DotName.createSimple(java.util.Set.class.getName());
+    public static final Type SET_TYPE = Type.create(SET_INTERFACE_NAME, Type.Kind.CLASS);
     // Enum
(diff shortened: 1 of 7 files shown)


phillip-kruger:
Thanks @MikeEdgar !

PR #364, opened 2020-06-09
Update ParameterProcessor.java

hubinix:
We want to input JWT(JsonWebToken) in http header,from swagger-ui 's page.
But isIgnoredParameter function ingnored the AUTHORIZATION parameter.
The AUTHORIZATION parameter can not be seen from the swagger-ui , and we can not from the "try it out " button.


phillip-kruger:
Closing.

Issue #376, opened 2020-06-23
@Extension on pathParam is added to operation


tkalmar:
When annotating an `@Param` with an `@Extension` the extension is generated at the wrong position in the yaml and json
adding
```
   @GET
    @Path("pathParamExtension/{param}")
    public String getParamWithExtension(@PathParam("param") @Extension(name = "x-operation-extension-3", value = "Operation extension wrapper value (3).") String param) {
        return "Path param extension value.";
    }
```
to `io.smallrye.openapi.tck.extra.extensions.ExtensionResource` and then 
```
     @RunAsClient
        @Test(dataProvider = "formatProvider")
        public void testPathParamExtensions(String type) {
            ValidatableResponse vr = this.callEndpoint(type);
            String operation = "paths.'/extensions/pathParamExtension/{param}'.get";
            vr.body(operation + ".x-parameter-extension", nullValue());
            vr.body(operation + ".parameters[0].x-parameter-extension", equalTo("Operation extension wrapper value (3)."));
        }
```
to `io.smallrye.openapi.tck.extra.ExtensionsTest` demonstrates the wrong behaviour.

tkalmar:
cc @famod

PR #377, opened 2020-06-23
Fix for #376

tkalmar:
this is a first try at fixing #376

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
@@ -1,5 +1,6 @@
 package io.smallrye.openapi.runtime.scanner.spi;
 
+import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -68,6 +69,8 @@ import io.smallrye.openapi.runtime.scanner.processor.JavaSecurityProcessor;
 import io.smallrye.openapi.runtime.util.JandexUtil;
 import io.smallrye.openapi.runtime.util.ModelUtil;
 import io.smallrye.openapi.runtime.util.TypeUtil;
+
+import static org.jboss.jandex.AnnotationTarget.Kind.METHOD_PARAMETER;
 
 /**
  * This represent a scanner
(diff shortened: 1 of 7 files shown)


phillip-kruger:
@tkalmar Thanks for this. I think something went wrong with you PR, it shows 300+ file changes ?

tkalmar:
@phillip-kruger reformat the PR so only the new files are actualy new ... now it even builds

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
```diff
@@ -628,7 +631,15 @@ default void processExtensions(final AnnotationScannerContext context, final Met
         }
         for (AnnotationInstance annotation : extensionAnnotations) {
             String name = ExtensionReader.getExtensionName(annotation);
-            operation.addExtension(name, ExtensionReader.readExtensionValue(context, name, annotation));
+            Object value = ExtensionReader.readExtensionValue(context, name, annotation);
+            if (annotation.target() != null && METHOD_PARAMETER.equals(annotation.target().kind())) {
+                final MethodParameterInfo methodParameterInfo = annotation.target().asMethodParameter();
+                final short paramPosition = methodParameterInfo.position();
+                operation.getParameters().get(paramPosition).addExtension(name, value);
```
MikeEdgar, line 638: I don't think this approach is going to work outside of the basic scenario. I.e., the position in the operation's parameters may not always be the same as the position in the method parameters. Think of the case where `@BeanParam` is used or where parameters are injected into the resource class itself rather than as method arguments.

phillip-kruger: @MikeEdgar - Any suggestion on how @tkalmar can change the PR so that this will work better ?

MikeEdgar:
@tkalmar , @phillip-kruger  - I think a good approach would look something like this. I haven't tested this out :-)

1. `io.smallrye.openapi.runtime.io.extension.ExtensionReader`: overload `readExtensions` method, pass `AnnotationScannerContext` and `List<AnnotationInstance>`. Returns a `Map<String, Object>`, same as existing `readExtensions` by iterating over the `List` and creating a map entry for each list entry.
2. Pass `AnnotationScannerContext` from `JaxRsAnnotationScanner` to `ParameterProcessor` and set as an instance variable (the `Index` parameter can be removed and set the instance variable from the `AnnotationScannerContext` in the `ParameterProcessor` constructor. The calls from `JaxRsAnnotationScanner` can simply pass the `context` instead of `context.getIndex()`.
3. Read the extensions in `Parameter` processor after the code for `setDeprecated`. Something like below. In this example, `this.scannerContext` is the new instance field from change (2) above.
```java
List<AnnotationInstance> extensionAnnotations = ExtensionReader.getExtensionsAnnotations(context.target);

if (param.getExtensions() == null && !extensionAnnotations.isEmpty()) {
    param.setExtensions(ExtensionReader.readExtensions(this.scannerContext, extensionAnnotations));
}
```

tkalmar:
@phillip-kruger @MikeEdgar Thanks for the help, i extended the test to include one @BeanParam and it broke as expected. With the suggested changes it was a piece of cake to do the required changes. Test works now and is better than my first draft. Thanks a lot.

tkalmar:
Should this be backported at the 1.x branch to get it faster into quarkus or are there plans to switch quarkus to the 2.x branch in the near future?

phillip-kruger:
Quarkus is on 2.x  starting 1.6. we can try and get it in a.s.a.p. no need to backport.

PR #378, opened 2020-06-23
Set required = true in ParameterImpl::setIn when in == In.PATH

### LINKED ISSUE #372: Request: ParameterImpl::setIn should set required = true for In.PATH
## Overview

The current implementation of the `ParameterImpl::setIn` method simply sets the value of the **in** attribute, as follows:

```
@Override
public void setIn(In in) {
    this.in = in;
}
```
However, the Open API Specification states the following with relation to the **required** property on the **Parameter Object**:

> Determines whether this parameter is mandatory. If the parameter location is "path", this property is REQUIRED and its value MUST be true. Otherwise, the property MAY be included and its default value is false.

As a result, consumers of the SmallRye implementation of OpenAPI must explicitly set **required** to `true` on a path `Parameter`... which requires more code.

It would be more convenient to update the implementation as follows:

```
@Override
public void setIn(In in) {
    if (in == In.PATH) {
        this.required = true;
    }
    this.in = in;
}
```

msmiths:
This changes simply sets the `required` attribute on `ParameterImpl` to true in the `setIn` method if the input argument equals In.PATH.

Resolves #372

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/models/parameters/ParameterImpl.java
+++ b/core/src/main/java/io/smallrye/openapi/api/models/parameters/ParameterImpl.java
@@ -275,6 +275,9 @@ public class ParameterImpl extends ExtensibleImpl<Parameter> implements Paramete
      */
     @Override
     public void setIn(In in) {
+        if (in == In.PATH) {
+            this.required = true;
+        }
         this.in = in;
     }
 


phillip-kruger:
Thanks @MikeEdgar and @msmiths !!

Issue #374, opened 2020-06-19
@JsonIgnoreProperties ignored on superclass


tkalmar:
We have a quarkus (1.4.2.Final) project with  smallrye-openapi in version 1.2.3.

We have an abstract Superclass

```
import com.fasterxml.jackson.annotation.JsonIgnoreProperties;

@JsonIgnoreProperties(value = {"field1"})
public abstract class SuperClass {

abstract getField1() {
}
abstract setField1(String value) {
}
}
```
and some Child classes
```
public final class ChildClass {

@Override
protected getField1(){
...
}
@Override
protected setField1(String value){
...
}
}
```

In swagger-ui we see field1 in all subclasses of Superclass. If we put the `@JsonIgnoreProperties(value = {"field1"})` on the child classes all is fine and the field is ignored.

Same for `@JsonIgnore` or `@Schema(hidden=true)` all work at child class level but are ignored if used on super class.

It is cumbersome to always remember to annotate each new child, so it would be great if this works on super class level

tkalmar:
cc @famod

PR #385, opened 2020-07-02
Test and fix inheritance propagation of @JsonIgnoreProperties

tkalmar:
After fixing  #374 a colleague pointed out, that this fix isn't  working. So i sat down and analyzed the problem. Only investigating the superclass isnt sufficient one has to inspect the whole class hierarchy to get the same behaviour as jackson. I ensured the upstreambehaviour with an test and fixed the missing ascension up the inheritance tree.
I can leave out the upstream behaviuor test, but i think it is saver to get an notice if upstream changes behaviour.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
@@ -209,10 +209,8 @@ public class OpenApiDataObjectScanner {
 
     // Scan depth first.
     private void depthFirstGraphSearch() {
-        DataObjectDeque.PathEntry currentPathEntry;
-
         while (!objectStack.isEmpty()) {
-            currentPathEntry = objectStack.pop();
+            DataObjectDeque.PathEntry currentPathEntry = objectStack.pop();
 
             ClassInfo currentClass = currentPathEntry.getClazz();
             Schema currentSchema = currentPathEntry.getSchema();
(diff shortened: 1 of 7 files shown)


MikeEdgar:
@tkalmar , @phillip-kruger - thinking more about this issue, I am wondering if a better approach would be to check whether the properties are ignored in the `TypeResolver` as the hierarchy is processed there. I'm still thinking through that, I'll reply with some more thoughts on that approach if you agree.

MikeEdgar:
This is where I landed: https://github.com/MikeEdgar/smallrye-open-api/commit/8cb217c84a44ab54c8061949dc64921b072f0533, which seems to work with @tkalmar's original test case in the TCK project. I haven't incorporated the test changes from this PR, however. I can cherry pick the test changes from this PR if this seems like the correct approach.

tkalmar:
> 
> 
> This is where I landed: [MikeEdgar@8cb217c](https://github.com/MikeEdgar/smallrye-open-api/commit/8cb217c84a44ab54c8061949dc64921b072f0533), which seems to work with @tkalmar's original test case in the TCK project. I haven't incorporated the test changes from this PR, however. I can cherry pick the test changes from this PR if this seems like the correct approach.

This would help i think. I run my new set of tests against your tree and one is failing, i think it is the same problem as in my initial solution the direct descendant case is solved but deeper nesting is not looked at.

MikeEdgar:
@tkalmar - I've applied your commit to my branch and made a few updates based on your feedback. If you don't mind, I will go ahead and push it to your `jsonignoreproperties` branch and we can merge this PR.

tkalmar:
@MikeEdgar  LGMT

PR #393, opened 2020-07-15
Use nested type param names in schema name, resolve types earlier

### LINKED ISSUE #388: Wrong generation on generic model parametrized by a collection
Hi!

My Quarkus app has a wrapper for an API response with a generic field. 
I noticed that if my endpoint returns something like `Response<List<Cat>>`, my openapi model will look like this:
```json
      "ListCat" : {
        "type" : "object",
        "properties" : {
          "empty" : {
            "type" : "boolean"
          }
        }
      },
```
Expected result: 
```json
      "ListCat" : {
        "type" : "array",
        "items" : {
          "$ref" : "#/components/schemas/Cat"
        }
      },
```

After some reverse engineering, I added this test case (as an example) to 'GenericModelTypesResourceTest': https://github.com/egorklimov/smallrye-open-api/commit/8c329f456d5b28d4915cbc2e12dbdfc2a83ccf04

Test result is the same:
```json
      "ListPOJO" : {
        "type" : "object",
        "properties" : {
          "empty" : {
            "type" : "boolean"
          }
        }
      },
```

Could you please confirm that this is actually a bug?

P.S.
Also, I noticed that while processing a field of type `List<String>` in  `SchemaRegistry#checkRegistration`  it has type `array`, whereas when processing a generic field parametrized by List<POJO> (`type.toString()` -> `"T"`) it has type `object`.

MikeEdgar:
Fixes #388 

Creating this PR as a draft so it can be re-based after the other open PRs. The changes here impact the naming of schemas under `#/components/schemas/...` - not sure if that merits a bump to the minor version or not.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/SchemaRegistry.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/SchemaRegistry.java
@@ -244,6 +244,19 @@ public class SchemaRegistry {
      * @return a reference to the newly registered {@link Schema}
      */
     private Schema register(TypeKey key, Schema schema, String schemaName) {
+        String name = deriveName(key, schemaName);
+        Schema schemaRef = new SchemaImpl();
+        schemaRef.setRef(OpenApiConstants.REF_PREFIX_SCHEMA + name);
+
+        registry.put(key, new GeneratedSchemaInfo(name, schema, schemaRef));
+        names.add(name);
+
+        ModelUtil.components(oai).addSchema(name, schema);
+
+        return schemaRef;
+    }
+
+    String deriveName(TypeKey key, String schemaName) {
         /*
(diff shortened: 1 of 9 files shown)


MikeEdgar:
@egorklimov - this pull request includes the commit you authored and referenced in #388. Please confirm that is not a problem.

MikeEdgar:
@phillip-kruger - this is ready for review, but I was thinking we should wait to merge until we hear back from @egorklimov regarding his potential contribution. Perhaps we can do a release after this change also.

egorklimov:
Wow, thank you ;)

I could rename the endpoint ("/noo...") from test, if you want, also I changed the generic type boundary in `Result` and `Result Builder`, is it ok?

MikeEdgar:
@egorklimov - I think those things are all fine (the names and boundaries).

egorklimov:
LGTM

Issue #392, opened 2020-07-13
Add support for @SchemaProperty (to pass AirlinesAppTckTest from the TCK)


jmini:
Currently the `AirlinesAppTckTest` is ignored:

https://github.com/smallrye/smallrye-open-api/blob/669d071521e0c7b1a4f0053c36111c06daf62839/tck/src/test/java/io/smallrye/openapi/tck/AirlinesAppTckTest.java#L30-L34

Because SmallRye is used to validate that the TCK is good prior to MicroProfile release, it would be good to add support for `@SchemaProperty`.

---

If you remove the `@Ignore` annotation from the `AirlinesAppTckTest`, the build is failing with:

```
[ERROR] Failures: 
[ERROR]   AirlinesAppTest.testSchemaProperty:671 1 expectation failed.
JSON path components.schemas.User.properties.phone.description doesn't match.
Expected: Telephone number to contact the user
  Actual: null

[ERROR]   AirlinesAppTest.testSchemaProperty:671 1 expectation failed.
JSON path components.schemas.User.properties.phone.description doesn't match.
Expected: Telephone number to contact the user
  Actual: null

[INFO] 
[ERROR] Tests run: 274, Failures: 2, Errors: 0, Skipped: 0
```

jmini:
The `@SchemaProperty` support is only relevant for MicroProfile OpenAPI `2.0` support.

I have opened a PR to un-ignore the test on `master` which is targeting MP `1.x`: https://github.com/smallrye/smallrye-open-api/pull/401

---

For the MP `2.0` you can follow progress on my branch: [mp-oa-2.0](https://github.com/jmini/smallrye-open-api/tree/mp-oa-2.0). @msmiths is discussing on gitter how to add this `@SchemaProperty` support.

PR #401, opened 2020-07-16
Restore AirlinesAppTckTest

jmini:
Related to issue #392.

One `master` (MicroProfile `1.x`) the test is passing. This commits proves it and can be merged.

For MicroProfile `2.0` compliance, support for `@SchemaProperty` must be added, in order to pass this test.

### DIFF
--- a/tck/src/test/java/io/smallrye/openapi/tck/AirlinesAppTckTest.java
+++ b/tck/src/test/java/io/smallrye/openapi/tck/AirlinesAppTckTest.java
@@ -17,7 +17,6 @@
 package io.smallrye.openapi.tck;
 
 import org.eclipse.microprofile.openapi.tck.AirlinesAppTest;
-import org.junit.Ignore;
 
 import test.io.smallrye.openapi.tck.BaseTckTest;
 import test.io.smallrye.openapi.tck.TckTest;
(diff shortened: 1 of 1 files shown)


jmini:
I must admit I am not familiar with how your `@TckTest` and `BaseTckTest<T>` works, but if we let the method empty it will be an empty test.

By comparison have a look at PetStoreAppTckTest:

https://github.com/smallrye/smallrye-open-api/blob/edefd879b6b04ab92caad274fb7f1b978c0e002b/tck/src/test/java/io/smallrye/openapi/tck/PetStoreAppTckTest.java#L27-L30

This is also empty.

Github actions does not seems to keep or to display tests result. But when I run locally I have a file:
`tck/target/surefire-reports/TEST-org.eclipse.microprofile.openapi.tck.AirlinesAppTest.xml`

With:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report.xsd" name="org.eclipse.microprofile.openapi.tck.AirlinesAppTest" time="13.718" tests="90" errors="0" skipped="0" failures="0">
  <properties>
...
```

MikeEdgar:
This change is correct. Once the pom is updated to MP 2.0 with @jmini 's other branch, this test will start to fail until SchemaProperty is implemented.

jmini:
> Once the pom is updated to MP 2.0 with @jmini 's other branch, this test will start to fail until `@SchemaProperty` is implemented.

Which is all good. 

If SmallRye is used to validate that the MP-OpenAPI Tck is valid, then we need to have it implemented. (personally I have no time to look at it).

PR #412, opened 2020-07-24
Allow setting version and other info elements

### LINKED ISSUE #298: Request: option to set the OpenAPI version
This comes from someone on the Microprofile OpenAPI gitter chat room:

> 
> Franck Arnulfo
> @farnulfo_twitter
> Hi, is there anyway to have a "dynamic backport of OpenApi version" ?
> Context: we use Thorntail v2.6.0 which implement OpenAPI 3.0.1 so /openapi return a 3.0.1 document but our API Gateway (WSO2) only support 3.0.0 so we can't use the /openapi endpoint to configure the API.  We can manually import an OpenApi 3.0.0 by editing and modifying the dynamic 3.0.1 but I would like to avoid it.  So I was thinking about a way to have the /openapi endpoint to output a back version of the current one. Something like /openapi?version=3.0.0.

phillip-kruger:
This PR adds the following:

- Allow the user to set the version (openapi) by providing a schema file
- Allow the user to set the version (openapi) via MP Config (mp.openapi.extensions.openapi)
- Up the default version to 3.0.3
- Allow the user to set all elements in info via MP Config.

Fixes #298


core/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
```diff
@@ -20,6 +20,17 @@
...
+    private static final String VERSION = OASConfig.EXTENSIONS_PREFIX + "openapi";
+    private static final String INFO_TITLE = OASConfig.EXTENSIONS_PREFIX + "info.title";
+    private static final String INFO_VERSION = OASConfig.EXTENSIONS_PREFIX + "info.version";
+    private static final String INFO_DESCRIPTION = OASConfig.EXTENSIONS_PREFIX + "info.description";
+    private static final String INFO_TERMS = OASConfig.EXTENSIONS_PREFIX + "info.termsOfService";
+    private static final String INFO_CONTACT_EMAIL = OASConfig.EXTENSIONS_PREFIX + "info.contact.email";
+    private static final String INFO_CONTACT_NAME = OASConfig.EXTENSIONS_PREFIX + "info.contact.name";
+    private static final String INFO_CONTACT_URL = OASConfig.EXTENSIONS_PREFIX + "info.contact.url";
+    private static final String INFO_LICENSE_NAME = OASConfig.EXTENSIONS_PREFIX + "info.license.name";
+    private static final String INFO_LICENSE_URL = OASConfig.EXTENSIONS_PREFIX + "info.license.url";
```
MikeEdgar, line 32: These should probably include `smallrye` in the name. Take a look in `io.smallrye.openapi.api.constants.OpenApiConstants` for the others.

phillip-kruger: Ok cool thanks. I'll actually move the constants to `io.smallrye.openapi.api.constants.OpenApiConstants` as well.

phillip-kruger:
Thanks @MikeEdgar !

Issue #404, opened 2020-07-16
AnnotationScannerFactory is hard coded to load AnnotationScanner instances from the TCCL.


pferraro:
OpenApiProcessor.modelFromAnnotations(...) creates a new OpenApiAnnotationScanner instance which currently loads AnnotationScanner instances from the TCCL.  Given that smallrye has no control which thread will call this method, it cannot assume that the intended annotation scanner instances (or indeed, if any) will be loadable from the TCCL.

Ideally OpenApiProcessor.modelFromAnnotations(...) should accept a ClassLoader instance from which to load AnnotationScanner instances.

pferraro:
@phillip-kruger I don't think this fix is sufficient.  AnnotationScannerFactory still loads AnnotationScanner instances using the TCCL, instead of the classloader passed to the OpenApiProcessor.modelFromAnnotations(...) method.  See:
https://github.com/smallrye/smallrye-open-api/blob/master/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScannerFactory.java#L19

phillip-kruger:
Thanks, I'll look again in the morning

PR #419, opened 2020-07-28
Load AnnotationScanner instances from the designated classloader.

pferraro:
Fixes https://github.com/smallrye/smallrye-open-api/issues/404


phillip-kruger:
@pferraro - how urgent is this for you ? Do you need a release ?

pferraro:
It's not super urgent since it has a workaround, i.e. setting the TCCL explicitly.

PR #424, opened 2020-07-30
Read interface types recursively while reading a type's properties

### LINKED ISSUE #423: Support multi-level interface inheritance
I have an inheritance hierarchy like Note<-extends-MutableEntity<-extends-ImmutableEntity and the documentation generated for Note does not include properties from ImmutableEntity. From looking at the code, it appears to only scan parent interfaces requiring Note to extend both interfaces.

MikeEdgar:
Fixes #423 

CC: @gtnicol


gtnicol:
Looks good. Thanks for the quick turnaround.

PR #427, opened 2020-07-30
Fix partial schema registration

### LINKED ISSUE #392: Add support for @SchemaProperty (to pass AirlinesAppTckTest from the TCK)

### LINKED ISSUE #403: Errors in enums generated in inline schemas
## Overview

Generating an inline schema for a class that contains a property whose type is an enum results in the generation of multiple refs in the OpenAPI document.  Both the enum definition and the property on the referencing class need to be annotated with the `@Schema` annotation.

You can see this in the `AirlinesAppTckTest`.

The **Gender** enum is defined as follows:

```
@Schema(enumeration = {"Male", "Female", "Other"})
public enum Gender {
    MALE, FEMALE, OTHER
}
```

The **User** class defines a `gender` property of type `Gender` that is also annotated with a `@Schema` annotation as follows:
```
    @Schema(required = true, example = "M")
    private Gender gender;
```
When the `User` class is referenced by a `@Schema` annotation on the `UserResource::createUser` method, it results in an inline schema being generated, but the definition of the `gender` property is as follows:
```
  gender:
    allOf:
    - $ref: '#/components/schemas/Gender'
    - example: M
    - $ref: '#/components/schemas/Gender'
    - example: M
    - $ref: '#/components/schemas/Gender'
    - example: M
    - $ref: '#/components/schemas/Gender'
    - example: M
```
The relevant annotation on the `UserResource::createUser` method is as follows:

```
    @RequestBody(
        ...
        content = @Content(
            mediaType = "application/json",
            schema = @Schema(
                name = "testUser",
                type = SchemaType.OBJECT,
                maxProperties = 1024,
                minProperties = 1,
                required = true,
                implementation = User.class
            ),
        ...
```

MikeEdgar:
Fixes #392 

May fix #403, being researched by @msmiths 
POMs updated for `2.1.x`.
After this merge, a new `2.0.x` maintenance branch will be created for changes needed prior to the MP 2.0 release.

CC: @jmini , @msmiths


MikeEdgar:
This still needs a little work.

PR #434, opened 2020-08-15
Consolidate common parameter processing into abstract core class

MikeEdgar:
Also use Java 11 and deep git clone/fetch for Sonar scan


phillip-kruger:
Thanks !! This is awesome !!

Issue #437, opened 2020-08-19
Ignore logic regression


phillip-kruger:
See https://github.com/RedHatInsights/policies-ui-backend/pull/144

PR #438, opened 2020-08-20
Use default value of @Schema#hidden(), handle ignore set/get separately (master)

MikeEdgar:
Relates to #437


phillip-kruger (approved):
LGTM

PR #445, opened 2020-08-25
Maven plugin for SmallRye Open API

### LINKED ISSUE #313: Create Maven/Gradle plugin to generate openapi schemafiles
Similar as this plugin for SmallRye GraphQL: https://github.com/smallrye/smallrye-graphql/issues/191

This is useful for storing the generated schema e.g. in git to track changes or share the generated schema across different teams.

phillip-kruger:
Fix #313

This plugin can create the schema document on build.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
+++ b/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
@@ -136,4 +136,38 @@ public interface OpenApiConfig {
         PACKAGE_CLASS_METHOD
     }
 
+    default Pattern patternOf(String configValue) {
+        return patternOf(configValue, null);
+    }
+
+    default Pattern patternOf(String configValue, Set<String> buildIn) {
+        Pattern pattern = null;
+
+        if (configValue != null && (configValue.startsWith("^") || configValue.endsWith("$"))) {
+            pattern = Pattern.compile(configValue);
+        } else {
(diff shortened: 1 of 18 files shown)


phillip-kruger:
I removed them from the scanners, as that cause problems. If I can get the plugin to "know" what frameworks is available I can scan better and the problem should go away, but I am not sure how to do that. We don't really need them in the scanner, apart from a few static vars. So for now this will do ?

jglink:
The Plugin works great.
One minor issue: the destination type (JSON | YAML) should be configurable; currently it´s hard-coded to YAML.

But the OpenAPI generation still needs an application class for the JAX-RS annotations to get scanned. If I only have an interface with JAX-RS annotations (implementation is in another module), the OpenAPI documentation is basically empty.

You can find a small reproducer project here: https://github.com/jglink/openapi-doc-from-interface
The project only has 1 interface class (my.project.api.SomeApi) with a JAX-RS annotation and so the generated openapi.yaml only defines some basic stuff.

All the OpenAPI annotations (like @Operation, @APIResponse) are on the interface. So I´d expect that it´s possible to generate the documentation out of it.

phillip-kruger:
@jglink - I created the following issues:

- https://github.com/smallrye/smallrye-open-api/issues/448 - to allow the type (yaml/json)
- https://github.com/smallrye/smallrye-open-api/issues/449 - scan interfaces in modules

PR #461, opened 2020-08-30
Contribute openapi-as-reactforms

tarilabs:
As discussed.

1. Commit 23db773 move **_the already existing_** `ui` down one level, into a submodule
2. the remainder of the commits contribute the https://github.com/tarilabs/openapi-as-reactforms to SmallRye, here

# Demo

Here is adding the dependency to a previously existing Kogito and Quarkus based application, added to the dependency `pom.xml` as per README:

image

Results:

image

We can notice as expected, this creates an entry in the drop-down per each of the POST endpoint exchaning JSON, as per description. The Form is build at runtime programmatically via React and ReactForms, following the Swagger/OpenAPI schema definition file selected (by default at `/openapi` but can be changed with the proper field as per the UI)


phillip-kruger (approved):
Thanks !

PR #465, opened 2020-09-04
Remove the schema-references.enable property

msmiths:
These changes remove the `mp.openapi.extensions.smallrye.schema-references.enable` MP Config property.  Based on discussions in Gitter, it appear that this property was left over from when support for $ref was still experimental. Now that the implementation of $ref support is complete, this property should be removed.


MikeEdgar (approved):
Looks good to me, thanks @msmiths

PR #466, opened 2020-09-04
Scan sibling `@Extension` annotations for extensible models

### LINKED ISSUE #267: @Extension annotations are not processed when declared on fields.
Hi,

It seems that the microprofile-openapi `@Extension` annotation is only processed if its on the Resource class or on a jax-rs method, but skipped if its on an object (placed on the class or on a field), so in these cases the defined extension is not present in the openapi.yml model.

I created an example in the tck project (in ExtensionsTest: https://github.com/petrenyi-mark/smallrye-open-api/commit/85eea7ed3265bf31ad666900ed31def59b4b29a0)

MikeEdgar:
Fixes #267 

Basic approach is to match `@Extension`s found on the same target with `Extensible` models generated from the same target (with some limitations, e.g. `@Parameter`).

This is targeted for `2.1.0`


phillip-kruger (approved):
Thanks @MikeEdgar !!

PR #488, opened 2020-10-02
Allow reuse of a component response's status code on operations (2.0.x)

### LINKED ISSUE #487: Allow Global APIResponse Status Reuse
When used inside of `@Components`, the `responseCode` value of an `@APIResponse` is not usable from the operations where it is referenced. Using multiple `@APIResponse` annotations on a method with a `ref` to a global response results in a single `default` response. 

Example:
```java
    @OpenAPIDefinition(
            info = @Info(title = "Cool API", version = "0.1"),
            components = @Components(responses = { 
                @APIResponse(responseCode = "404", description = "Not Found!", name = "NotFound"),
                @APIResponse(responseCode = "500", description = "Server Error!", name = "ServerError")    
            }))
    static class CoolAPIApp extends Application {
    }

    @Path("pets")
    static class PetResource {
        @Path("{id}")
        @GET
        @Consumes(MediaType.APPLICATION_JSON)
        @APIResponse(responseCode = "200")
        @APIResponse(ref = "NotFound")
        @APIResponse(ref = "ServerError")
        public Object getPet(@PathParam("id") String id) {
            return null;
        }
    }
```
The resulting OpenAPI shows that the `404` status is "lost".
```json
{
  "openapi" : "3.0.3",
  "info" : {
    "title" : "Test title",
    "version" : "0.1"
  },
  "paths" : {
    "/pets/{id}" : {
      "get" : {
        "parameters" : [ {
          "name" : "id",
          "in" : "path",
          "required" : true,
          "schema" : {
            "type" : "string"
          }
        } ],
        "responses" : {
          "default" : {
            "$ref" : "#/components/responses/ServerError"
          },
          "200" : {
            "description" : "OK",
            "content" : {
              "application/json" : {
                "schema" : {
                  "type" : "object"
                }
              }
            }
          }
        }
      }
    }
  },
  "components" : {
    "responses" : {
      "NotFound" : {
        "description" : "Not Found!"
      },
      "ServerError" : {
        "description" : "Server Error!"
      }
    }
  }
}
```
My expectation as a user is that the response status from the global component should be known to the scanner and used as the key to the operation's `responses` map (via an implementation-specific field in `APIResponseImpl`). Specifying both `responseCode` and `ref` on the operation's annotation is possible, but results are undefined for other properties when `ref` is given, and it also results in repeated information.

@EricWittmann , @phillip-kruger - w.d.y.t.? I've already coded this, but wanted to hold off on a PR in case some additional discussion is needed or a different approach is possible that I'm not thinking of.

MikeEdgar:
Fixes #487 on branch `2.0.x`

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/models/responses/APIResponseImpl.java
+++ b/core/src/main/java/io/smallrye/openapi/api/models/responses/APIResponseImpl.java
@@ -23,6 +23,7 @@ public class APIResponseImpl extends ExtensibleImpl<APIResponse> implements APIR
     private Map<String, Header> headers;
     private Content content;
     private Map<String, Link> links;
+    private String responseCode;
 
     /**
      * @see org.eclipse.microprofile.openapi.models.Reference#getRef()
@@ -143,4 +144,11 @@ public class APIResponseImpl extends ExtensibleImpl<APIResponse> implements APIR
         ModelUtil.remove(this.links, name);
     }
 
+    public String getResponseCode() {
+        return responseCode;
+    }
+
+    public void setResponseCode(String responseCode) {
+        this.responseCode = responseCode;
+    }
 }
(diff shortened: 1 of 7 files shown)


phillip-kruger (approved):
Nice ! I like that the model is now available on the scanning context - might be useful down the line.

PR #524, opened 2020-10-24
Support XmlAccessorType, JSON-P types, option to disable array refs (2.0.x)

### LINKED ISSUE #508: Ignore some classes for schema generation by default
Since upgrading from Thorntail to Quarkus, the schema generation became much more verbose.

1. For every `Set` and `List` an anonymous type is generated, which clutters the Swagger UI Schema overview.
2. `javax.json.JsonValue` and its descendants are also listed and all their getters are parsed (generates another schema for `JsonValue.ValueType` for example).
3. Date fields are now parsed too, even though they're almost never transmitted as a complex object but as a string or number.

Here's an old vs. current comparison:
old new
Left was under Thorntail 2.5, right is with Quarkus 1.8.3.

I propose to not autogenerate schemas for those types, since they don't provide much additional info and at the cost of clarity. Especially when those are inner fields and not the root node. Knowing what's a Set, List or Date interally doesn't really make a difference since we're still talking REST, so there's no corresponding type. Auto generating a hint at the field level might be an alternative, if we don't want to omit the info completely.

Thanks for considering.

MikeEdgar:
Fixes #508 for `2.0.x` branch.

Adds new configuration property: `mp.openapi.extensions.smallrye.array-references.enable` with default `true`.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
+++ b/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
@@ -73,6 +73,10 @@ public interface OpenApiConfig {
         return true;
     }
 
+    default boolean arrayReferencesEnable() {
+        return true;
+    }
+
     default String customSchemaRegistryClass() {
         return null;
     }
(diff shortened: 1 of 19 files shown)


phillip-kruger (approved):
LGTM

PR #530, opened 2020-10-28
Allow multiple filters.

phillip-kruger:


phillip-kruger:
Backport of already approved multiple filters change

PR #549, opened 2020-11-16
Fix case where Spring use path and not value for request mapping (2.0.x)

### LINKED ISSUE #548: Spring Web app fails when using `@RequestMapping.path` vs `value`
Original issue: https://github.com/quarkusio/quarkus/issues/13323

phillip-kruger:
Fix #548


MikeEdgar (approved):
Looks good to me!

phillip-kruger:
Hold on, some more fixes coming. For the method as well...

PR #568, opened 2020-12-01
Preemptive scan of classes with Schema annotation into components

### LINKED ISSUE #554: Include schema for additional Java class in generated OpenAPI file
I'm using the `smallrye-open-api-maven-plugin` to generate the OpenAPI json/yaml from a Java application which uses the `org.eclipse.microprofile.openapi.annotations`.
I have a few classes which are not exposed over the REST API (not "in a typed way" at least), but I would like to include them in the generated OpenAPI schema file. Simply adding an annotation like `@Schema` to the class does not seem to work.

Is there a way to include additional Java classes when generating the schema file using the maven plugin?

Is it maybe possible to use a custom implementation of `OASModelReader` or similar, which somehow uses the smallrye implementation and then adds a few classes/types?
However, my first simple attempt resulted in a ClassNotFoundException when defining `mp.openapi.model.reader` (or `modelReader`).

Thanks a lot in advance!

MikeEdgar:
Fixes #554 

- Provide helpers for a couple `AnnotationScanner` methods in `CurrentScannerInfo` since no scanner implementation is available at the point the class-level `@Schema` annotations are scanned.
- Minor `pom.xml` housekeeping to suppress console logging during unit tests

Open question - does this also require a configuration option? My opinion is that it's not particularly surprising that `@Schema` annotated classes would be scanned and users can simply remove the annotation if necessary.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -88,11 +88,17 @@
             <artifactId>jsonassert</artifactId>
             <scope>test</scope>
         </dependency>
-        
     </dependencies>
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <redirectTestOutputToFile>true</redirectTestOutputToFile>
+                </configuration>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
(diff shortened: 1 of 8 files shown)


phillip-kruger (approved):

phillip-kruger:
Not sure about the config question. Maybe we can add it if users ask ?
Are you going to do one for 2.0.x as well ?

MikeEdgar:
Yeah, I'll do the same for 2.0. I'm fine with adding the config later if necessary.

phillip-kruger (approved):

PR #577, opened 2020-12-03
Run TCK with Arquillian, WELD and RESTEasy.

radcortez:
This is for us to have a more consistent environment to run the TCK for MP compliance.

The old TCK runner is still around. We can either remove it or keep it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,6 +39,12 @@ jobs:
       - name: build with maven
         run: mvn -B formatter:validate verify --file pom.xml
 
+      - uses: actions/upload-artifact@v2
+        name: tck-report
+        with:
+          name: tck-report
+          path: testsuite/tck/target/surefire-reports
+
   quality:
     needs: [build]
     if: github.event_name == 'push' && github.repository == 'smallrye/smallrye-open-api'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,6 +49,12 @@ jobs:
           git push
           git push --tags
 
+      - uses: actions/upload-artifact@v2
+        name: tck-report
+        with:
+          name: tck-report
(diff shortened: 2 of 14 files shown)


MikeEdgar:
Thanks @radcortez , this looks good. One question I have is regarding the Github action - should the path be testsuite/tck/target/... rather than tck/target/... ? Also, where are the files uploaded, as attachments to the repository?

radcortez:
> Thanks @radcortez , this looks good. One question I have is regarding the Github action - should the path be testsuite/tck/target/... rather than tck/target/... ? 

Yes! Good catch.

> Also, where are the files uploaded, as attachments to the repository?

They are uploaded to each CI Build. Here is an example: https://github.com/smallrye/smallrye-config/actions/runs/398298378. The new MP Compatible Certification requires the TCK results to be accessible somehow. This was the easiest way to do it.

phillip-kruger (approved):

phillip-kruger:
I think we should remove the 'old' TCK and remame this to TCK. @MikeEdgar  ? W.d.y.t ?

phillip-kruger:
Merging this PR, if all agree we can remove old TCK in a separate PR.

MikeEdgar:
@phillip-kruger - I agree, but there are a few non-TCK tests in that module that should be kept somewhere.

phillip-kruger:
Good point ! Can we just add it to the new TCK module maybe ?

MikeEdgar:
Maybe, but they rely on the custom framework the old TCK uses. It might be good to keep the new TCK module purely TCK tests. I can take a quick look and see if it's feasible to have them run differently.

radcortez:
The additional tests still use the Arquillian annotations, so in theory,  they should work with the new TCK structure by removing the custo annotations and keeping the Arquillian stuff.

I can have a look into that if you prefer :)

PR #579, opened 2020-12-04
Added option to configure default produces and consumes (master)

phillip-kruger:


MikeEdgar (approved):
Looks good

PR #580, opened 2020-12-04
Make sure we can not add the same filter twice (master)

phillip-kruger:


MikeEdgar (approved):

phillip-kruger:
Thanks @MikeEdgar !

PR #583, opened 2020-12-05
Remove TCK test class wrappers and move remaining tests to `extra` module

MikeEdgar:
Follow up to #577 

Add log formatting to new TCK for readability.


MikeEdgar:
@radcortez - this is the follow-up to #577 with the remaining cases moved to a new module under `testsuite`. This matches the layout in Config as well and allows the TCK tests to stay separate in their own module.

phillip-kruger (approved):

phillip-kruger:
Thanks @MikeEdgar !

radcortez (commented):
+1

PR #590, opened 2020-12-08
small fixes support xml

abatalev:
added support xml/name for classes with XmlRootElement(name) 
added support attribute for Property  with XmlAttribute

```yaml
components:
  schemas:
    book:
      type: object
      xml:                <--
        name: 'xml-book'  <--
```

```yaml
book:
      type: object
      properties:
        id:
          type: integer
          xml:               <--
            attribute: true  <--
```

see https://swagger.io/docs/specification/data-models/representing-xml/


MikeEdgar:
@abatalev, thanks for the contribution. I plan on reviewing this in the next day or two.

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
```diff
@@ -189,10 +187,41 @@ Schema processField() {
...
+            processXmlAttr(fieldSchema, field.annotation(DotName.createSimple("javax.xml.bind.annotation.XmlAttribute")));
+            return;
+        }
+        MethodInfo readMethod = typeResolver.getReadMethod();
+        if (readMethod != null) {
+            processXmlAttr(fieldSchema, readMethod.annotation(DotName.createSimple("javax.xml.bind.annotation.XmlAttribute")));
+            return;
+        }
+        MethodInfo writeMethod = typeResolver.getWriteMethod();
+        if (readMethod != null) {
```
MikeEdgar, line 208: Should this be `writeMethod` ?

core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
```diff
@@ -253,6 +256,19 @@ private void depthFirstGraphSearch() {
         }
     }
 
+    private void processClassAnnotations(Schema schema, ClassInfo classInfo) {
+        AnnotationInstance annotation = TypeUtil.getAnnotation(classInfo,
+                DotName.createSimple("javax.xml.bind.annotation.XmlRootElement"));
```
MikeEdgar, line 261: Please add constants for this and the other `DotName`s in `io.smallrye.openapi.api.constants.JaxbConstants`

MikeEdgar (changes requested):
@abatalev - I added a few comments in a few places. Please take a look. There also seems to be a very large number of test files added for the nature of the change. Is there any way for the tests to be more focused on the particular change?

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
abatalev: ок

core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
abatalev: ок

abatalev:
>  Is there any way for the tests to be more focused on the particular change?
I have no options. Any ideas?

MikeEdgar:
I think a single REST resource with methods using the JAXB objects you defined as the request body and the response would do nicely. You might want to have one of the JAXB objects also reference a nested JAXB object to cover the field scanning part.

abatalev:
Ок. 
1. I will delete JaxbWithNameJaxRsAnnotationScannerBasicTest and XmlJaxRsAnnotationScannerBasicTest
1. I will modify JaxbJaxRsAnnotationScannerBasicTest for use multiple resources

abatalev:
@MikeEdgar - Done

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
```diff
@@ -189,10 +190,41 @@ Schema processField() {
             fieldSchema = MergeUtil.mergeObjects(typeSchema, fieldSchema);
         }
 
+        processFieldAnnotations(fieldSchema, typeResolver);
+
```
MikeEdgar, line 194: @abatalev - sorry, one more. I think this method call should occur before the if/else starting on line 177. The reason is because the XML changes should be applied prior to the potential schema composition that occurs in that condition. 

The other thing I would mention is that `processFieldAnnotations` will not find method annotations if an unannotated field is present in the class (likewise for the writer method, if a reader method is present).

MikeEdgar (changes requested):
This should be the last change request - thanks @abatalev

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
abatalev: Done

MikeEdgar:
@abatalev - this looks good. @radcortez, are we OK with changing functionality in `master` before the 2.1.0 release?

radcortez:
> @abatalev - this looks good. @radcortez, are we OK with changing functionality in `master` before the 2.1.0 release?

Yes. It is fine.

MikeEdgar (approved):

MikeEdgar (approved):

PR #609, opened 2020-12-20
Support XmlWrappedElement

abatalev:
Support XmlWrappedElement

```
books:
  type: array
  items: 
    type: string 
    xml:
      name: 'item'
  xml:                <--
    wrapped : true    <--
    name: books-array <--
```

https://swagger.io/docs/specification/data-models/representing-xml/

### DIFF
--- a/extension-jaxrs/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/JaxbWithNameGreeting.java
+++ b/extension-jaxrs/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/JaxbWithNameGreeting.java
@@ -1,6 +1,9 @@
 package test.io.smallrye.openapi.runtime.scanner.entities;
 
+import java.util.List;
+
 import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlElementWrapper;
 import javax.xml.bind.annotation.XmlRootElement;
 
 @XmlRootElement(name = "MyGreeting")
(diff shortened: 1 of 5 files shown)


extension-jaxrs/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/JaxbWithNameGreeting.java
```diff
@@ -9,11 +12,20 @@
...
         this.message = message;
+        this.books = items;
     }
 
     public String getMessage() {
         return message;
     }
+
+    @XmlElementWrapper(name = "books-array")
+    private final List<String> books;
```
MikeEdgar, line 25: If you add `@XmlElement(name = "item")`, does the result match the example in the OpenAPI article you linked?

MikeEdgar (commented):
Thanks @abatalev - question on the naming of the array items (see below). It might be nice to have the result include the xml name for that as well. Thoughts?

extension-jaxrs/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/JaxbWithNameGreeting.java
abatalev: Yes. I wanted to add support for XmlElement in the following PR.

MikeEdgar (approved):
Looks good to me. Thanks @abatalev.

PR #629, opened 2021-01-14
Merge class-level schema property attributes with field schemas

### LINKED ISSUE #628: @SchemaProperty will not render on existing fields when utilized within class source
When creating a demo application, I ran into a potential bug. 

When `@SchemaProperty `is used within a class file to define the `description` attribute for **existing** fields, it seems as though the resultant yaml does not include the `description` attribute's value.

```
// Example where the field already exists in the src
@Schema(name="Car", 
        description="POJO that represents a single car.", 
        properties = {
                @SchemaProperty(name = "manufacturer",  description = "Test description")
        })
public class Car {
    @Schema(required = true, example="Ford")
    private String manufacturer;

//output
components:
  schemas:
    Car:
      description: POJO that represents a single car.
      required:
      - manufacturer
      type: object
      properties:
        manufacturer:
          type: string
          example: Ford
```

Furthermore, when defining the `description` attribute for a field which does **not exist** within the class' src, both the name and description are rendered in the resultant yaml.

```
// Example where the field does not exist in the src
@Schema(name="Car", 
		description="POJO that represents a single car.", 
		properties = {
			@SchemaProperty(name = "fakeField", description = "Test description")
		})
public class Car {
	private String manufacturer;
//output yaml
components:
  schemas:
    Car:
      description: POJO that represents a single car.
      type: object
      properties:
        fakeField:
          description: Test description
        manufacturer:
          type: string
```

The MP OpenAPI specification  states the following for the `@SchemaProperty` annotation, which seems to suggest `@SchemaProperty` should render, assuming it is nested within a `@Schema` annotation:
> Allows the definition of a property nested within a parent @Schema.

This leads me to believe that the behaviour within the first example is currently a bug.

MikeEdgar:
Fixes #628 

@scottcurtis2605 - please have a look and let me know if this covers your demo cases.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
@@ -21,6 +21,7 @@ import io.smallrye.openapi.runtime.scanner.SchemaRegistry;
 import io.smallrye.openapi.runtime.scanner.dataobject.BeanValidationScanner.RequirementHandler;
 import io.smallrye.openapi.runtime.scanner.spi.AnnotationScannerContext;
 import io.smallrye.openapi.runtime.util.JandexUtil;
+import io.smallrye.openapi.runtime.util.ModelUtil;
 import io.smallrye.openapi.runtime.util.TypeUtil;
 
 /**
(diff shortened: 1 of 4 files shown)


scottcurtis2605:
Planning to get around to testing this today.

scottcurtis2605:
> Planning to get around to testing this today.

Currently having some issues with verifying whether I'm building OL correctly with the forked version of the JAR. Will investigate this further on Monday.

scottcurtis2605 (approved):
LGTM!

phillip-kruger (approved):
Nice ! Thanks

PR #630, opened 2021-01-15
Bump Config to 2.0.0, RestEasy to 4.6.0, update test classes

MikeEdgar:


phillip-kruger (approved):
Thanks Mike. I wonder if we also need to backport to 2.0.x ?

PR #637, opened 2021-01-19
RestEasy Reactive support - new annotations and derived path params

### LINKED ISSUE #576: Support RESTEasy Reactive extensions
We currently have extensions that SR-OpenApi needs to be aware of:

- `@RestPath`, `@RestQuery` and friends, which have their value name optional, like the `@PathParam` and friends in RESTEasy, but we renamed them because most users were super confused with them having the same name
- default produces/consumes rules (not yet finalised)
- `@PathParam` is optional and inferred if there's an existing path param declared with the same parameter name
- `@Context` is optional and inferred if there's an existing context type known at build time
- We're thinking of doing the same for CDI beans (not yet finalised)
- We're investigating a genericised `Response` which would allow you to know the return type even when returning a response (not yet finalised).

Did I forget anything @geoand ?

MikeEdgar:
Fixes #576

Note: un-annotated path params will work for any JAX-RS annotation set (since we can't tell which annotation was meant to be omitted).

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/ResourceParameters.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/ResourceParameters.java
@@ -3,6 +3,8 @@ package io.smallrye.openapi.runtime.scanner;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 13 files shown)


phillip-kruger (approved):
Thanks @MikeEdgar ! Waiting for CI. Please backport !!

phillip-kruger (approved):

phillip-kruger:
CI Broken. Merging.

PR #643, opened 2021-01-21
Remove 204 response code default for DELETE method

### LINKED ISSUE #642: Incorrect Response Code - DELETE
When implementing the `@APIResponseSchema` tests as described within #621, the following behaviour was observed: 

When testing that the correct `responseCode` attribute on `@APIResponseSchema` is set in cases when an invalid `responseCode` value is passed to the attribute, a JAX-RS `@DELETE` method with `void` return type, and a JAX-RS `AsyncResponse` parameter, returns a `204` response code. This can be seen [within my fork](https://github.com/scottcurtis2605/smallrye-open-api/blob/4ec855fc724dec556c603d7c4faf01a563126446/extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/ApiResponseTests.java#L183).

According to the [JavaDoc](https://download.eclipse.org/microprofile/microprofile-open-api-2.0/apidocs/), I believe expected response code in this scenario should be `200`:

```
 If no value is specified, a default shall be determined using REST conventions as follows:

- If the method's return type is void and the HTTP method is @POST, the code will be 201.
- Otherwise, if the method's return type is void the method does not list a JAX-RS AsyncResponse parameter, the code will be 204.
- Otherwise, the code will be 200. 
```

Assuming the JavaDoc should read, "Otherwise, if the method's return type is void **_AND_** the method does not list a JAX-RS AsyncResponse parameter, the code will be 204.", I believe the current behaviour displayed by the library is incorrect.

MikeEdgar:
Fixes #642 

@phillip-kruger - I think the logic for DELETE was added with the Spring changes, but since there is no `AsyncResponse` for Spring, it doesn't appear necessary. Please have a look and let me know if you disagree and we can adjust this.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
@@ -438,8 +438,6 @@ public interface AnnotationScanner {
         if (isVoidResponse(method)) {
             if (isPostMethod(method)) {
                 status = 201; // Created
-            } else if (isDeleteMethod(method)) {
-                status = 204; // No Content (Maybe this should be 202 Accepted ?)
             } else if (!isAsyncResponse(method)) {
                 status = 204; // No Content
             } else {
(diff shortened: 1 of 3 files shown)


phillip-kruger (approved):
Thanks @MikeEdgar !!

Issue #621, opened 2021-01-06
2.0 Negative Test Coverage for API/SPI Changes


scottcurtis2605:
As per the listed [API/SPI changes in 2.0](https://download.eclipse.org/microprofile/microprofile-open-api-2.0-RC3/microprofile-openapi-spec.html#_apispi_changes), I was unable to identify unit tests within the codebase which examine & verify expected behaviour within the following negative scenarios:

`@SchemaProperty`:
- Implementation class also specified with conflicting properties/values (ensure `@SchemaProperty` values override implementation class)
- Testing scenarios with malformed properties (e.g. malformed properties such as name or value, if these exist)

`@RequestBodySchema`:
- Reference non-existent class
- Referenced class does not define a valid schema
- Malformed properties (wrong name, wrong data type)

`@APIResponseSchema`:
- Reference non-existent class
- Referenced class does not define a valid schema
- Malformed properties (wrong name, wrong data type)

`mp.openapi.schema.*` MP Config keys:
- Reference non-existent class
- Referenced class does not define a valid schema
- Malformed properties (wrong name, wrong data type)

MP Config extensions defined by SmallRye for limiting classes/dependencies that are scanned :
- Reference non-existent jar(s)
- Malformed list (not comma-separated, specify URLs, etc.)

I will attempt to implement the following tests, as the behaviour above is required to be examined for Open Liberty's OpenAPI 2.0 release approval.

If I have missed any tests which already examine the following behaviour, please feel free to direct me to those!

MikeEdgar:
Thanks @scottcurtis2605 . Tests for these scenarios would be a great addition. The test that `@SchemaProperty` overrides implementation class attributes also might be a good addition to the TCK in the spec project.

Azquelt:
I looked through this list with Scott, here's what we're going look at:

`@SchemaProperty`
 - test invalid values for each attribute

`@RequestBodySchema` / `@APIResponseSchema` / `@Schema`
 - these are all used to do similar things and reference a class to define a schema
 - is it possible for a referenced class not to define a valid schema or to be invalid in some other way?

`@APIResponseSchema`
 - test invalid description/code (presumably the code must be a number)

`mp.openapi.schema.*`
 - test if a key is defined for a class which does not exist (expect that it would be ignored since we never need to generate a schema for the type)
 - test with a value which is not valid JSON
 - test with a value which is valid JSON but is not a valid schema

[...]

PR #644, opened 2021-01-21
621 - @APIResponseSchema Negative Test Coverage 

scottcurtis2605:
Implements the negative tests specified for `@APIResponseSchema` in relation to issue #621. These investigate the behaviour specified within @Azquelt's comment:

- Response code validation
  - Multiple tests to check `responseCode` is appropriately set in accordance to the java doc when a valid number isn't given, for each common type of HTTP method, with and without the JAX-RS `AsyncResponse` parameter and `void/non void` types.

- Description validation 
  - Single test to check the default description of the given `responseCode` passed is set appropriately in accordance to the [HTTP/1.1 documentation](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10).


mehmetanbaki (commented):
nice work

phillip-kruger (approved):
LGTM. @MikeEdgar can you also confitm.

MikeEdgar (approved):
Looks good to me as well. Thanks @scottcurtis2605.

PR #647, opened 2021-01-27
mp.openapi.schema.* Config Property Negative Test Coverage

scottcurtis2605:
Implements the negative tests specified for `mp.openapi.schema.* ` in relation to issue #621. These investigate the behaviour specified within @Azquelt's comment:

- Test the output for utilising a key defined for a class which does not exist.
- Test config property values which are not valid JSON.
- Test config property values which are valid JSON, but are malformed properties on the schema.

I have currently left comments in on some of these tests which I will remove before merging (assuming all else is good!). These are in place to check that the output produced by these tests is what should be expected in negative scenarios, regarding feedback to the developer in these scenarios. This is especially the case in the test for malformed property names - I have left a [comment](https://github.com/smallrye/smallrye-open-api/issues/621#issuecomment-768493657) in #621 regarding this.

Other changes which were made:
- Renamed `ConfigTest` to `ConfigExtensionsTest`. I believed it would be better to keep tests referring to `mp.openapi.extensions.*` separate from those for `mp.openapi.schema.* `. 

The reasoning for this change was for future-proofing purposes. This avoids this class growing too large (covering all config properties) at a later point in time, adding some separation and further maintainability to the code base if more Config tests are to be implemented. Perhaps `ConfigTest` could be introduced as a test suite to encapsulate all `Config*Test` classes at a later point in time.

### DIFF
--- a/extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/ConfigExtensionsTest.java
+++ b/extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/ConfigExtensionsTest.java
@@ -16,11 +16,12 @@ import test.io.smallrye.openapi.runtime.scanner.entities.Greeting;
 import test.io.smallrye.openapi.runtime.scanner.resources.GreetingGetResource;
 
 /**
- * Basic tests to check the setting of certain Info elements via config
+ * Basic tests to check the setting of certain Info elements via config, using
+ * extensions config properties
  * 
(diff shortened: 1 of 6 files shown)


phillip-kruger (approved):
LGTM ! @MikeEdgar ?

MikeEdgar (approved):
Looks good, thanks @scottcurtis2605 !

MikeEdgar:
Note, the merge was a rebase/merge.

PR #651, opened 2021-02-03
Ignore all unimplemented interfaces unless included via config

### LINKED ISSUE #449: Allow scanning in interfaces in separate module
Allow the creation of schema when the annotations are in a seperate module in an interface:

From https://github.com/smallrye/smallrye-open-api/pull/445:

"But the OpenAPI generation still needs an application class for the JAX-RS annotations to get scanned. If I only have an interface with JAX-RS annotations (implementation is in another module), the OpenAPI documentation is basically empty.

You can find a small reproducer project here: https://github.com/jglink/openapi-doc-from-interface
The project only has 1 interface class (my.project.api.SomeApi) with a JAX-RS annotation and so the generated openapi.yaml only defines some basic stuff.

All the OpenAPI annotations (like @Operation, @APIResponse) are on the interface. So I´d expect that it´s possible to generate the documentation out of it"

MikeEdgar:
Fixes #449

Also :
- Add Junit 5 to JAX-RS extension module

Note to @jglink - with this change you will need to add the interface class to you configuration using `mp.openapi.scan.classes` or `mp.openapi.scan.packages`


phillip-kruger (approved):

phillip-kruger:
Thanks @MikeEdgar !! Would you back port as well ?

jglink:
Thanks for the information. 
I will have a look at this asap.

MikeEdgar:
> Thanks @MikeEdgar !! Would you back port as well ?

@phillip-kruger - I'm thinking about doing another (much smaller this time) merge from `master` to `2.0.x` after a few additional fixes.

PR #652, opened 2021-02-04
Preserve schema model's type when value from annotation is null

### LINKED ISSUE #649: SchemaFactory overwrites default type with null when not specified in `@Schema`
See https://github.com/quarkusio/quarkus/issues/14670 for original issue.

MikeEdgar:
Fixes #649 

The null schema type is causing the generated schema to be excluded from the output, as described in https://github.com/quarkusio/quarkus/issues/14670.


phillip-kruger (approved):
Thanks @MikeEdgar . I see you have set the milestone as 2.0.x, but we merge into master, so I am changing to 2.1.x.
We would need backports for 2.0.x :)

PR #653, opened 2021-02-04
Add Jandex `equals` implementations

### LINKED ISSUE #646: @Parameter ignored on parameter if @PathParam also present
(transcribed with some editing from a Google Groups discussion https://groups.google.com/g/smallrye/c/xvhDsfsE6F4)

OpenAPI seems to ignore a `@Parameter` annotation on a parameter if there is also a `@PathParam` annotation.

Here's an example:

https://github.com/tjquinno/helidon/blob/0be4b63cfaadd04ac55cb4ba716716117b396d09/examples/microprofile/openapi-basic/src/main/java/io/helidon/microprofile/examples/openapi/basic/GreetResource.java#L114-L117

```java
    public JsonObject getMessage(
            @Parameter(description = "This is ignored")
            @PathParam("name") String name) {
```
I've seen this in 2.0.17 and earlier releases.

Here's what I saw stepping through the code in my case.

The code happens to process the `@PathParam` first (not that it should matter), then the argument-level `@Parameter` annotation.

When the code processes `@Parameter`, `ParameterProcessor#getParameterContext` tries to find the pre-existing context for the parameter. There seems to be one — with key `name: name, in: path` but the key passed to `getParameterContext` associated with the `@Parameter` annotation is `name: name, in: null`. I'm not sure if that's OK or not, but in any event the code does not find the pre-existing context using any of the three attempts:

- in the map directly (this seems expected),
- using `haveSameAnnotatedTarget` in 1205:1211 (to me, this is a surprise), or
- the "last chance" check at the end of the method.

I'm guessing that maybe the intent was that `haveSameAnnotatedTarget` would return `true` for this case, given that the pre-existing context entry created from the `@PathParam` and the context being created for the parameter-level `@Parameter` annotation should reflect the same target.

I did not see that `MethodParameterInfo` implements `equals`, so it would inherit the implementation from `Object` which does a simple `==` comparison of the two object references. And, indeed, the code is comparing two different instances of `MethodParameterInfo` although they look to me to be referring to the same target. 

If `haveSameAnnotatedTarget` returned `true` here, then I think the rest of the code would combine the information from the two annotations which, from the outside anyway, would seem to make sense. As it is, I see two contexts in the processor's `params` collection where I'd expect to see a single, combined context for the parameter.

But maybe there's more to it that I'm missing.

Mike reported seeing the expected results, not mine, when he tried the same example. I can't explain why we are getting different results unless perhaps the order in which the `@PathParam` and `@Parameter` annotations are dealt with does matter and for some reason our test runs process them in different orders.

Perhaps if `MethodParameterInfo` overrode `equals` the order wouldn't matter (if in fact that's the problem). Maybe it should implement `equals` anyway? I realize that class is not in SmallRye OpenAPI's code base.

If there's anything else I can try from here, please let me know. For the moment I'm not sure what I might be doing to cause the problem.

MikeEdgar:
Fixes #646 

Ensure de-serialized index associates multiple method parameter annotations

CC:  @tjquinno

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
@@ -1179,7 +1179,7 @@ public abstract class AbstractParameterProcessor {
          */
         boolean nameMatches = (context.name == null || name == null || Objects.equals(context.name, name));
 
-        if (target.equals(context.target)) {
+        if (JandexUtil.equals(target, context.target)) {
             /*
              * The name must match for annotations on a method because it is
              * ambiguous which parameters is being referenced.
--- a/core/src/main/java/io/smallrye/openapi/runtime/util/JandexUtil.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/util/JandexUtil.java
@@ -18,6 +18,7 @@ import org.jboss.jandex.AnnotationTarget.Kind;
 import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
+import org.jboss.jandex.FieldInfo;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.MethodParameterInfo;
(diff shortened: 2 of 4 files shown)


phillip-kruger (approved):
Nice ! Thanks @MikeEdgar

tjquinno:
@MikeEdgar Thank you!

PR #658, opened 2021-02-05
Unwrap generic type from JAXBElement (master)

### LINKED ISSUE #650: Is there any way override the conversion of Java object (JAXBElement) to JSON?
I wonder if there is any way I could take control over the conversion of the Java Object (JAXBElement) to JSON? The problem is that when openapi generates document, the JSON gets the wrong format. It looks like this in swagger-ui.

bild

It shouldn't include the "namespaceURI", "prefix" and so on.

I have generated the classes from a WSDL-file and I need to marshall the objects like this:
bild

- Is there any way I can achieve this so the JSON gets the correct format in swagger-ui? 
- Is it possible to override the method that openapi uses to marshall Object to JSON? 
- Which method in which interface is this?

MikeEdgar:
Fixes #650

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -78,6 +78,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>jakarta.xml.bind</groupId>
+            <artifactId>jakarta.xml.bind-api</artifactId>
+            <version>${version.jakarta.xml.bind.api}</version>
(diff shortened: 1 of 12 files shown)


phillip-kruger (approved):
Awesome !! Thanks @MikeEdgar !!

phillip-kruger:
Backport ?

PR #662, opened 2021-02-07
 Parameter schema improvement & bug fix

### LINKED ISSUE #659: @Parameter(ref = ...) overwrites other parameters
## Affected Artifact
`io.smallrye:smallrye-open-api-maven-plugin:2.1.1`

## Problem Context
I use `smallrye-open-api` to generate OpenAPI Specs from JAX-RS/MP-OpenAPI annotations. I have a method like following:

```java
  @GET
  @Path("/{pathParam1}/{pathParam2}")
  @Parameter(ref = "queryParam1")
  String exampleEndpoint(
    @PathParam("pathParam1") String pathParam1,
    @PathParam("pathParam2") String pathParam2);
```

While there's two `@PathParam`s, there's also a query parameter given as `@Parameter` with a reference to the definition like following:

```java
@OpenAPIDefinition(
    // ...
    components = @Components(parameters = {
        @Parameter(name = "queryParam1", in = ParameterIn.QUERY)
    })
)
```

## Expected Behaviour
The generated OpenAPI specs contain all three parameters in the `paths: /{pathParam}: get: parameters:` section.

## Actual Behaviour
The generated OpenAPI specs do _only_ contain the `queryParam1` and the `pathParam2`. The `pathParam1` is missing. 

## Further Observations
* When `queryParam1` would not use a `ref`, but was defined directly on the method, it would work and we could see all three parameters in the spec.
* When there'd be only `pathParam1` in the code, without `pathParam2`, then `queryParam1` would be the only parameter in the generated specs.
* When there's no `@Parameter` annotation at all, all two path parameters are generated.
* When the path params are also given als `@Parameter`, the behaviour doesn't change compared to what's given above under "Actual Behaviour":
  ```java
    @Parameter(name = "pathParam1", in = ParameterIn.PATH)
    @Parameter(name = "pathParam2", in = ParameterIn.PATH)
    @Parameter(ref = "queryParam1")
  ```
* If we would declare `pathParam1` and `pathParam2` through `@Parameter(ref = ...)`, we also get similar results. However, if we leave out the `queryParam1`, we get `pathParam2` twice -- once as `ref`, and once as explicit specification.

## Reproducer
* Checkout https://github.com/fxnn/reproducer-smallrye-parameter
* Run `mvn package`
* Open the generated `target/generated/openapi.yaml`
* Compare the output with what's declared in `src/main/java/ExampleResource.java`

### LINKED ISSUE #660: Reference Schema Parameters drop local Schema information
Original Quarkus issue: https://github.com/quarkusio/quarkus/issues/14865

Local non-MP annotations that augment the parameter schema are not applied due to missing `type` on the `Schema` (it is a `ref`). The local schema can include local and referenced information using `allOf` similar to how `Schema` fields are processed.

MikeEdgar:
Fixes #659 
Fixes #660 
Resolves downstream issue https://github.com/quarkusio/quarkus/issues/14865

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/models/media/SchemaImpl.java
+++ b/core/src/main/java/io/smallrye/openapi/api/models/media/SchemaImpl.java
@@ -59,9 +59,14 @@ public class SchemaImpl extends ExtensibleImpl<Schema> implements Schema, ModelI
     private Boolean nullable;
     private Boolean writeOnly;
     private Boolean deprecated;
+    private int modCount;
 
     public static boolean isNamed(Schema schema) {
         return schema instanceof SchemaImpl && ((SchemaImpl) schema).name != null;
+    }
+
+    public static int getModCount(Schema schema) {
+        return schema instanceof SchemaImpl ? ((SchemaImpl) schema).modCount : -1;
     }
 
     public SchemaImpl(String name) {
@@ -74,6 +79,10 @@ public class SchemaImpl extends ExtensibleImpl<Schema> implements Schema, ModelI
 
     public String getName() {
         return name;
+    }
+
+    private void incrementModCount() {
+        modCount++;
     }
 
     /**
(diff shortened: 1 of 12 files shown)


phillip-kruger (approved):
@MikeEdgar , you are a legend !! Thanks  :)

PR #666, opened 2021-02-08
enable noHtml5Validate, validation msg reset feat.

tarilabs:
Executive summary: html5 validation interfers with mandatory boolean attribute from JSON schema. This resolve by disabling html5 validation, and providing a mean for end-user to reset library's form validation messages.

## Details

When an OpenAPI operation has object property as `boolean` and `required`:

Screenshot 2021-02-08 at 11 43 22

this interfers with the library form submission if html5 validation is also performed:

Screenshot 2021-02-08 at 11 42 34

This is documented here: https://react-jsonschema-form.readthedocs.io/en/latest/usage/validation/#html5-validation

This change disables html5 form validation per reference, thus setting `noHtml5Validate` to true.

Screenshot 2021-02-08 at 11 45 32

Additionally, this change allow end-user to reset the form validation messages (emitted by the library validation itself) taking a similar approach as in https://github.com/aerogear/mobile-developer-console/pull/338
Example.
Step1, some fields are missing:
image
Step2, User may opt to re-apply the payload (or reset it to blank):
Screenshot 2021-02-08 at 11 58 34
at this stage any missing fields can be valorized by the end-user, will reflect on the payload, and when all is okay the library's validation will not emit any "missing" message and fully submit, as detailed above.

I will shortly open a PR to cherry-pick and backport to `2.0.x` branch too.

### DIFF
--- a/ui/open-api-ui-forms/package-lock.json
+++ b/ui/open-api-ui-forms/package-lock.json
@@ -1148,17 +1148,17 @@
       }
     },
     "@rjsf/bootstrap-4": {
-      "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/@rjsf/bootstrap-4/-/bootstrap-4-2.3.0.tgz",
-      "integrity": "sha512-SuuLj/GokP5v2VK8o07d2R/qO+rv+XpRLQYV39Bhh3XIkdhfirYU46JXBt0Kb0usyyvyvRSRTO7i+6Lggnz88w==",
+      "version": "2.4.2",
+      "resolved": "https://registry.npmjs.org/@rjsf/bootstrap-4/-/bootstrap-4-2.4.2.tgz",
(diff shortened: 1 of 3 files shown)


MikeEdgar (approved):
Looks good to me. 

@tarilabs - one question I have about this module is - what is the purpose of using `lock-treatment-tool` in the npm build? I've been wondering if we can use `npm ci` instead of `npm install` to use the resolved `package-lock.json` versions for each build.

tarilabs:
hi @MikeEdgar thank you for the feedback!

No particular reason, I am not that much experienced with front-end, my understanding has been that when I need to make an NPM build using Maven (like in a small web-app inside of a broader Java/Maven project, like this case), it is best practice to define the pom.xml as here:
https://github.com/smallrye/smallrye-open-api/blob/30bb8c9810cd4412d51d0ac5c45529eef538d4fb/ui/open-api-ui-forms/pom.xml#L21-L76

If you believe there is a better setup for SmallRye and this f/e component inside the Maven build, please feel free to refactor it, I will be following with interest to learn something new :)

PR #672, opened 2021-02-09
Support inherited BV constraint annotations

### LINKED ISSUE #669: Constraints only appears in OpenAPI when annotations are against the field
**Describe the bug**
When constraints (such as @Max) are specified on a class field, these correctly display in OpenAPI (and via Swagger UI).
However when the fields are on the get method, they are not displayed in OpenAPI, although validation still occurs.

**Expected behavior**
Annotations should perform at the field and getter level in the same way. This is necessary because otherwise it is not possible to have validation specified on an interface or superclass.

**Actual behavior**
Constraints set on the getter form part of bean validation but do not display in the generated OpenAPI

**To Reproduce**
OpenAPI shows constraints correctly on a class when they are specified on the field

```
public class User {
   @Positive
   @Max(9999)
   private Integer id;

   private Integer getId() {
       return id;
   }
}
```
View in swagger-ui:

```
id
  integer($int32)
  maximum: 9999
  minimum: 1
```
However when these constraints are set on getId instead, the constraints no longer appear in the swagger-ui.

```
public class User {
   private Integer id;

   @Positive
   @Max(9999)
   private Integer getId() {
       return id;
   }
}
```

```
id
   integer($int32)
```

However in both cases, validation does succeed. 

See previous discussion in quarkusio/quarkus#12909

MikeEdgar:
Fixes #669

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
@@ -167,7 +167,10 @@ public class AnnotationTargetProcessor implements RequirementHandler {
             fieldSchema = MergeUtil.mergeObjects(new SchemaImpl(), typeSchema);
         }
 
-        BeanValidationScanner.applyConstraints(annotationTarget, fieldSchema, propertyKey, this);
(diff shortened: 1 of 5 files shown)


phillip-kruger (approved):
Thanks @MikeEdgar !!

PR #689, opened 2021-02-14
Update NPM packages and use `npm ci` (requires package-lock.json)

MikeEdgar:
* Update several NPM packages / regenerate `package-lock.json`
* Remove use of `lock-treatment-tool` - seems unnecessary, but glad to keep it if there is a good case for using it
* Use `npm ci` to install dependencies based on fixed versions in `package-lock.json`

If this all looks OK I'll create a PR for `2.0.x` also.

CC: @tarilabs

### DIFF
--- a/ui/open-api-ui-forms/package-lock.json
+++ b/ui/open-api-ui-forms/package-lock.json
@@ -5,367 +5,323 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.10.4",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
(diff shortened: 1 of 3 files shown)


phillip-kruger:
@tarilabs - please review. Thanks @MikeEdgar !

tarilabs (approved):
Verified locally, thanks for the ping it looks good to me! 👍 

Another thing I wanted to take a look at this was to replace the form library with Uniforms so to try leverage a PatternFly theme, but I have very thin bandwidth and not able to make sure if it does cover all the functionalities.. happy to share more details if anyone is interested to spend some cycles!

screenshot of local built and verify:
image

phillip-kruger:
Thanks @tarilabs !!

phillip-kruger (approved):
As per @tarilabs

PR #690, opened 2021-02-14
Fix/enhance handling of nested and custom collection types

### LINKED ISSUE #688: Nested collections generate dangling references
Original Quarkus issue: https://github.com/quarkusio/quarkus/issues/14814

MikeEdgar:
Fixes #688 
Relates to quarkusio/quarkus#14814

Changes behaviour to prevent the registration of known JDK collection types in `#/components/schemas`. This means that schemas previously generated such as `ListString` or `MapStringCustomBean` for types `List<String>` or `Map<String, CustomBean>` will instead be in-lined as array types. When the generic type argument for the collections are eligible for registration, they will still be referenced from the in-lined schemas from `items` for arrays or `additionalProperties` for objects.

Also addressed in an bug that allows for unintentional side affects during scanning when `Class.forName` is called during the inheritance check for classes absent from the Jandex index.

This still needs a bit of work to re-factor some of the added code to `TypeProcessor` that really belongs in `TypeResolver`.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeProcessor.java
@@ -10,13 +10,9 @@ import static io.smallrye.openapi.runtime.scanner.OpenApiDataObjectScanner.SET_T
 import static io.smallrye.openapi.runtime.scanner.OpenApiDataObjectScanner.STRING_TYPE;
 import static io.smallrye.openapi.runtime.util.TypeUtil.isTerminalType;
 
-import java.util.List;
-import java.util.Map;
-
 import org.eclipse.microprofile.openapi.models.media.Schema;
 import org.jboss.jandex.AnnotationTarget;
 import org.jboss.jandex.ArrayType;
-import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.ParameterizedType;
 import org.jboss.jandex.Type;
 import org.jboss.jandex.Type.Kind;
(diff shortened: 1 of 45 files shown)


MikeEdgar:
@phillip-kruger - if you get a chance to take a look at this PR later this week, note that the significant changes are in commits 2a99153 and 74844c9. I noticed that the Junit 5 tests were not running from the Maven CLI (but DO run from Eclipse), so the entire `extension-jaxrs` is now running Junit 5 with an updated version of the Surefire plugin.

phillip-kruger:
@MikeEdgar - LGTM, what about the others ? (Spring / vertx)

MikeEdgar:
@phillip-kruger - do you mean moving the other modules to Junit 5? They are still running fine as pure Junit 4. The issue with the JAX-RS tests was that they were a mix of 4 and 5 (which is supposed to work).

phillip-kruger:
Yes, I mean we move everything to 5, just for consistency ?

MikeEdgar:
I think we should do it, but maybe in a new PR so that the schema changes aren't lost in the noise. I'm happy to do a PR for that this week.

phillip-kruger:
Cool :)

phillip-kruger:
@MikeEdgar - I think we should merge and do a release so we can patch Quarkus . w.d.y.t. ?

MikeEdgar:
@phillip-kruger - agreed. I'm working on the `2.0.x` version of this right now. I'll open the PR asap.

phillip-kruger:
Thanks @MikeEdgar !

phillip-kruger (approved):

PR #703, opened 2021-02-19
SchemaProperty negative tests

Azquelt:
Add negative tests for `@SchemaProperty`

Also make two improvements revealed by testing:
* Log when a class is referenced as a schema from an annotation but isn't found in the index
* Correctly ignore non-numbers for `minimum` and `maximum`

In other cases, the tests assert the current behaviour. In most cases this means that invalid values from the annotations are put directly into the constructed document.

For #621


MikeEdgar:
Thanks @Azquelt - I'll try to look today. You may need to run via CLI and commit the formatting changes to pass the CI.

Azquelt:
Ah thanks, I'll do that

MikeEdgar:
Also, we just recently moved to Junit 5, so I'm afraid the tests will need to be updated with the newer annotations, and possibly the `LogCapture` functionality. You might consider rebasing on `master`, etc. Thank you!

Azquelt:
Ah rats, I thought I was up to date :(

I'll have to go and have a look at what the Junit 5 equivalent of `TestRule` is.

Azquelt:
@MikeEdgar I think that should be all done now.

MikeEdgar (approved):
Looks nice, thank you @Azquelt !

PR #729, opened 2021-03-07
Support OAS servers URLs and some general fixes

tarilabs:
As the title says.
If/once approved, I will cherry-pick to backport to `2.0.x` branch too.

## Demo 1 (no OAS servers defined)

If the OAS descriptors does not provide a `servers`, the application will not present any specific changes

In this example a simple Quarkus based app:
image

as before:
image

## Demo 2 (the OAS descriptors define servers)

If the OAS descriptor defines `servers` URLs, those are selectable via dropdown-box.

In this example a simple kie-server based container:
image

provides in the OAS descriptor some `servers` URLs, hence that is used to address the POST request:
image

Details:
image


MikeEdgar (approved):
Thanks @tarilabs - this looks good.

PR #731, opened 2021-03-07
Map `@JsonProperty(required = true)` properties to `required` array 

### LINKED ISSUE #714: Jackson annotation @JsonProperty(required=true) does not translate to @Schema(required=true)
My current classes are annotated with `@JsonProperty(required=true)` Jackson annotations. However, to mark those properties as required in the generated OpenAPI yaml, I need to add a `@Schema(required=true)' annotation. It would be nice if that property could be auto-generated.

Just for more context: I'm using smallrye-openapi together with kotlin, which has specific language features to distinguish between nullable and non-nullable types. Together with a dependency `com.fasterxml.jackson.module:jackson-module-kotlin`, the `@JsonProperty(required=true)` annotations are autogenerated already. Having the `@Schema` annotations also autogenerated would allow you to handle kotlin nullability/non-nullability at least for Jackson users.

MikeEdgar:
Fixes #714 

Note that `required = true` does not impact whether the property is `nullable`, per the `JsonProperty` docs.


phillip-kruger (approved):
Thanks @MikeEdgar !

PR #739, opened 2021-03-10
Sort scanned component entries, paths, and tags after all scanners

### LINKED ISSUE #735: Order list of classes when scanning annotations
I develop a Quarkus application where I use SmallRye OpenAPI to generate an OpenAPI definition. Now it is bugging me that every time I build the application the generated `openapi.yaml` file, the list of schemas can have a different order. This bugs me because I use the generated file in a Terraform resource for an Azure API Management API definition. Terraform picks up on the difference and consequently also updates the resource while applying the Terraform script.

I have tracked this down to `io.smallrye.openapi.jaxrs.JaxRsAnnotationScanner#processResourceClasses()` which calls `getJaxRsResourceClasses()` and the result is not sorted and can apparently have any order. So I am thinking if it would make sense to change `getJaxRsResourceClasses()` to apply a `sorted(Comparator<ClassInfo>)` operation which sorts the objects by name (fully qualified class name).

MikeEdgar:
Fixes #735 for `2.0.x`


phillip-kruger (approved):
This is awesome !! Thanks @MikeEdgar

PR #748, opened 2021-03-15
Properly handle single character property methods (2.0.x)

### LINKED ISSUE #746: maven plugin bug parsing getter of boolean fields with one letter names
The Maven plugin seems to have a problem with class fields whose name is only one letter long, e.g. 
```
public class TestClass {  
  private boolean b;
  public boolean isB() {return b;}  
  public void setB(boolean b) {this.b = b;}  
}
```

```
@GET
@Operation(description = "desc")
@Tag(name = "MyTag")
@APIResponse(
        responseCode = "200",
        description = "Something",
        content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = TestClass.class)))
public TestClass doSomething(){
  return null;
}
```
In this setup calling the smallrye plugin to generate a schema fails with `String index out of range: 0` . As far as I was able to debug it I believe that this is due to the `propertyName` method in `IgnoreResolver`, where it is implicitly assumed that any method fed into it either starts with "get" or "set", ignoring the fact that "isB" would be the standard name for the getter for a boolean named b. 
Perhaps we could check for getters starting with "is" before removing the prefix?

MikeEdgar:
Fixes #746 for `2.0.x`

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/IgnoreResolver.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/IgnoreResolver.java
@@ -174,8 +174,7 @@ public class IgnoreResolver {
                 return target.asField().name();
             }
             // Assuming this is a getter or setter
-            String name = target.asMethod().name().substring(3);
-            return Character.toLowerCase(name.charAt(0)) + name.substring(1);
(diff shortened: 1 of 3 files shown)


phillip-kruger (approved):
Thanks @MikeEdgar !!! Quick as always :)

PR #753, opened 2021-03-19
requestInterceptor and responseInterceptor may not be quoted

tisoft:
requestInterceptor and responseInterceptor need to be functions, so they may not be quoted


tisoft:
I tried to set the request-interceptor in quarkus:

`quarkus.swagger-ui.request-interceptor=req => {console.log(req);return req;}`

This was rendered as:

`                    requestInterceptor: 'req => {console.log(req);return req;}',`

Which does not work, as a real funtion is expected:

```
Fetch error
n.requestInterceptor is not a function /q/openapi
```

tisoft:
Just for reference, setting the in complete function via quarkus works:

`quarkus.swagger-ui.on-complete=req => {console.log('lala');return req;}`

This prints `lala` on the console on page load.

`                    onComplete: ${onComplete},`

is not quoted in the template.

MikeEdgar (approved):

tisoft:
If anyone needs a workaround, until this is merged:

`quarkus.swagger-ui.on-complete=function() {ui.getConfigs().requestInterceptor=req => {if(req.body){req.body=req.body.replaceAll('\\n','\\r');};return req;}}`

Since the on-complete function can be set, you can set the requestInterceptor inside the onComplete.

phillip-kruger (approved):

phillip-kruger:
Thanks !

PR #758, opened 2021-03-24
Adding Jakarta namaspace support

### LINKED ISSUE #751: Add suport for Jakarta namespace

phillip-kruger:
This PR adds support for the Jakarta namespace.

- Changed all Constants in JaxRS to be a List, returning both javax and jakarta for jax-rs,  jaxb, jsonb, security, bean validation and all code changes that goes with that.
- Changed the poms to use both the old javax artifact and the new jakarta artifact as test dependencies so that we can test both
- Changed all related tests to now test both javax (existing) and jakarta.

When doing a review it might be better to go through it per commit. The commits follow the explanation above

Fix #751


MikeEdgar (approved):
@phillip-kruger this is awesome! I definitely like the approach. Once I saw the pattern with the unit test changes I skipped over most of that, but the implementation changes look nice.

Issue #769, opened 2021-04-05
Add Instant to list of known datetime types.


damonsutherland:
```
    @GET
    @Path("/data/{field}")
    public Object getData(
        @PathParam("field") String sourceField,
        @QueryParam("id") Integer id,
        @QueryParam("account") Integer accountId,
        @QueryParam("date") LocalDate evalDate,
        @QueryParam("asKnownOn") @Schema(implementation = String.class, format = "date-time") Instant asKnownOn
    )
```

In the above example, the Instant argument is represented as an object:

```
{
  "nanos": 0,
  "seconds": 0,
  "epochSecond": 0,
  "nano": 0
}
```

I would expect it to be an example string in the format "yyyy-MM-ddThh:mm:ss.sssZ".

MikeEdgar:
@damonsutherland , wrapping the `@Schema` with `@Parameter(schema = ...)` should give you what you need. It does seem reasonable/natural to expect the annotation to work directly on the method argument, however. Also, we can likely add `Instant` to the list of known types.

damonsutherland:
@MikeEdgar Thanks.  Would you like me to create a feature request for the items you mention, or leave this ticket open?

MikeEdgar:
> @MikeEdgar Thanks. Would you like me to create a feature request for the items you mention, or leave this ticket open?

@damonsutherland - a feature request for `@Schema` support on resource method parameters would be great. The other feature for `Instant` is already implemented by @phillip-kruger :)

PR #773, opened 2021-04-07
Support `@Schema` annotation directly on parameters (2.0.x)

MikeEdgar:
As discussed in #769, same as #772 for 2.0.x (minus the Junit modifier change)


phillip-kruger (approved):
Thanks @MikeEdgar !

PR #788, opened 2021-04-25
Support JSON-B and Jackson property naming strategies

### LINKED ISSUE #712: Support for naming schemas (camelCase / sname_case) for properties
Hi,
I'm currently looking for support to globally change the naming strategy used.

All my classes use standard Kotlin camelCase notations, but the API itself should always be exported as snake_case. I configured Jackson globally, like so:

```
ObjectMapper().setPropertyNamingStrategy(PropertyNamingStrategies.SNAKE_CASE)
```

and I also tried adding an `@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy::class)` to all the data classes. Works like a charm for serialization/deserialization (either way). However, the exported OpenAPI definitions do not obey the settings (totally get it for the first via ObjectMapper one). I could in theory go with explicit naming via `@JsonProperty`, but given that this is a global setting everywhere else makes it a bit tedious.

Did you already have something like this in mind? I'd also be fine with adding some kind of configuration flag somewhere to make it possible.

MikeEdgar:
Fixes #712 

The naming strategy code is based on the Apache Johnzon [implementation](https://github.com/apache/johnzon/blob/master/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/PropertyNamingStrategyFactory.java).

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -22,8 +22,11 @@ mvn clean install
 * link:core[] - The core OpenAPI code, independent of entry point dependencies.
 * link:extension-jaxrs[] - The JAX-RS entry point. This module depends on JAX-RS and core.
 * link:extension-spring[] - The Spring entry point. This module depends on Spring and core.
+* link:extension-vertx[] - The Vert.x entry point. This module depends on Vert.x and core.
 * link:implementation[] - Implementation of the Eclipse MicroProfile OpenAPI specification. This just pulls in Core and the JAX-RS extension .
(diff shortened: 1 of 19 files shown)


phillip-kruger (approved):
Very nice !!! Thanks @MikeEdgar !

gwenneg:
Thanks @MikeEdgar!

lenalebt:
Very nice, thank you! This will simplify a lot of places where I manually set property names!

PR #795, opened 2021-05-04
Fix compatiblity with eclipse EE9 transformer.

pferraro:
The eclipse EE9 transformer (https://projects.eclipse.org/projects/technology.transformer), which is used by WildFly (and other EE8+9 servers), performs runtime transformation of "javax" -> "jakarta", including the strings in JaxRsConstants.
This results in lists containing 2 identical elements, which causes all sorts of issues, since the smallrye-open-api codebase assumes element uniqueness of these lists.
Converting these lists to sets solves the problem.


MikeEdgar (approved):
Looks good to me, thanks @pferraro

PR #810, opened 2021-05-28
Improve handling of `Optional` with array schema types

### LINKED ISSUE #809: Optional of List does not generate proper schema ref
Hi there,

In our application we rely on Optional<...> for each of our models properties to allow us to distinguish between "undefined" vs. "null" vs. "an actual value". When generating OpenAPI all schemas are properly referenced except for array-like data structure.

For example, using WildFly 23.0.2.Final which bundle SmallRye OpenAPI 2.1.2:
```
@Path("/aresources")
public class AResource {
    public static class A {
        public UUID id;
        public Optional<B> optionalOfB;
        public List<B> listOfB;
        public Optional<List<B>> optionalListOfB;
    }

    public static class B {
        public UUID id;
    }

    @GET @Path("/") @Operation(summary = "List A")
    public List<A> listA() {
        return Collections.emptyList();
    }
}
```

will generate the following yaml (excerpt):
```
openapi: 3.0.3
components:
  schemas:
    A:
      type: object
      properties:
        id:
          $ref: '#/components/schemas/UUID'
        listOfB:
          type: array
          items:
            $ref: '#/components/schemas/B'
        optionalListOfB:
          type: object
          nullable: true
        optionalOfB:
          allOf:
          - $ref: '#/components/schemas/B'
          - nullable: true
    B:
      type: object
      properties:
        id:
          $ref: '#/components/schemas/UUID'
```

but should have instead:
```
openapi: 3.0.3
components:
  schemas:
    A:
      type: object
      properties:
        id:
          $ref: '#/components/schemas/UUID'
        listOfB:
          type: array
          items:
            $ref: '#/components/schemas/B'
        optionalListOfB:
          type: array
          items:
            $ref: '#/components/schemas/B'
          nullable: true
        optionalOfB:
          allOf:
          - $ref: '#/components/schemas/B'
          - nullable: true
    B:
      type: object
      properties:
        id:
          $ref: '#/components/schemas/UUID'
```

I've also tested out the latest SmallRye OpenAPI 2.1.4, which had lead to the exact same result.

My guess is #210 does not cover this specific use case and/or there's an issue within the `io.smallrye.openapi.runtime.util.TypeUtil`.

Let me know if I can be any help to resolve this issue; this is last one preventing us migrating from Swagger to OpenAPI.

Thank you!

MikeEdgar:
Fixes #809

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeProcessor.java
@@ -79,12 +79,13 @@ public class TypeProcessor {
             type = resolveTypeVariable(schema, type, false);
         }
 
-        if (type.kind() == Type.Kind.ARRAY) {
-            readArrayType(type.asArrayType(), this.schema);
+        if (TypeUtil.isWrappedType(type)) {
+            // Unwrap and proceed using the wrapped type
+            type = TypeUtil.unwrapType(type);
         }
 
-        if (TypeUtil.isWrappedType(type)) {
-            return readWrappedType(type, this.schema);
+        if (type.kind() == Type.Kind.ARRAY) {
+            return readArrayType(type.asArrayType(), this.schema);
         }
 
         if (isA(type, ENUM_TYPE) && index.containsClass(type)) {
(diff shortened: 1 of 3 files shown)


MikeEdgar:
@matlach - the optional types are not being properly unwrapped. This change should fix the incorrect schema you reported in #809 .

matlach (approved):
This all looks fine to me, thank you so much for being so quick to address this issue.

phillip-kruger (approved):
Thanks @MikeEdgar

PR #820, opened 2021-06-09
Use resource class for class-based operation naming strategies

### LINKED ISSUE #819: operationId auto generation fails for inheritance
When I use `CLASS_METHOD` as operationId generation strategy, and my resource implements an interface, the interface name is used for operationId instead of the concrete jaxrs resource class's name.

So, in the case of

`GreetingInterface` -> `GreetingResource`

it should generate: `operationId: GreetingResource_hello`,
but it does: `operationId: GreetingInterface_hello`

This is a problem since when I have for example

`GreetingResourceA implements GreetingInterface`
`GreetingResourceB implements GreetingInterface`

I will get duplicate operationId errors when generating a client (`GreetingInterface_hello`, `GreetingInterface_hello`).

My concrete use case is an annotated interface (and implementation) for basic CRUD operations so I don't need to repeat them for each model.

I tried to do some digging and testing, but couldn't get the existing unit test to pick up my example resource (result was always empty).

Here is the implementation and the test in the smallrye-openapi code:

https://github.com/smallrye/smallrye-open-api/blob/main/core/src/main/java/io/smallrye/openapi/runtime/io/operation/OperationReader.java#L145

https://github.com/smallrye/smallrye-open-api/blob/main/extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/OperationIdTest.java

Here are an example and a link to a quarkus example project.

Example:

```java
public interface GreetingInterface {

    @GET
    public String hello();

}
```

```java
@Path("/hello")
public class GreetingResource implements GreetingInterface {

    public String hello() {
        return "Hello RESTEasy";
    }

}
```

Result:
```json
---
openapi: 3.0.3
info:
  title: Generated API
  version: "1.0"
paths:
  /hello:
    get:
      operationId: GreetingInterface_hello
      responses:
        "200":
          description: OK
          content:
            text/plain:
              schema:
                type: string
```

expected: `operationId: GreetingResource_hello`
actual: `operationId: GreetingInterface_hello`

Example Repo URL: https://github.com/codegourmet/openapi-operationid-example

MikeEdgar:
Fixes #819 

- Move operation Id strategy logic to scanner after reading operation annotation

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/operation/OperationReader.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/operation/OperationReader.java
@@ -6,7 +6,6 @@ import org.jboss.jandex.MethodInfo;
 
 import com.fasterxml.jackson.databind.JsonNode;
 
-import io.smallrye.openapi.api.OpenApiConfig;
 import io.smallrye.openapi.api.models.OperationImpl;
 import io.smallrye.openapi.runtime.io.IoLogging;
 import io.smallrye.openapi.runtime.io.JsonUtil;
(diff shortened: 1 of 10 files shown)


phillip-kruger (approved):
Thanks @MikeEdgar !

PR #822, opened 2021-06-09
Decode Json string from example fields in Open API annotation to a json object

### LINKED ISSUE #821: Json formatted example in annotation are not decoded
Currently all openAPI examples written in JSON in annotation are not decoded and stored as an escaped string.

Following the Open API specification for the example Object, the annotation value can be a Json string but should be stored as Json Object. https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#example-object

**Code**
```java

public class JsonExampleApiResourceTestResource {

    @GET
    @APIResponse(responseCode = "200", content = @Content(mediaType = MediaType.APPLICATION_JSON, examples = {
            @ExampleObject(name = "response_success_1", value = "{\"key\":\"value\"}"),
            @ExampleObject(name = "response_success_2", value = "{\"key2\":12}")
    }, example = "{\"key\":\"value\"}"), description = "Description 200")
    public Pet getPet() {
        return null;
    }

}
```
**Expected**
Json example value should be decoded and stored as Json.
```json
"responses" : {
    "200" :  {
      "description" : "Description 200",
      "content" : {
        "application/json" : {
          "example" : {
              "key" : "value"
          },
          "examples" : {
            "response_success_1" : {
              "value" : {
                "key" : "value"
              }
            },
            "response_success_2" : {
              "value" : {
                "key2" : 12
              }
            }
          }
        }
      }
    }
  }
```

**Actual**
Json example value are stored as string with escaped characters.
```json
"responses" : {
    "200" :  {
      "description" : "Description 200",
      "content" : {
        "application/json" : {
          "example" : "{\"key\":\"value\"}",
          "examples" : {
            "response_success_1" : {
              "value" : "{\"key\":\"value\"}"
            },
            "response_success_2" : {
               "value" : "{\"key2\":12}"
            }
          }
        }
      }
    }
  }
```

devnied:
Closes #821

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/components/ComponentsReader.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/components/ComponentsReader.java
@@ -52,7 +52,7 @@ public class ComponentsReader {
         // TODO for EVERY item below, handle the case where the annotation is ref-only.  then strip the ref path and use the final segment as the name
         components.setCallbacks(
                 CallbackReader.readCallbacks(context, nested.value(ComponentsConstant.PROP_CALLBACKS)));
-        components.setExamples(ExampleReader.readExamples(nested.value(ComponentsConstant.PROP_EXAMPLES)));
+        components.setExamples(ExampleReader.readExamples(context, nested.value(ComponentsConstant.PROP_EXAMPLES)));
         components.setHeaders(HeaderReader.readHeaders(context, nested.value(ComponentsConstant.PROP_HEADERS)));
         components.setLinks(LinkReader.readLinks(nested.value(ComponentsConstant.PROP_LINKS)));
         components.setParameters(
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/example/ExampleReader.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/example/ExampleReader.java
@@ -17,6 +17,8 @@ import io.smallrye.openapi.runtime.io.IoLogging;
 import io.smallrye.openapi.runtime.io.JsonUtil;
 import io.smallrye.openapi.runtime.io.Referenceable;
 import io.smallrye.openapi.runtime.io.extension.ExtensionReader;
+import io.smallrye.openapi.runtime.scanner.AnnotationScannerExtension;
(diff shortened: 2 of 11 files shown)


devnied:
I still have an issue with `CustomExtensionParsingTests` unit test but I don't understand the goal of this test (Check if Jackson is not in the classpath).
If someone could enlighten me on this subject ?

phillip-kruger (approved):
LGTM

phillip-kruger:
@MikeEdgar please double check.

MikeEdgar:
I'll need a little time to review more closely, but we need to be careful about using `JsonUtil` as it pulls in the hard dependency on Jackson. Some downstream projects use other JSON libraries which is the purpose of those additional tests with jackson excluded.

phillip-kruger:
Can this be done without the Util ?

MikeEdgar:
I think it needs to be done like extensions with a default method in the `AnnotationScannerExtension` that downstream can override and use something besides the `JsonUtil`.

MikeEdgar:
@devnied - please try this:

- Add a method `parseJson(String value)` to `AnnotationScannerException` that returns `Object`. The default implementation should just call `JsonUtil.parseValue(value)`
- Add a static method to `ExampleReader` that takes `AnnotationScannerContext context` and `String value` and loops over all `AnnotationScannerExtension`s in the `context` and returns the first non-null `Object` returned by `context.parseJson` (see `ExtensionReader#readExtensionValue` for an example of that). 
- In each of the locations where you're calling `JsonUtil` to obtain the example now, replace it with a call to the new static method in `ExampleReader`, passing the context and the `String` value.

devnied:
@MikeEdgar thanks for the feedback I will try that

devnied:
@MikeEdgar Using `AnnotationScannerExtension` work's well but I still have the hard dependency on Jackson in the classpath.

phillip-kruger:
CI now passing. @MikeEdgar  ? W.d.y.t ?

MikeEdgar (approved):
This looks good - thank you @devnied.

FYI @tjquinno for updating to SmallRye OpenAPI 2.1.6+. There is a new method to override to avoid utilizing Jackson for JSON parsing.

phillip-kruger (approved):

PR #829, opened 2021-06-21
Disable non-public properties, rather than private

### LINKED ISSUE #825: Non-public properties are included the generated OpenAPI document
While an option was added to remove _private_ properties for #418, protected and package-scoped properties are still generated.

To be consistent with JSON-B (and Jackson in its default configuration), [only public properties][jsonb-only-public] should be included in object schemas.

Could we expand the existing config property to cover all non-public properties, or add additional config properties to cover protected and package scoped fields?

[jsonb-only-public]:https://jakarta.ee/specifications/jsonb/2.0/jakarta-jsonb-spec-2.0.html#scope-and-field-access-strategy

Azquelt:
Update the private-properties.enable config property to enable or
disable non-public properties, rather than just private ones.

Fixes #825 

There wasn't much feedback on the issue so I've gone ahead and created this PR which changes the meaning of `private-properties.enable` to affect all non-public properties. Let me know if you'd prefer this was fixed in a different way, e.g. by introducing separate properties to configure protected or package scoped properties.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
@@ -495,9 +495,9 @@ public class TypeResolver {
             properties.values()
                     .stream()
                     .filter(TypeResolver::isExposedByDefault)
-                    .filter(resolver -> isPrivateOrAbsent(resolver.field))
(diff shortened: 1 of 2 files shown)


Azquelt:
The build check failed to download an artifact from maven central, repushed to prompt it to have another go...

MikeEdgar (approved):
Looks good to me, thanks @Azquelt

PR #859, opened 2021-07-22
Add targets with `@NotBlank` and `@NotEmpty` to `required` properties

### LINKED ISSUE #851: @NotBlank doesn't make the field required
Using `@NotBlank` on a field doesn't make it required as `@NotNull` does.

MikeEdgar:
Fixes #851


phillip-kruger (approved):
Thanks @MikeEdgar !

Issue #857, opened 2021-07-19
How to post list of files?


deepfunction:
I try to post body with object and list of files.
![1](https://user-images.githubusercontent.com/48287188/126158045-ce9d0b15-23c1-4434-84a4-22e088f42e00.jpg)
![2](https://user-images.githubusercontent.com/48287188/126158134-a1bce558-9b45-4600-a01b-2ad314bdd1ae.jpg)
And what i see in the swagger ui.
![3](https://user-images.githubusercontent.com/48287188/126158168-79d6a0a4-763e-4f20-be37-19eb587a5f82.jpg)
So, how to create something like this
![4](https://user-images.githubusercontent.com/48287188/126158331-828835dc-572c-47c6-b471-9d4829ffb119.jpg)

MikeEdgar:
@deepfunction - does using a `List` of `File`s actually work at runtime? Did you try a list of `InputStream` or `byte[]` ? I think this will require a change to the scanner.

PR #871, opened 2021-08-03
Use `type: string` with `format: binary` for `InputStream` and `byte[]`

MikeEdgar:
Looking at #857, I realized that we are not supporting the `binary` string format (https://swagger.io/specification/#data-types). @phillip-kruger, does this seem reasonable? The change is to default `InputStream` and `byte[]` to be strings of format `binary`. As mentioned in that issue, I'm don't think that `File` can be used for JAX-RS.


phillip-kruger (approved):
LGTM

MikeEdgar:
@phillip-kruger - added another commit with a change to skip the default type attributes if a `type` properties has been given in the `@Schema` annotation.

phillip-kruger (approved):

PR #872, opened 2021-08-04
Ignore array type when schema annotation specifies non-array

### LINKED ISSUE #831: generated items are not removed, when changing schema type via @Schema
Hi, while trying to work around an issue with `byte[]` representation in the generated openapi, I stumbled across something that feels like a bug to me.

POJO:
```
public class Sample {
    @Schema( type = SchemaType.STRING, format = "base64" )
    public byte[] data;
}
```

The generated schema:
```
components:
  schemas:
    Sample:
      type: object
      properties:
        data:
          format: base64
          type: string
          items:
            format: byte
            type: string
```

I think, only `type: array` schemas should have the `items` property.

I have been able to work around that using the following `OASFilter`:

```
public class OpenApiConfig implements OASFilter {
    @Override
    public Schema filterSchema( Schema schema ) {
        if ( schema.getType() != SchemaType.ARRAY ) {
            schema.setItems( null );
        }
        return schema;
    }
}
```

Is this something that could be fixed?

MikeEdgar:
Fixes #831


phillip-kruger (approved):
Thanks @MikeEdgar !

phillip-kruger (approved):

PR #873, opened 2021-08-04
Add JDK 16 to build matrix (test latest in addition to LTS)

MikeEdgar:
@phillip-kruger - it would be good to have the latest JDK and any LTS we want to support in the build matrix (i.e. replace 16 with 17 when released). w.d.y.t? Maybe this will fix the build from thinking that it still needs to build JDK 8 :-)


phillip-kruger:
Nice !

PR #870, opened 2021-08-02
Add multipleOf test

Azquelt:
`0.1` is used as the value here as it's not unusual and jandex currently has issues reading it.

This test won't pass until a new version of jandex is released and dependencies updated.

For #869

PR #888, opened 2021-08-26
Default to unsorted properties, add option to use sorted properties

### LINKED ISSUE #87: Don't automatically sort object fields

MikeEdgar:
Fixes #87 . Makes the default object property order the same as the order from the class file, as read by Jandex 2.4.0. This also adds a new option to restore the previous sorted property order. Can not be merged without smallrye/smallrye-parent#270 and a new parent version.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
+++ b/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
@@ -91,6 +91,10 @@ public interface OpenApiConfig {
         return JsonbConstants.IDENTITY;
     }
 
+    default boolean sortedPropertiesEnable() {
+        return false;
+    }
+
     default Map<String, String> getSchemas() {
         return new HashMap<>();
     }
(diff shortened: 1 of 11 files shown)


MikeEdgar:
@phillip-kruger , once this is merged, I'll update #870 as well.

phillip-kruger (approved):

Issue #896, opened 2021-09-02
The absence of annotation confuses openapi/swagger


cescoffier:
With resteasy reactive, you can omit some annotations.
For example, the following endpoint is valid:

```
@GET
@Path("/{id}")
public Response getOne(long id) {
    Movie movie = Movie.findById(id);
    if (movie == null) {
        return Response.status(404).build();
    } else {
        return Response.ok(movie).build();
    }
}
```

However, it confuses openapi/swagger and we can't invoke it from the Quarkus dev console:

![image](https://user-images.githubusercontent.com/402301/131792330-5b57e416-93ea-4387-be68-50d1542dc42e.png)

cescoffier:
CC @phillip-kruger @geoand

phillip-kruger:
We will have to add support for that in openapi. At the moment it will assume that is the body

MikeEdgar:
This is definitely a bug. There are unit tests specifically for this.

phillip-kruger:
So the problem is not that it did not create the path param, that actually worked (see below). The problem is that it also added the response body for that param, and that should be omitted.

```
/movies/{id}:
    get:
      tags:
      - Movie Resource
      parameters:
      - name: id
        in: path
        required: true
        schema:
          format: int64
          type: integer
      requestBody:
        content:
          application/json:
            schema:
              format: int64
              type: integer
      responses:
        "200":
          description: OK
```

phillip-kruger:
mmm, I can not recreate this in our tests in SmallRye....

phillip-kruger:
Here the test in SmallRye, seems ok https://github.com/smallrye/smallrye-open-api/pull/899

I am still trying to recreate this.

PR #899, opened 2021-09-03
Added tests for #896

phillip-kruger:


MikeEdgar (dismissed):
@phillip-kruger - these specifically test the un-annotated parameter with `@Path` on the method, right?

phillip-kruger:
Yes, I am adding one more test with returning Response, so that it is exactly like the bug report

phillip-kruger:
Ok now we have a test exactly like the bug report. So this seems to be an issue in Quarkus ?!? I am going to add some tests there to see

MikeEdgar (approved):

PR #919, opened 2021-09-19
Ignore synthetic implementations of annotated interfaces

### LINKED ISSUE #915: Not hiding RestClient operations
see https://github.com/geoand/quarkus-java17-demo/pull/1

MikeEdgar:
Fixes #915 

@phillip-kruger - this worked for me with a custom-built index from the example project linked from #915. If we want a new unit test it would likely involve some custom `ClassInfo` with the synthetic flag set. Let me know what you think.


phillip-kruger (approved):
Awesome! Thanks @MikeEdgar !

PR #958, opened 2021-11-07
maven plugin - add option to attach schemas as build artifacts

Postremus:
Often, the schema from service a is used in project b, to e.g. generate a rest client.
This process is easier, when the schema is just available as another dependency.

This adds an option to attach all generated schemas as build artifacts, using following gavct:
project-groupid:project-artifactid:project-version:schemaName:json|yaml


phillip-kruger (approved):
Thanks !

PR #957, opened 2021-11-07
maven plugin - Reduce amount of created jandex indexes.

Postremus:
This introduces a blacklist of artifacts, for which no jandex index is ever created, since they won't have impact on the generated scheme.
Also, cache the indexes that are created using their gavct as key.

I tested this with a project with 2 modules, each module having 4 executions of the maven-plugin.
[sroa-scanpackages.zip](https://github.com/smallrye/smallrye-open-api/files/7492927/sroa-scanpackages.zip)
I ran `mvn clean compile` 3 times, taking the average reported by mvn as total time.

2.1.15: 19.067s
only blacklist: 7.003s
only cache: 5.471s
blacklist and cache: 4.447s

closes #954

PR #959, opened 2021-11-08
maven plugin - Generate schema during incremental compilation

Postremus:
This triggers the schema build, whenever m2e decides to do an incremental build.
Right now, it is not possible to use the incremental build api (BuildContext class), since it is not compatibly with current maven versions. Therefore, index caching of the maven project artifact itself is removed for now.
This also adds the necessary metadata for eclipse to circumvent the infamous error "Plugin execution not covered by lifecycle configuration".

As reference, please see: 
* https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html
* https://github.com/eclipse-m2e/m2e-core/issues/364#issuecomment-939987848

Draft for now, since this depends on #957 being merged first. That branch also fixes some NPEs when running inside m2e, e.g. Parameter${project.build.outputDirectory} resolves as null, but using the (imo cleaner) MavenProject API works fine.

demo-build-on-save-eclipse


phillip-kruger (approved):
Great ! Thanks

phillip-kruger:
It would be good to get some documentation / blog post or something about this ?

PR #965, opened 2021-11-10
Add integration test for the maven-plugin info parameters

Postremus:
Adds a basic test which makes sure that the infoX Parameters from the plugin are correctly used in the openapi schema.

Uses the integration testing framework extension to make integration tests for the maven-plugin possible.
The test pom has to be in resources-its.
The directory structure is based upon the Fully Qualified Class Name of the Test Class, and the last directory is named after the test.

This also adds everything needed (imo) to test against yaml / json schemas (SchemaTestBase.java), which makes it easier to later write more integration tests.

For reference:
* https://github.com/khmarbaise/maven-it-extension
* https://blog.soebes.de/blog/2020/08/18/itf-part-i/

### DIFF
--- a/tools/maven-plugin/pom.xml
+++ b/tools/maven-plugin/pom.xml
@@ -14,6 +14,8 @@
     
     <properties>
         <sonar.skip>true</sonar.skip>
+
+        <itf-maven.version>0.11.0</itf-maven.version>
     </properties>
     
     <dependencies>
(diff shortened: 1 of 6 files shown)


phillip-kruger (approved):
LGTM.

phillip-kruger:
Thanks @Postremus !

PR #967, opened 2021-11-11
maven-plugin make it easier to scan dependend schemas

### LINKED ISSUE #952: maven-plugin make it easier to scan dependend schemas
I want to create multiple openapi schemas from the same maven module.
One of the schemas is for public consumption, the other one is internal.

My source tree looks like this:
image

For the internal schema, I just define `<scanPackages>^org.acme</scanPackages>`. All of the classes needed are found.
For the public schema, I have to define `<scanPackages>^org.acme.pub|org.acme.model</scanPackages>`. 
Otherwise only the resources in org.acme.pub are found, and I would get following warning:
[INFO] --- smallrye-open-api-maven-plugin:2.1.15:generate-schema (public-openapi) @ sroa-maven-plugin ---
Nov. 04, 2021 8:08:52 VORM. io.smallrye.openapi.runtime.scanner.OpenApiDataObjectScanner process
WARN: SROAP04005: Could not find schema class in index: org.acme.model.SomeModel

This is only a simplified example of my source tree. In my real project, I have the model classes sometimes in different artifacts (e.g. enums), which would quickly make this configuration unreadable.

Is there any way to only specifiy the location of the resource classes `org.acme.pub`, and let the maven plugin figure out all dependend schemas itself?

Reproducer project:
[sroa-maven-plugin.zip](https://github.com/smallrye/smallrye-open-api/files/7472919/sroa-maven-plugin.zip)

Postremus:
Added new options scan.profiles and scan.exclude.profiles to filter the operations that are kept in the schema.
Only one of the configured profiles is needed to either include, or exclude, the operation. Exclusions are checked first, then inclusions. Extensions containing profiles have to start with "x-smallrye-profile", and will not be included in the openapi document.
Naming of these new options are kept in line with the openapi scan.package and scan.external.package options.

closes #952 

demo-annotation-inclusion


core/src/main/java/io/smallrye/openapi/runtime/io/operation/OperationReader.java
```diff
@@ -106,7 +107,27 @@ public static Operation readOperation(final JsonNode node) {
...
+
+        if (!context.getConfig().getScanAnnotations().isEmpty()) {
+            for (DotName scanAnnotation : context.getConfig().getScanAnnotations()) {
+                if (method.hasAnnotation(scanAnnotation)) {
+                    return false;
+                }
+            }
+
+            return true;
+        }
```
Postremus, line 129: Not quite happy how these inclusion / exlusion checks turned out, but did not find a cleaner way.

core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
```diff
@@ -10,6 +10,7 @@
 
 import io.smallrye.openapi.api.constants.JsonbConstants;
 import io.smallrye.openapi.api.constants.OpenApiConstants;
+import org.jboss.jandex.DotName;
```
phillip-kruger, line 13: I am not sure if we want Jandex in the Config ? @MikeEdgar  ? Maybe also use pattern or even just String

MikeEdgar: Yeah, I agree that the config should stick to JDK types. DotName is effectively a string anyway, from a configuration perspective.

MikeEdgar:
@Postremus - thank you for putting this together, definitely some cool functionality. The thing I am wondering is, can it be done using standard annotations? For example, instead of a custom annotation, what if the `@Extension` annotation was used with names like `x-smallrye-profile-external` or `x-smallrye-profile-internal`. That would allow the user to assign a "profile" to endpoint methods that could be selected in the config - in this case `external` and `internal`. Extensions starting with `x-smallrye-profile-` would be dropped from the final result. It would also have the benefit of being usable with the static document.

Just an idea.

Postremus:
@MikeEdgar
I like your idea.
This allows the user to simply set the include (or exclude) profiles using configuration. On the operation or paths, the profile extensions  can be specified like this:
```
@Extension(name = "x-smallrye-profile-external", value = "")
```

Way simpler to use, and better user experience. I will update the implementation accordingly.

Thank you!

core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
phillip-kruger: I think you can remove this import of `DotName`

phillip-kruger (changes requested):

Postremus:
Hi @MikeEdgar, I implemented the approach using profiles defined as part of extensions, like you suggested.
Please review again. Thanks!

phillip-kruger:
@MikeEdgar please can you have a look ?

MikeEdgar:
Thanks @Postremus. I'm planning on looking at this later today or tomorrow morning.

core/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
```diff
@@ -408,6 +410,30 @@ public void doAllowNakedPathParameter() {
         return defaultConsumes;
     }
 
+    @Override
+    public Set<String> getScanProfiles() {
+        if (scanProfiles == null) {
+            String classes = getStringConfigValue(OpenApiConstants.SCAN_PROFILES);
+            if (classes == null) {
+                classes = getStringConfigValue(OpenApiConstants.SCAN_PROFILES);
+            }
```
MikeEdgar, line 419: Not required, but if you re-base again you might rename this variable and the equivalent in the "exclude" method.

MikeEdgar (approved):
This looks good, thanks @Postremus . I left one small comment that you might update if you rebase again, but otherwise this looks reasonable. At some point it might be worthwhile to rearrange how things are scanned in the JAX-RS (and other) scanners so that the profiles are checked more pre-emptively, but I wouldn't worry about that now.

Issue #969, opened 2021-11-17
Kotlin Continuations should be excluded from requestBody


rgmz:
_Cross-posting this from:_ https://github.com/quarkusio/quarkus/issues/21509

This appears to relate to #893. It should be possible to fix by filtering `kotlin.coroutines.Continuation` from `requestBody`. I'm happy to create a PR, if someone can point me in the right direction. :)

MikeEdgar:
@rgmz - thank you. If you can dig in and see why the code change in #920 isn't working for this case, that would help a lot. You should be able to add a new test case similar to the one added in that PR that adjusts the endpoint to match the Java-equivalent of the Kotlin endpoint.

rgmz:
@MikeEdgar: I opened #970 to troubleshoot this. The schema generates as expected in the tests, so I'll need to dig deeper to figure out what's going on...

MikeEdgar:
There's a reason why the Jax-RS extension doesn't show up in the dep tree. @phillip-kruger, can you remind me again?

phillip-kruger:
Maybe for testing ? I am not sure

MikeEdgar:
It may be because it's only used by the Quarkus internals during the scan and it's not actually depended on for the project to compile or run. Something along those lines...

rgmz:
I can troubleshoot the [Quarkus extension](https://github.com/quarkusio/quarkus/tree/main/extensions/smallrye-openapi) and add some tests. Any pointers on where to get started?

For example: what are the functions of the [`deployment`](https://github.com/quarkusio/quarkus/tree/main/extensions/smallrye-openapi/deployment), [`runtime`](https://github.com/quarkusio/quarkus/tree/main/extensions/smallrye-openapi/runtime), and  [`spi`](https://github.com/quarkusio/quarkus/tree/main/extensions/smallrye-openapi/spi) modules? Which one should I be looking into?

rgmz:
I _think_ I'm on the right track, but I'm unable to run the tests due to a confusing error.

```sh
$ ./mvnw test -f extensions/smallrye-openapi/deployment/ -Dtest=OpenApiPathWithSegmentsTestCase
...
[...]

PR #970, opened 2021-11-17
Add test for Kotlin continuation in requestBody

rgmz:
Two things to note:

1. The schema generated by the [reproducer application](https://github.com/rgmz/quarkus-kotlin-openapi-21509) (which I used to model the `Resource` class) differs from that generated in tests. It's possible that: **a)** it's a Quarkus-specific issue; **b)** the test aren't accurately representing "real" Kotlin class files; **c)** there's an additional step not currently performed in the tests which cases this.

   ```yml
	---
	openapi: 3.0.3
	info:
	  title: quarkus-kotlin-openapi API
	  version: 1.0.0-SNAPSHOT
	paths:
	  /hello:
	    get:
	      tags:
	      - Reactive Greeting Resource
	      requestBody:
		content:
		  application/json:
		    schema:
		      $ref: '#/components/schemas/ContinuationSuperString'
	      responses:
		"200":
		  description: OK
		  content:
		    text/plain:
		      schema:
			type: string
	components:
	  schemas:
	    ContinuationSuperString:
	      type: object
	      properties:
		context:
		  $ref: '#/components/schemas/CoroutineContext'
	    CoroutineContext:
	      type: object
    ```

2. Using the Kotlin Decompiler in IntelliJ produces code with raw `Continuation` types. The schema works as expected when I manually change those to `Continuation<? super String>`, however, this leads me to believe that the cause may be **(b)**.

   `Continuation<? super String>` works as expected:
   ```yml
    openapi: 3.0.3
    paths:
      "/hello":
        get:
          responses:
            '200':
              description: OK
              content:
                text/plain:
                  schema:
                    type: string
   ```
   
   With `Continuation`
   ```yml
    openapi: 3.0.3
    paths:
      "/hello":
        get:
          responses:
            '200':
              description: OK
              content:
                text/plain:
                  schema:
                    type: object
   ```


MikeEdgar:
If I remember correctly, the Kotlin to Java function in Intellij drops the generic type information from the bytecode. I think if you look at the Kotlin class in the bytecode viewer it's apparent that the generic information is still there in the method attributes.

rgmz:
> I think if you look at the Kotlin class in the bytecode viewer it's apparent that the generic information is still there in the method attributes.

Ah, you're right:
```
  // access flags 0x1
  // signature (Lkotlin/coroutines/Continuation<-Ljava/lang/String;>;)Ljava/lang/Object;
  // declaration:  hello(kotlin.coroutines.Continuation<? super java.lang.String>)
  public hello(Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```

Given that the tests are working as expected, I'm thinking that it's a Quarkus issue after all (see my comment in #969).

MikeEdgar:
@rgmz - just to test get more information, maybe see if you can create a jar of the problematic project, add it as a test dependency to the unit test and attempt to scan those classes directly.

rgmz:
@MikeEdgar: Excellent idea.

That generates the expected response, which furthers my suspicion that the jaxrs extension isn't being applied in `quarkus-smallrye-openapi`.

Test:
```java
@Test
void testKotlinContinuationRequestBody() throws IOException, JSONException {
     Index index = Index.of(ReactiveGreetingResource.class, String.class);
    OpenApiAnnotationScanner scanner = new OpenApiAnnotationScanner(emptyConfig(), index);
    OpenAPI result = scanner.scan();
    printToConsole(result);
    assertJsonEquals("responses.kotlin-continuation-no-body.json", result);
}
```

Schema:
```json
{
  "openapi" : "3.0.3",
  "paths" : {
    "/hello" : {
      "get" : {
        "responses" : {
          "200" : {
            "description" : "OK",
            "content" : {
              "text/plain" : {
                "schema" : {
                  "type" : "string"
                }
              }
            }
          }
        }
      }
    }
  }
}
```

phillip-kruger:
@MikeEdgar - This build still fails ? Can you have a look ?

MikeEdgar:
Let's close this for now. @rgmz , please let us know if the issue is not resolved when 2.1.16 is released.

PR #980, opened 2021-11-29
Update open-api-ui-forms npm dependencies and build node version

MikeEdgar:
@tarilabs - let me know if you have any concerns with this update. The new major version of rjsf did not seem to break anything.


MikeEdgar:
@tarilabs - I agree, I'm not sure about automated tests for it... probably just manual smoke tests for non-trivial dependency updates? I mentioned this one because of the major version upgrade (it doesn't look like it uses any of the APIs that were "broken" though).

PR #983, opened 2021-12-01
Set properties read/write only using TypeResolver property data

### LINKED ISSUE #982: Does not include properties annotated with @JsonbTransient in schema
In our projects we use JAX-RS, JSON-B and BeanValidation
When programming POST requests we usually transmit Entity object directly which usually has id field.
For security purposes we set @JsonbTransient on setId method.

Generated UI shows that you can pass id key to the POST request body. It would be nice to hide that field.

MikeEdgar:
Fixes #982


phillip-kruger (approved):
Thanks @MikeEdgar !

PR #997, opened 2021-12-09
maven-plugin - ignore resteasy-core

### LINKED ISSUE #994: maven plugin generates additional operations
Hi,

I'm using the plugin v2.1.15 in quarkus 2.4 api.
When querying /q/api, I have exactly operations of my code but when executing the plugin the produced json/yaml contains a /{job-id} operation

```
/{job-id}:
    get:
      parameters:
      - name: job-id
        in: path
        required: true
        schema:
          type: string
      - name: wait
        in: query
        schema:
          format: int64
          default: -1
          type: integer
      responses:
        "200":
          description: OK
    post:
      parameters:
      - name: job-id
        in: path
        required: true
        schema:
          type: string
      - name: wait
        in: query
        schema:
          format: int64
          default: -1
          type: integer
      responses:
        "200":
          description: OK
    delete:
      parameters:
      - name: job-id
        in: path
        required: true
        schema:
          type: string
      responses:
        "204":
          description: No Content
```

I have search for job-id in my code without result. Is there a way to find out where the plugin find these operations ?

Postremus:
It contains resources in org.jboss.resteasy.core.AsynchronousDispatcher, which we would include in the scheme.
This artifact is not part of the quarkus CombinedIndexBuildItem, so it seems ok to also ignore it for the maven plugin index.

closes #994

### DIFF
--- a/tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/MavenDependencyIndexCreator.java
+++ b/tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/MavenDependencyIndexCreator.java
@@ -59,6 +59,7 @@ public class MavenDependencyIndexCreator {
         ignoredArtifacts.add("com.thoughtworks.xstream:xstream");
         ignoredArtifacts.add("com.github.javaparser:javaparser-core");
         ignoredArtifacts.add("org.jboss:jandex");
+        ignoredArtifacts.add("org.jboss.resteasy:resteasy-core");
 
         ignoredArtifacts.add("antlr");
         ignoredArtifacts.add("io.netty");


tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/MavenDependencyIndexCreator.java
```diff
@@ -59,6 +59,7 @@ public MavenDependencyIndexCreator() {
         ignoredArtifacts.add("com.thoughtworks.xstream:xstream");
         ignoredArtifacts.add("com.github.javaparser:javaparser-core");
         ignoredArtifacts.add("org.jboss:jandex");
+        ignoredArtifacts.add("org.jboss.resteasy:resteasy-core");
```
Postremus, line 62: In the future, it might be better to remove the whole static artifact blacklist from the maven plugin.
We should do the same as quarkus does for building the CombinedIndexBuildItem.
* search in each dependency for Marker files (like META-INF/jandex.idx, META-INF/beans.xml)
* Only include matching dependencies in the index

At least that would be consistent, and (imo) the expected behaviour.

@phillip-kruger WDYT?

phillip-kruger: Sounds good to me! Thanks

phillip-kruger (approved):

PR #990, opened 2021-12-08
Split and trim media types from @Consumes and @Produces

Azquelt:
Fixes #989

PR #1018, opened 2022-01-03
Trim media types separate from string split

MikeEdgar:
PR #990 includes functionality to split multiple media types from a single string in the produces/consumes `String` arrays. The use of the greedy regex pattern was flagged as a security hotspot (https://sonarcloud.io/organizations/smallrye/rules?open=java%3AS5852&rule_key=java%3AS5852). While the risk is low, the fix is very simple without much impact.

@Azquelt - FYI


Azquelt:
Looks fine, I hadn't noticed that the regex is problematic when used to split a string as it isn't achored at the start.

In the original code I avoided streams to try to do as little as possible in the case where there's only one media type, but looking at it again that probably wasn't necessary.

Azquelt (approved):

phillip-kruger (approved):
LGTM. Thanks @MikeEdgar !

PR #1054, opened 2022-02-15
Handle `@APIResponseSchema` without generated content

### LINKED ISSUE #1026: NPE when @Produces is missing
### Code to follow for reproducing the issue

1. `ResponseReader.addApiReponseSchemaFromAnnotation`:
```
APIResponse response = ResponseReader.readResponseSchema(context, annotation);

if (response.getDescription() == null) {
    response.setDescription(getReasonPhrase(status));
}
```
2. `ResponseReader.readResponseSchema`:

```
if (annotation == null || CurrentScannerInfo.getCurrentProduces() == null) {
    // Only generate the APIResponse if the endpoint declares an @Produces media type
    return null;
}
```

When `@Produces` is missing on the scanned method, it will raise the most known java exception: `NPE` 

Also, we are using @APIResponseSchema to specify the schema for a certain method, because we work with `javax.ws.rs.core.Response`. Maybe an enhancement is to prioritize it before returning null in method no 2 from above.

### Bonus maybe irelevent bug
Rename addApi**Reponse**SchemaFromAnnotation -> addApi**Response**SchemaFromAnnotation

### Env
* `quarkus` - 2.6.1
* `quarkus-smallrye-openapi1` - 2.6.1

MikeEdgar:
Fixes #1026

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/response/ResponseReader.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/response/ResponseReader.java
@@ -7,13 +7,14 @@ import java.util.Map;
 
 import org.eclipse.microprofile.openapi.models.Components;
 import org.eclipse.microprofile.openapi.models.media.Content;
-import org.eclipse.microprofile.openapi.models.media.MediaType;
+import org.eclipse.microprofile.openapi.models.media.Schema;
 import org.eclipse.microprofile.openapi.models.responses.APIResponse;
 import org.eclipse.microprofile.openapi.models.responses.APIResponses;
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationTarget;
 import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.Type;
 
 import com.fasterxml.jackson.databind.JsonNode;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/smallrye/openapi/runtime/io/response/ResponseReader.java
```diff
@@ -179,25 +181,31 @@ public static APIResponse readResponse(final AnnotationScannerContext context,
...
 
         APIResponseImpl response = new APIResponseImpl();
         response.setDescription(JandexUtil.value(annotation, ResponseConstant.PROP_RESPONSE_DESCRIPTION));
-        response.setContent(content);
         response.setResponseCode(JandexUtil.value(annotation, ResponseConstant.PROP_RESPONSE_CODE));
+
+        Type responseType = JandexUtil.value(annotation, ResponseConstant.PROP_VALUE);
+
+        if (CurrentScannerInfo.getCurrentProduces() != null && !TypeUtil.isVoid(responseType)) {
+            // Only generate the content if the endpoint declares an @Produces media type
```
MikeEdgar, line 196: Checking `void` doesn't really enable anything that couldn't be done with `@APIResponse`, but reduces surprises if `void.class` is given as the `value` for `@APIResponseSchema` and there *is* `@Produces` annotation.

phillip-kruger (approved):
LGTM

Issue #1042, opened 2022-01-31
microprofile-open-api 3.0


MikeEdgar:
@radcortez, I think you've started something for this in the `jakarta` branch, right?

radcortez:
Yes. Everything should be done there.

The strategy is not to merge new changes there. We just redo the migration again, until we move this to `main`. I have a script for this. I'll push it to the repo.

radcortez:
Ok, this should handle the migration:
https://github.com/smallrye/smallrye-open-api/blob/main/to-jakarta.sh

What I'm doing is resetting the `jakarta` branch with `main` and applying the migration again. I'm going to do an RC release of the current one.

MikeEdgar:
@radcortez any concerns if I periodically reset and run the migration on `jakarta`? I'm using that branch as a base to verify MP 3.1 TCK changes (the _next_ release) so I'd like to keep it as current with `main` as possible.

PR #1093, opened 2022-03-15
Implemented equals and hashCode in TagImpl

hbelmiro:
It would be nice to have `equals` and `hashCode` implemented in `TagImpl`, so we can compare instances of `org.eclipse.microprofile.openapi.models.tags.Tag`.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/models/tags/TagImpl.java
+++ b/core/src/main/java/io/smallrye/openapi/api/models/tags/TagImpl.java
@@ -74,12 +74,14 @@ public class TagImpl extends ExtensibleImpl<Tag> implements Tag, ModelImpl {
             return false;
         }
         TagImpl tag = (TagImpl) o;
-        return Objects.equals(name, tag.name) && Objects.equals(description, tag.description)
-                && Objects.equals(externalDocs, tag.externalDocs);
+        return Objects.equals(name, tag.name)
+                && Objects.equals(description, tag.description)
+                && Objects.equals(externalDocs, tag.externalDocs)
+                && Objects.equals(getExtensions(), tag.getExtensions());
     }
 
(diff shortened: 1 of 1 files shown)


hbelmiro:
@MikeEdgar it looks like I unconsciously found a bug.

```
Error:  Failures: 
Error:    ModelConstructionTest>Arquillian.run:138->openAPITest:422->checkListImmutable:1672 The list is expected to not contain the value: io.smallrye.openapi.api.models.tags.TagImpl@745f expected [false] but found [true]
```

MikeEdgar:
@hbelmiro this may need to wait for MP OpenAPI 3.0 support to be in `main`. See https://github.com/eclipse/microprofile-open-api/pull/488. 3.0 is currently in the `jakarta` branch, but that initial release should be just `main` with the javax -> jakarta migration (also see #1042 for background). Do you mind leaving this open for a while until that's done?

hbelmiro:
No problem @MikeEdgar. I'll leave this open.
Thank you.

MikeEdgar:
@hbelmiro the build it now passing with the MP OpenAPI 3.1 snapshot TCK. Suggestion on the implementation - it might be good to include `getExtensions()` in `equals`/`hashCode`. What do you think?

hbelmiro:
> @hbelmiro the build it now passing with the MP OpenAPI 3.1 snapshot TCK. Suggestion on the implementation - it might be good to include `getExtensions()` in `equals`/`hashCode`. What do you think?

Thanks for letting me know @MikeEdgar.

Yes. Makes sense to me. Also, I think we should include `TagImpl#externalDocs`. What do you think?
I pushed the changes including `getExtensions` and `externalDocs`. If you think we shouldn't include `externalDocs`, just let me know and I remove it.

PR #1096, opened 2022-03-17
Ignore synthetic JAX-RS resource classes by default

### LINKED ISSUE #1095: Synthetic classes not ignored
See https://github.com/quarkusio/quarkus/issues/24344

MikeEdgar:
Fixes #1095


phillip-kruger:
I think the problem is for some synthetic classes we do not want this. Like panache-rest, where the server side is generated ? Not sure if that applies here, can you double check ?

MikeEdgar:
> Like panache-rest, where the server side is generated ?

Ah, maybe that's where checking for `@RegisterRestClient` comes back into the picture.

phillip-kruger:
ok are we already doing that ? Maybe add a test for that scenario ? (class without @RegisterRestClient)

MikeEdgar:
We are not, due to some previous issues. I'm checking what a Panache REST resource looks like as well.

phillip-kruger:
Ok cool

MikeEdgar:
@phillip-kruger updated, please have another look. We need a way to put the source code for the test index in the project. Maybe we should add another module that is a test dependency of the others that builds a Quarkus app?

phillip-kruger (approved):
Thanks!

PR #1097, opened 2022-03-17
Support for SecurityRequirementsSet

Azquelt:
Support SecurityRequriementsSet, passing the TCKs added to
https://github.com/eclipse/microprofile-open-api/pull/519

Objects which support security requirements now support having both
individual security requirements and security requirements sets.

An individual security requirement is equivalent to a security
requirement set with a single element. Therefore, when reading
annotations, we always process both lists together.

I've opened this against the `jakarta` branch, since it requires changes which will be in MP OpenAPI 3.1 and left it as draft for the moment since there isn't a release available yet which includes these changes.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/securityrequirement/SecurityRequirementConstant.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/securityrequirement/SecurityRequirementConstant.java
@@ -2,6 +2,8 @@ package io.smallrye.openapi.runtime.io.securityrequirement;
 
 import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirement;
 import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirements;
+import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirementsSet;
+import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirementsSets;
 import org.jboss.jandex.DotName;
 
 /**
@@ -17,6 +19,8 @@ public class SecurityRequirementConstant {
 
     static final DotName DOTNAME_SECURITY_REQUIREMENT = DotName.createSimple(SecurityRequirement.class.getName());
     static final DotName DOTNAME_SECURITY_REQUIREMENTS = DotName.createSimple(SecurityRequirements.class.getName());
+    static final DotName DOTNAME_SECURITY_REQUIREMENTS_SET = DotName.createSimple(SecurityRequirementsSet.class.getName());
+    static final DotName DOTNAME_SECURITY_REQUIREMENTS_SETS = DotName.createSimple(SecurityRequirementsSets.class.getName());
 
     public static final String PROP_NAME = "name";
     public static final String PROP_SCOPES = "scopes";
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/smallrye/openapi/runtime/io/securityrequirement/SecurityRequirementReader.java
```diff
@@ -82,19 +99,43 @@ private SecurityRequirementReader() {
      * @return SecurityRequirement model
      */
     public static SecurityRequirement readSecurityRequirement(AnnotationInstance annotationInstance) {
+        SecurityRequirement requirement = new SecurityRequirementImpl();
+        addSecurityRequirement(requirement, annotationInstance);
+        if (requirement.getSchemes().isEmpty()) {
```
MikeEdgar, line 104:
```suggestion
        // True when the annotation is missing the `name` property.
        if (requirement.getSchemes().isEmpty()) {
```

MikeEdgar (commented):
This looks good, thank you. Since there are effectively 3 work streams (prod/main, MP OpenAPI 3.0 with the Jakarta changes, and MP OpenAPI 3.1 currently without a branch) let's leave it open for now.

core/src/main/java/io/smallrye/openapi/runtime/io/securityrequirement/SecurityRequirementConstant.java
```diff
@@ -17,6 +19,8 @@
 
     static final DotName DOTNAME_SECURITY_REQUIREMENT = DotName.createSimple(SecurityRequirement.class.getName());
     static final DotName DOTNAME_SECURITY_REQUIREMENTS = DotName.createSimple(SecurityRequirements.class.getName());
+    static final DotName DOTNAME_SECURITY_REQUIREMENTS_SET = DotName.createSimple(SecurityRequirementsSet.class.getName());
+    static final DotName DOTNAME_SECURITY_REQUIREMENTS_SETS = DotName.createSimple(SecurityRequirementsSets.class.getName());
```
MikeEdgar, line 23: If the string constant values are used here instead of `Class#getName`, I'm curious if this will pass the 3.0 or even 2.0 MP OpenAPI TCK. If it does, we can merge earlier since Jandex buffers us a bit from API changes.

MikeEdgar:
@Azquelt, this can be re-based on `main` now that it has been updated for the 3.1-SNAPSHOT of MP OpenAPI.

core/src/main/java/io/smallrye/openapi/runtime/io/securityrequirement/SecurityRequirementReader.java
Azquelt: Hmm, this is correct, but looking at it again, a `SecurityRequirement` annotation should never be missing the `name` property.

I didn't like that there was a code path where `addSecurityRequirement()` could actually not add the requirement and I definitely didn't want to end up adding an empty `SecurityRequirement` to the model in that case.

Can you shed any light on why we're checking for `null` when retrieving the name? I agree that we can add a comment here, but it should be clear that this is a path we expect to never be taken.

MikeEdgar: I agree it seems like an impossibility that `name` would be missing (since it has no default value). The `null` check does not seem necessary in `addSecurityRequirement`.

MikeEdgar:
Included in #1137

PR #1125, opened 2022-04-16
Update `to-jakarta.sh` to exclude `javax` test cases

MikeEdgar:
This moves the custom `javax` test cases to their own subpackages (similar to the `jakarta` equivalents) and updates the Jakarta migration script to ignore these source files.


MikeEdgar:
Once this PR is merged, the `2.1.x` and `3.0.x` branches will be created and `main` will start tracking the Microprofile OpenAPI 3.1 changes.

This PR looks huge, but the only changes are moving the test classes dealing with `javax` annotations to sub-packages (and updating references) in the first commit and updating the jakarta migration script in the second commit.

@phillip-kruger , @radcortez Please let me know if you have any questions, etc.

phillip-kruger (approved):

PR #1144, opened 2022-05-20
Map BV `@Pattern` constraint to schema `pattern`

### LINKED ISSUE #1108: `@Pattern` annotation should reflect in data structure
If specifying a regular expression constraint via `@javax.validation.constraints.Pattern`, it should automatically appear in the output, similar to the treatment of `@NotNull` in https://github.com/smallrye/smallrye-open-api/issues/131 

As context, I am using quarkus 2.7.5 with openapi/swagger-ui with Kotlin, and currently, I need to explicitly add a duplicate `pattern` to `@Schema` to make it reflect in the generated schema correctly.

MikeEdgar:
Fixes #1108


phillip-kruger (approved):
Thanks @MikeEdgar

PR #1145, opened 2022-05-24
Bean validation spec

Azquelt:
Updated the bean validation annotation scanning to match the spec update in https://github.com/eclipse/microprofile-open-api/pull/531

The changes needed were:
* respect the new config property to disable reading bean validation annotation scanning
* process constraints which apply to the default group (rather than those which _only_ apply to the default group)
* use an exclusive range for `@Positive` and `@Negative`


core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
```diff
@@ -674,10 +677,12 @@ protected void setSchemaProperties(Schema schema,
...
-            BeanValidationScanner.applyConstraints(paramTarget,
-                    paramSchema,
-                    paramName,
-                    (target, name) -> setRequired(name, schema));
+            if (beanValidationScanner.isPresent()) {
+                beanValidationScanner.get().applyConstraints(paramTarget,
+                        paramSchema,
+                        paramName,
+                        (target, name) -> setRequired(name, schema));
+            }
```
MikeEdgar, line 685: For consistence with the one above.
```suggestion
            beanValidationScanner.ifPresent(s -> s.applyConstraints(paramTarget,
                        paramSchema,
                        paramName,
                        (target, name) -> setRequired(name, schema)));
```

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
```diff
@@ -340,8 +340,9 @@ void negative(AnnotationTarget target, Schema schema) {
         if (constraint != null && schema.getMaximum() == null) {
             Boolean exclusive = schema.getExclusiveMaximum();
 
-            if (exclusive != null && exclusive) {
+            if (exclusive == null || exclusive) {
                 schema.setMaximum(BigDecimal.ZERO);
+                schema.setExclusiveMaximum(Boolean.TRUE);
```
MikeEdgar, line 345: I think a null/undefined value should be interpreted as false.

https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00#section-5.3

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
```diff
@@ -169,9 +169,11 @@ Schema processField() {
...
 
-        for (AnnotationTarget contraintTarget : typeResolver.getConstraintTargets()) {
-            // Apply constraints from all bean properties associated with the schema property
-            BeanValidationScanner.applyConstraints(contraintTarget, fieldSchema, propertyKey, this);
+        if (context.getBeanValidationScanner().isPresent()) {
+            for (AnnotationTarget contraintTarget : typeResolver.getConstraintTargets()) {
+                // Apply constraints from all bean properties associated with the schema property
+                context.getBeanValidationScanner().get().applyConstraints(contraintTarget, fieldSchema, propertyKey, this);
+            }
         }
```
MikeEdgar, line 177: For consistency with the other calls for BV scanning.
```suggestion
        context.getBeanValidationScanner().ifPresent(s -> {
            for (AnnotationTarget contraintTarget : typeResolver.getConstraintTargets()) {
                // Apply constraints from all bean properties associated with the schema property
                s.applyConstraints(contraintTarget, fieldSchema, propertyKey, this);
            }
        });
```

MikeEdgar (commented):
Looks good overall. Just a few comments.

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
MikeEdgar: Disregard - I did not consider non-integer cases for positive/negative, so your change makes sense to me now.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
Azquelt: This can't be done because `paramSchema` isn't effectively final.

I agree though that consistency would be good so I'll do the change the other way, replacing `ifPresent` with the `if` block.

Azquelt:
I've made updates in response to the review comments.

MikeEdgar (approved):

PR #1148, opened 2022-05-27
Update scanning config

Azquelt:
Update scanning config to match https://github.com/eclipse/microprofile-open-api/pull/533

The existing scanning code had a problem when a single package was matched by more than value of the scanning config.

For example:
```mp.openapi.scan.packages=com.example.a,com.example.a.b.c```
Here, the package `com.example.a.b.c.d` matches both entries. To correctly follow the logic now defined by the spec, we need to ensure we get the longest match, whereas the previous code may return either match.

Changes:
* Split the config property value on `,` and match against each part separately
* Treat each part as a regex if it starts with `^` or ends with `$`
  * Previous code would treat the whole property as a regex
* If the config property is a regex, just try to match it against the whole string
  * Previous code would require that a package match was a prefix, even if the regex didn't start with `^`
  * Previous code would require that a class match ended with the entire simple class name, even if the regex didn't end with `$`

Enhancements beyond the spec requirements:
* Allow a string in `mp.openapi.scan.classes` or `mp.openapi.scan.classes.exclude` to match the simple class name, or the simple class name with a package suffix, in addition to matching the fully qualified class name
* Treat a config property part as a regex if it starts with `^` or ends with `$`

One other difference, potentially a bug:
* The spec says that `mp.openapi.scan.packages` and `mp.openapi.scan.packages.exclude` should match the package or any parent package. We are doing a string prefix match against the package name, which is not quite the same.
  * We would say that `com.example.f` matches `com.example.foo`, even though `com.example.f` is not a parent package of `com.example.foo`

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
+++ b/core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
@@ -33,22 +33,20 @@ public interface OpenApiConfig {
         return false;
     }
 
-    default Pattern scanPackages() {
+    default Set<String> scanPackages() {
         return null;
     }
 
-    default Pattern scanClasses() {
+    default Set<String> scanClasses() {
         return null;
     }
 
-    default Pattern scanExcludePackages() {
-        return Pattern.compile(
-                "(" + OpenApiConstants.NEVER_SCAN_PACKAGES.stream().map(Pattern::quote).collect(Collectors.joining("|")) + ")");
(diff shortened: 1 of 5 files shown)


MikeEdgar (approved):
This looks good to me. I'll wait to merge until there is a snapshot build with https://github.com/eclipse/microprofile-open-api/pull/533 that we can re-run the CI against. Thanks @Azquelt

MikeEdgar (approved):

Issue #1166, opened 2022-06-28
Support different encoding in smallrye-open-api-maven-plugin


ColdanR:
Hello,
I'm currently setting up an automated documentation for OpenAPI, but on Windows 10 the plugin exports the schema files with a different encoding than UTF8. (System Default)

Looking up the code the generated documentation will be transformed to an byte array without usage of encoding.
https://github.com/smallrye/smallrye-open-api/blob/0af0bec05a1fc0c477254c161e54103ac32460f4/tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/GenerateSchemaMojo.java#L431-L433

So please allow to configure encoding or use utf-8 encoding as default.

```
Charset charset = StandardCharsets.UTF_8;
if (charsetParameter != null) {
try {
    charset = Charset.forName(charsetParameter);
} catch (IllegalArgumentException e) {
    throw new MojoExecutionException("Cannot find charset for " + charsetParameter);
}
}

writeSchemaFile(directory, "yaml", yaml.getBytes(charset));
writeSchemaFile(directory, "json", json.getBytes(charset));
```

Thanks,
Bernd Schmidt

phillip-kruger:
Hi Bernd, thanks for the issue. Are you not keen to do a PR to fix this ?

ColdanR:
Hi Phillip,
currently I don't have the time to do so. I also didn't find any information about contributing. Would be nice, if you can add a code of conduct and contribution.md file für information.
I will try to find some time next week.

PR #1179, opened 2022-07-12
Solves #1166

ColdanR:
Added encoding as parameter to set output file encoding

### DIFF
--- a/tools/maven-plugin/README.adoc
+++ b/tools/maven-plugin/README.adoc
@@ -49,6 +49,7 @@ E.g. for includeDependenciesScopes could be configured as:
 - `configProperties` (String) - Load any properties from a file. Example `${basedir}/src/main/resources/application.properties`.
 - `attachArtifacts` (boolean, default: false) - Attach the built OpenAPI schema as build artifact.
 - `skip` (boolean, default: false) - Skip execution of the plugin.
+- `encoding` (String) - Encoding of output OpenAPI files.
 
 == MicroProfile OpenAPI Properties
 
--- a/tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/GenerateSchemaMojo.java
+++ b/tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/GenerateSchemaMojo.java
@@ -6,6 +6,9 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.nio.charset.Charset;
+import java.nio.charset.IllegalCharsetNameException;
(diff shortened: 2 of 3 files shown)


tools/maven-plugin/README.adoc
```diff
@@ -49,6 +49,7 @@ E.g. for includeDependenciesScopes could be configured as:
 - `configProperties` (String) - Load any properties from a file. Example `${basedir}/src/main/resources/application.properties`.
 - `attachArtifacts` (boolean, default: false) - Attach the built OpenAPI schema as build artifact.
 - `skip` (boolean, default: false) - Skip execution of the plugin.
+- `encoding` (String) - Encoding of output OpenAPI files.
```
phillip-kruger, line 52: We can probably make the default UTF-8 to stay backward compatible ?

tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/GenerateSchemaMojo.java
```diff
@@ -223,6 +227,12 @@ public class GenerateSchemaMojo extends AbstractMojo {
     @Parameter(property = "scanExcludeProfiles")
     private List<String> scanExcludeProfiles;
 
+    /**
+     * Output encoding for openapi document.
+     */
+    @Parameter(property = "encoding")
```
phillip-kruger, line 233: So here defaultValue = "UTF-8"

phillip-kruger (changes requested):

tools/maven-plugin/README.adoc
MikeEdgar: @phillip-kruger, I thought the same thing, but previously it was using `String#getBytes`, which internally calls `Charset.defaultCharset`.

phillip-kruger: Ok, but it should still be able to do this. Not sure if you can make these properties Optional ? If so , use Optional with empty setting to default

MikeEdgar: The default `defaultValue` in the `@Parameter` is an empty string. That flows into the `(!StringUtils.isBlank(encoding))` check and then uses the platform default encoding.

phillip-kruger: Ok great !

phillip-kruger: So then we can merge ?

MikeEdgar (approved):

tools/maven-plugin/README.adoc
MikeEdgar: Yeah, lgtm!

phillip-kruger (approved):
Thanks !

PR #1181, opened 2022-07-14
Do not set `nullable: false` due to Bean Validation annotations

### LINKED ISSUE #1180: `@NotNull` should not generate `nullable: false`
I just stumbled upon a similar issue, where 
```
@NotNull
public Access access
```
results in 
```
access:
  allOf:
          - $ref: '#/components/schemas/Access'
          - nullable: false
```
Naturally I don't want to remove the `@NotNull` as it is used for other validations. And for some reason when I process the generated document with `openapi-generator` to create Typescript API this results in
```
access: Access | null
```

_Originally posted by @rvansa in https://github.com/smallrye/smallrye-open-api/issues/1023#issuecomment-1144634469_

MikeEdgar:
Fixes #1180 

Nullable of `false` is the default assumed value in OpenAPI 3

### DIFF
--- a/extension-jaxrs/src/test/resources/io/smallrye/openapi/runtime/scanner/params.optional-types.json
+++ b/extension-jaxrs/src/test/resources/io/smallrye/openapi/runtime/scanner/params.optional-types.json
@@ -246,8 +246,7 @@
             "in": "query",
             "required": true,
             "schema": {
-              "type": "string",
-              "nullable": false
+              "type": "string"
             }
           }
         ],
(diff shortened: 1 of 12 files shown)


extension-jaxrs/src/test/resources/io/smallrye/openapi/runtime/scanner/params.optional-types.json
```diff
@@ -274,7 +273,7 @@
         "properties": {
           "title": {
             "type": "string",
-            "nullable": false
+            "nullable": true
```
MikeEdgar, line 276: Field type is `Optional`, resulting in the `true`

MikeEdgar:
@phillip-kruger let me know if this looks OK.

phillip-kruger (approved):
LGTM

PR #1198, opened 2022-07-31
Include annotated property methods without is/get/set JavaBean prefix

### LINKED ISSUE #1083: Add support for methods not using the javabean style
As originally reported to the Quarkus team: https://github.com/quarkusio/quarkus/issues/24085

It seems Jackson's `@JsonProperty` annotations are only picked up on methods if they have a `get` or `is` prefix. It would be great if annotated methods are picked up regardless of any prefix.

MikeEdgar:
Fixes #1083


phillip-kruger (approved):
LGTM

phillip-kruger:
Thanks @MikeEdgar !

PR #1208, opened 2022-08-07
Add support for removing unused models

### LINKED ISSUE #1183: Support for removing unused models
It would be convenient to have a way to automatically remove unused models from generated OpenAPI documents.

Currently, all classes marked with `@Schema` will automatically be included in the OpenAPI document. This can lead to unused classes being included anyways if they exist in the index. This can be worked around by explicitly specifying the packages or classes to be scanned but it can sometimes feel a little cumbersome with an evolving or dynamic API.

Use cases:
1. A central library contains many standard models shared across many APIs for conformity. The project consuming may not have direct control over these annotations. What if the central library adds an unused class to an existing package?
2. A Quarkus app is configured to include/exclude different endpoints depending on the active profile. Configuring the scanning on a per-profile basis results in a lot of repetition and reconfiguring profile endpoints becomes a complex task.

MikeEdgar:
Fixes #1183


phillip-kruger (approved):
LGTM ! Thanks @MikeEdgar

PR #1209, opened 2022-08-07
Add support for `@JsonView`

### LINKED ISSUE #1008: Add support for @JsonView annotations
Hello,
I've seen that Swagger already support the usage of anootation `@JsonView` in order to filter input and output of a model class in RequestBody or ResponseBody (see https://github.com/swagger-api/swagger-core/issues/479 fixed with https://github.com/swagger-api/swagger-core/pull/2681 since 2018), but If I try using this annotation in my controller in order to selectively hide some fields of my model class from input or output in the swagger-ui, they are anyway displayed.
This is my example

```java
    @JsonView(RestViews.ViewAdmin.class)
    public CompletionStage<FilterRest> saveFilter(
            @Valid @JsonView(RestViews.Save.class) @ConvertGroup(to = ValidationGroups.Save.class) FilterRest filterRest) {
    }
```
My model class is:
```java
public class FilterRest {

    @JsonView({RestViews.Update.class, RestViews.ViewUser.class})
    @Null(groups = {ValidationGroups.Create.class})
    private String id;

    @JsonView({RestViews.Create.class, RestViews.ViewAdmin.class})
    @NotNull
    private Set<String> countryValidity;
```
Event the field `id` is marked with a different JsonView actually is showed in swagger-ui for the request in the API.

What am I doing wrong?

MikeEdgar:
Closes #1008

When endpoints responses and request bodies are found with `@JsonView`, the schema tree generated from the scan will include the name(s) of the views that were applied to the request/response. For cases where a nested bean should not be different between the views, the user may annotate a field with `@Schema(ref = "<NameOfCommonSchema>")`.

Users will likely want to set `mp.openapi.extensions.smallrye.remove-unused-schemas.enable` to `true`.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/constants/JacksonConstants.java
+++ b/core/src/main/java/io/smallrye/openapi/api/constants/JacksonConstants.java
@@ -22,10 +22,12 @@ public class JacksonConstants {
             .createSimple("com.fasterxml.jackson.annotation.JsonPropertyOrder");
     public static final DotName JSON_UNWRAPPED = DotName
             .createSimple("com.fasterxml.jackson.annotation.JsonUnwrapped");
-    public static final DotName JSON_NAMING = DotName
-            .createSimple("com.fasterxml.jackson.databind.annotation.JsonNaming");
     public static final DotName JSON_VALUE = DotName
             .createSimple("com.fasterxml.jackson.annotation.JsonValue");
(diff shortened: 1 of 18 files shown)


phillip-kruger (approved):

phillip-kruger:
Thanks @MikeEdgar !

PR #1215, opened 2022-08-12
Ignore class members having illegal Java identifiers in name

### LINKED ISSUE #1195: Kotlin value classes not represented correctly.
When adding a kotlind value class to a response, the property name is added twice. I tried different approaches with specifying schema, but wasn't able to correct or hide the additional property.

value class:

```kotlin
@JvmInline
value class IdWithOverwrittenSchema(val value: Long)
```

Response:

```kotlin
data class TestResponse(
    @field:Schema(type = SchemaType.INTEGER, implementation = Long::class, name = "idWithOverwrittenSchema")
)
```

Resource:

```kotlin
@Path("/test")
class TestResource {
    @GET
    fun testMethod(): TestResponse {
        return TestResponse()
    }
}
```

I tried different variations and put some of them in a sample project: https://github.com/dirkbolte/openapi-test

The result looks like this:
value_class_bug

value classes as method parameters (e.g. on `@GET` with `@PathParam` do not have this issue and are parsed correctly.

Versions:
- smallrye-open-api-core: 2.1.22
- microprofile-openapi-api: 2.0.1
- Quarkus (incl. resteasy): 2.10.3.Final
- Java: 11
- Kotlin: 1.6.21

MikeEdgar:
- Do not include properties where the name is not a valid Java identifier (covers the case of generated methods for Kotlin value class fields)
- Add test data module, initally holding test Kotlin classes

Fixes #1195

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -19,15 +19,16 @@ mvn clean install
 
 === Project structure
 
-* link:core[] - The core OpenAPI code, independent of entry point dependencies.
-* link:extension-jaxrs[] - The JAX-RS entry point. This module depends on JAX-RS and core.
-* link:extension-spring[] - The Spring entry point. This module depends on Spring and core.
-* link:extension-vertx[] - The Vert.x entry point. This module depends on Vert.x and core.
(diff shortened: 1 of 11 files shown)


phillip-kruger (approved):
Great ! Thanks @MikeEdgar !

PR #1221, opened 2022-08-22
Enable Sonar for PRs via triggered job, skip JavaDoc for Kotlin code

MikeEdgar:
This allows the Sonar scan with PRs from forks by scanning the Jacoco build results in a separate job, trigger by the primary build job. The `sonar.yml` job will not run until it is merged.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,8 +2,11 @@ name: SmallRye Build
 
 on:
   push:
-    branches-ignore:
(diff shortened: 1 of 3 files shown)


phillip-kruger (approved):

MikeEdgar:
The `target` directory for `open-api-ui-forms` is a bit large (includes `node_modules`). I'll come back to clean that up once the Sonar part is verified to be fine.

PR #1224, opened 2022-08-24
TCK runner: Update the archive using BeforeDeploy

Azquelt:
Previously an ApplicationArchiveProcessor was used to run the
implementation and generate the OpenAPI document in the archive before
deploying it. However, an ApplicationArchiveProcessor is only called
for _testable_ archives (archives where tests run on the server),
whereas to pass the TCK we need to process all the archives.

Instead, use the BeforeDeploy event to update the archive which is
called for every deployment.

### DIFF
--- a/testsuite/tck/src/test/java/io/smallrye/openapi/tck/DeploymentProcessor.java
+++ b/testsuite/tck/src/test/java/io/smallrye/openapi/tck/DeploymentProcessor.java
@@ -22,8 +22,9 @@ import java.util.stream.Stream;
 
 import org.eclipse.microprofile.config.ConfigProvider;
 import org.eclipse.microprofile.openapi.models.OpenAPI;
-import org.jboss.arquillian.container.test.spi.client.deployment.ApplicationArchiveProcessor;
-import org.jboss.arquillian.test.spi.TestClass;
+import org.jboss.arquillian.container.spi.client.deployment.DeploymentDescription;
+import org.jboss.arquillian.container.spi.event.container.BeforeDeploy;
+import org.jboss.arquillian.core.api.annotation.Observes;
 import org.jboss.jandex.DotName;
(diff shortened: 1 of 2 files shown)


testsuite/tck/src/test/java/io/smallrye/openapi/tck/DeploymentProcessor.java
```diff
@@ -47,12 +48,21 @@
...
     private static Logger LOGGER = Logger.getLogger(DeploymentProcessor.class.getName());
     public static volatile ClassLoader classLoader;
 
-    @Override
-    public void process(Archive<?> archive, TestClass testClass) {
+    public void observeDeployment(@Observes final BeforeDeploy beforeDeploy) {
+        DeploymentDescription deployment = beforeDeploy.getDeployment();
+        Archive<?> testableArchive = deployment.getTestableArchive();
+        if (testableArchive != null) {
+            process(testableArchive);
```
MikeEdgar, line 59: This can presumably be removed with RC3, correct?

Azquelt: No, there are still a few testable archives which test the model creation APIs.

Azquelt: For reference it's [OASFactoryErrorTest](https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/tck/OASFactoryErrorTest.java) and [ModelConstructionTest](https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java)

MikeEdgar (approved):
LGTM. Thanks @Azquelt

PR #1229, opened 2022-08-26
Upgrade to Jandex 3.0.0

Ladicek:
Submitting as draft because this needs discussion on release coordination (I guess).

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -44,7 +44,7 @@
             <artifactId>jackson-dataformat-yaml</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.jboss</groupId>
+            <groupId>io.smallrye</groupId>
             <artifactId>jandex</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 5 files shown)


MikeEdgar:
Jandex is managed by the smallrye-parent pom currently. Will that be continued going forward?

Ladicek:
Frankly, I just don't like how the SmallRye parent manages (or not) some versions. To me, it feels like a big dump of random stuff that is only updated when someone remembers and there's no coordination around it. In SmallRye Fault Tolerance, I don't even use it (I use `smallrye-build-parent`).

This is a relatively big/important change that requires coordination -- adding the parent POM to the mix would just make it worse. I propose we replace Jandex 2.x with Jandex 3.0 in the parent POM when things settle.

Ladicek:
I also just noticed that SmallRye OpenAPI uses the `Type` hierarchy on a few places (`TypeProcessor`, `TypeResolver`). This code will likely have to be updated to deal with type variable references. In general, code that handles `UnresolvedTypeVariable` or `Type.Kind.UNRESOLVED_TYPE_VARIABLE` likely needs to be updated. I don't really understand how that code works, so I'll refrain from attempting to change it myself. Just letting you know.

MikeEdgar (approved):
lgtm - ignoring the code coverage check since they're just proxy methods and aren't being used (yet) anyway.

PR #1228, opened 2022-08-26
Upgrade to Jandex 3.0.0

Ladicek:
Submitting as draft because this needs discussion on release coordination (I guess).

PR #1227, opened 2022-08-26
Upgrade to Jandex 3.0.0

Ladicek:
Submitting as draft because this needs discussion on release coordination (I guess).

PR #1232, opened 2022-08-30
Upgrade to Jandex 3.0.0

Ladicek:
I noticed someone created the 2.3.x branch, but it was branched off of the 2.2.0 tag, which I believe is not right. So this PR, targeting the 2.3.x branch:

1. fast-forwards the 2.3.x branch one commit after the 2.2.0 tag in the 2.2.x branch (`[maven-release-plugin] prepare for next development iteration`);
2. fixes the `<tag>` element in the POM, as that should read `HEAD` outside of release commits;
3. changes the `<version>` elements to say `2.3.0-SNAPSHOT`;
4. upgrades to Jandex 3.0.0.

With this PR merged to current 2.3.x (which points to a6d1ad538e), I believe the 2.3.x branch would be in the correct state. Next, we would close #1227 (as we want to keep 2.2.x on Jandex 2.x). We could also close #1228, if 3.0.x is really automatically created from 2.2.x -- I believe we would actually also move 3.0.x to be based on 2.3.x.

Makes sense, or do I completely misunderstand how SmallRye OpenAPI is managed? Thank you!

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -44,7 +44,7 @@
             <artifactId>jackson-dataformat-yaml</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.jboss</groupId>
+            <groupId>io.smallrye</groupId>
             <artifactId>jandex</artifactId>
         </dependency>
         <dependency>
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/FilteredIndexView.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/FilteredIndexView.java
@@ -1,6 +1,7 @@
 package io.smallrye.openapi.runtime.scanner;
 
 import java.util.Collection;
+import java.util.Set;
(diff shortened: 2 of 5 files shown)


Ladicek:
Also, if you wanna take the first 3 commits from this PR and push them to 2.3.x directly, that would be fine with me :-) Or any other way you wish, of course.

MikeEdgar:
@Ladicek , this is good, thank you. Please go ahead and close the 3.0.x and 2.2.x PRs and I'll update the main PR and this with the changes for type references. A 3.0-RC is manually generated each time a release of 2.x occurs, with only the Jakarta changes applied with the `to-jakarta.sh` script. Once 3.0 is released that will stop, of course.

Ladicek:
PRs to 2.2.x and 3.0x closed, thank you!

MikeEdgar (approved):

PR #1235, opened 2022-09-02
Smallrye Open API Gradle plugin

### LINKED ISSUE #1231: Gradle plugin contribution
We have ported the existing Maven plugin to Gradle and are using it for a while now for our build of [projectnessie](https://github.com/projectnessie/nessie). The Gradle port is currently maintained [here](https://github.com/projectnessie/gradle-build-plugins/tree/main/smallrye-openapi), but not published as a Gradle plugin.
Is there interest in having a Gradle plugin in this repository?

If yes, then we would create a pull request, to
* introduce the Gradle plugin incl tests
* unify code that can be shared with the Maven plugin

snazy:
Implements the Gradle plugin. It has pretty much the same
functionality as the Maven plugin, with a minor exceptions:
The default charset in the Maven plugin is the "current" charset,
in the Gradle plugin it's "convention" is UTF-8.

Uses a similar machanism to integrate the Gradle plugin build and
tests as in Quarkus.

Gradle specifics: Implements the `SmallryeOpenApiPlugin`, which
registers a so called "extension" (`SmallryeOpenApiExtension`),
which is handy in Gradle to configure settings, and the task
(`SmallryeOpenApiTask`) , which performs the actual code generation.
The task itself is cacheable, "inputs" are the main java sources and
the actual configuration (i.e. the Open API schema files will not be
unnecessarily re-generated).

The plugin also registers a configuration called `openapi`, which
references the task's output, so the generated schema files can be
safely referenced from other Gradle projects.

Short snippet of the plugin usage:
```kotlin
plugins {
  id("io.smallrye.openapi") // assuming plugin versions are managed
}

smallryeOpenApi {
  infoTitle.set("My Open API stuff")
  infoContactName.set("Me, myself and I")
  // property names are similar to the Maven plugin property names
}
```

Tests for the Gradle plugin include some basic test that exercise
the configuration paths and "inheritance" of configurations from
the plugin extension to the generator task.

Also contains changes to the `SmallRye Release` GH workflow
(commented out). Crossing fingers, that it works out of the box,
requires the secrets to publish the plugins to the Gradle plugin
portal.

Tried to extract common code in the Maven + Gradle plugins, but it
became a bit too messie due to different loggers and some other
differences and dependencies.

The `gradlew*` files and stuff in the `gradle/` directory have been
generated by Gradle itself (untouched). 

Fixes #1231


phillip-kruger (approved):
LGTM, @MikeEdgar w.d.y.t ?

phillip-kruger:
@snazy can you please look at the CI failures

MikeEdgar (approved):

MikeEdgar:
lgtm - not sure why neither plugin shows up in Sonar, but that can be looked at separately. Also just noting that the plugin won't be published with keys when we do a release.

phillip-kruger (approved):

PR #1259, opened 2022-09-15
Prevent duplicate entries in operation and path item arrays

### LINKED ISSUE #1256: Duplicate path params for subresources that reference param multiple times
See https://github.com/quarkusio/quarkus/issues/27902

MikeEdgar:
Fixes #1256 

When a parameter is present in both the operator parameters array and the path item parameters array, prefer the operation parameter and exclude the path item parameter.


phillip-kruger (approved):
LGTM

PR #1269, opened 2022-09-22
Experimental support for multiple TCK versions

### LINKED ISSUE #1268: Investigate running multiple TCK versions
In order to reduce ongoing maintenance of multiple branches, it may be good to have a single version that passes multiple versions of the TCK - 2.0, 3.0, and 3.1 at the moment.

MikeEdgar:
Fixes #1268

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,3 +76,36 @@ jobs:
             !**/target/node/**/*
             !**/target/**/*.jar
             !**/target/**/*.war
+
+  tck-additional:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+        - tck-version: "2.0.1"
+          testng-version: "6.14.3"
+        - tck-version: "3.0"
+          testng-version: "6.14.3"
+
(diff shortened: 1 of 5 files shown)


MikeEdgar:
@radcortez , let me know if you think this is a viable approach to eliminating the need for maintaining multiple branches (pre-Jakarta, Jakarta, current MP 6). We'll continue to generate release 8 bytecode for as long as we need to support MP OpenAPI 2.0. This should be further refined to attach the one-page HTML TCK reports to releases.

/cc: @phillip-kruger

radcortez:
Looks good. Are you able to pass the TCK with the same code between `javax` and `jakarta`?

MikeEdgar:
> Looks good. Are you able to pass the TCK with the same code between `javax` and `jakarta`?

Yes, the scanner looks for both namespaces for all of the supported annotations. The only other thing I need to make sure of is that we're not using any methods new to Jandex 3 to ensure it works with the 2.4.x series if that's what is available.

radcortez:
Great. I'm fine if you go with this approach.

phillip-kruger (approved):
LGTM

PR #1289, opened 2022-10-17
Fail or warn on duplicate operationIds

renegrob:


extension-vertx/src/main/java/io/smallrye/openapi/vertx/VertxAnnotationScanner.java
```diff
@@ -120,7 +120,7 @@ private void processRoutes(final AnnotationScannerContext context, OpenAPI openA
         // Get all Vert.x routes and convert them to OpenAPI models (and merge them into a single one)
         Collection<AnnotationInstance> routeAnnotations = context.getIndex()
                 .getAnnotations(VertxConstants.ROUTE);
-        List<ClassInfo> applications = new ArrayList<>();
+        Set<ClassInfo> applications = new HashSet<>();
         for (AnnotationInstance annotationInstance : routeAnnotations) {
             if (annotationInstance.target().kind().equals(AnnotationTarget.Kind.METHOD)) {
                 ClassInfo classInfo = annotationInstance.target().asMethod().declaringClass();
```
renegrob, line 126: The same class is added for every method with a route annotation. Later on every class in the list was processed with `processRouteClass(...)`. I replaced the List with a Set to avoid processing the same class multiple times. I'm not sure, if we have to maintain the order of the classes as returned by the index but AFAIK there's no well defined order.

core/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
```diff
@@ -405,6 +406,17 @@ public OperationIdStrategy getOperationIdStrategy() {
...
 
+    @Override
+    public DuplicateOperationIdBehavior getDuplicateOperationIdBehavior() {
+        if (duplicateOperationIdBehavior == null) {
+            String behavior = getStringConfigValue(OpenApiConstants.DUPLICATE_OPERATION_ID_BEHAVIOR);
+            if (behavior != null) {
+                return DuplicateOperationIdBehavior.valueOf(behavior);
+            }
+        }
+        return null;
```
MikeEdgar, line 417: This approach will cache the value and also defer to the interface's default if not found in the configuration properties.

```suggestion
        if (duplicateOperationIdBehavior == null) {
            duplicateOperationIdBehavior = getConfig()
                    .getOptionalValue(OpenApiConstants.DUPLICATE_OPERATION_ID_BEHAVIOR, String.class)
                    .map(DuplicateOperationIdBehavior::valueOf)
                    .orElseGet(OpenApiConfig.super::getDuplicateOperationIdBehavior);
        }
        return duplicateOperationIdBehavior;
```

tools/gradle-plugin/src/main/java/io/smallrye/openapi/gradleplugin/Configs.java
```diff
@@ -240,6 +244,11 @@ public OperationIdStrategy getOperationIdStrategy() {
                 return operationIdStrategy.getOrNull();
             }
 
+            @Override
+            public DuplicateOperationIdBehavior getDuplicateOperationIdBehavior() {
+                return duplicateOperationIdBehavior.getOrNull();
```
MikeEdgar, line 249: Instead of null, can it default to the interface's default value? That way the logic checking this can always assume non-null.

tools/gradle-plugin/src/main/java/io/smallrye/openapi/gradleplugin/Configs.java
```diff
@@ -50,6 +51,7 @@ class Configs {
     final Property<String> infoLicenseName;
     final Property<String> infoLicenseUrl;
     final Property<OperationIdStrategy> operationIdStrategy;
+    final Property<DuplicateOperationIdBehavior> duplicateOperationIdBehavior;
```
MikeEdgar, line 54: Please also add to `io.smallrye.openapi.mavenplugin.MavenConfig` in the `maven-plugin` module.

MikeEdgar (changes requested):
@renegrob , thank you for the PR! This is looking pretty good. I left a few comments on some things that I think need to be changed. Please take a look.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
```diff
@@ -338,6 +339,28 @@ default Optional<Operation> processOperation(final AnnotationScannerContext cont
...
+                final String conflictingClassName = conflictingClass.name().toString();
+                final String conflictingMethodName = conflictingMethod.toString();
+                if (context.getConfig().getDuplicateOperationIdBehavior() == DuplicateOperationIdBehavior.WARN) {
+                    ScannerSPILogging.log.duplicateOperationId(operationId, className, methodName,
+                            conflictingClassName, conflictingMethodName);
+                } else {
+                    String conflictMessage = String.format(
+                            "Duplicate operationId: %s produced by Class: %s, Method: %s and Class: %s, Method: %s",
+                            operationId, className, methodName, conflictingClassName, conflictingMethodName);
+                    throw new IllegalStateException(conflictMessage);
```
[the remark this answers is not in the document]
renegrob: I guess the range should be `07900-079499` for logging and `07950-07999` for the new messages interface.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
```diff
@@ -338,6 +339,28 @@ default Optional<Operation> processOperation(final AnnotationScannerContext cont
             operation.setOperationId(operationId);
         }
 
+        // validate operationId
+        String operationId = operation.getOperationId();
+        if (operationId != null) {
+            final MethodInfo conflictingMethod = context.getOperationIdMap().put(operationId, method);
```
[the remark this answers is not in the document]
renegrob: I used now `putIfAbsent()` that way the first method stays in the map.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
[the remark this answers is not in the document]
MikeEdgar: Oops, yeah I messed that up. `07900-07949` for logging and `07950-07999` for messages.

core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
```diff
@@ -146,6 +146,10 @@ default OperationIdStrategy getOperationIdStrategy() {
         return null;
     }
 
+    default DuplicateOperationIdBehavior getDuplicateOperationIdBehavior() {
+        return DuplicateOperationIdBehavior.FAIL;
```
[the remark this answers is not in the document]
renegrob: True but with the next major release `FAIL` should become the standard.

renegrob:
@MikeEdgar I tried to apply your suggestions. Not sure if I got everything right.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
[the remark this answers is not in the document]
renegrob: Done.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/ScannerSPILogging.java
```diff
@@ -21,4 +21,8 @@ interface ScannerSPILogging { // NOSONAR (use of constants in an interface)
     @Message(id = 7902, value = "@Parameter annotation missing %s parameter. Class: %s, Method: %s, Parameter.name: %s, Parameter.in: %s")
     void missingFrameworkParam(String frameworkName, String className, String method, String paramName, String paramIn);
 
+    @LogMessage(level = Logger.Level.WARN)
+    @Message(id = 7903, value = "Duplicate operationId: %s produced by Class: %s, Method: %s and Class: %s, Method: %s")
+    void duplicateOperationId(String frameworkName, String className, String method, String conflictingClassName,
+            String conflictingMethod);
```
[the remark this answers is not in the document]
renegrob: Done.

core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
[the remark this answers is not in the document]
renegrob: Done.

tools/gradle-plugin/src/main/java/io/smallrye/openapi/gradleplugin/Configs.java
renegrob: Done.

core/src/main/java/io/smallrye/openapi/api/OpenApiConfigImpl.java
renegrob: Done.

tools/gradle-plugin/src/main/java/io/smallrye/openapi/gradleplugin/Configs.java
renegrob: Done.

extension-vertx/src/main/java/io/smallrye/openapi/vertx/VertxAnnotationScanner.java
renegrob: @MikeEdgar Is it okay to use a `HashSet` here or should I use a `LinkedHashSet` instead?

MikeEdgar: I haven't looked into any side-affects, but probably being cautious and keeping the same iteration order that was used for the `List` would be good. Specifically, I agree with using the `LinkedHashSet` :)

renegrob:
Ready for another review.

MikeEdgar:
> Ready for another review.

Thanks @renegrob . I'll try to review later today or tomorrow morning.

renegrob:
No hurry. Thank you!

renegrob:
About the failing `verfiy javadocs` build job: Local runs with `mvn -B package javadoc:javadoc -DskipTests` are successful. I guess that the job requires a `package` or even `install` goal to run first.

MikeEdgar:
> About the failing `verfiy javadocs` build job: Local runs with `mvn -B package javadoc:javadoc -DskipTests` are successful. I guess that the job requires a `package` or even `install` goal to run first.

Yes, I think we've only been lucky that it runs in the Github CI due to the Maven cache. It seems Gradle does not resolve the other projects unless they're `install`ed since it does not run in the Maven reactor. Nothing to do with your changes of course. I'm looking into this and I'll push a fix to the PR branch. Once that completes we should get a Sonar report just with some additional checks as well.

MikeEdgar (approved):
@renegrob this is all good from my perspective. Ok to merge?

renegrob:
> @renegrob this is all good from my perspective. Ok to merge?
@MikeEdgar Yes, sure. Should I update the documentation? Where is it located? (I was looking at https://github.com/smallrye/smallrye.github.io and https://github.com/smallrye/smallrye-antora-ui but didn't see anything obvious)

MikeEdgar:
@renegrob , yes thank you for bringing it up. This project isn't part of the overall SmallRye docs (yet), so maybe just another bullet point in the README [1] would be good for now.

[1] https://github.com/smallrye/smallrye-open-api#configuration-extensions

MikeEdgar (approved):

PR #1313, opened 2022-11-11
Fix special-case handling of oauth2Redirect token

### LINKED ISSUE #1304: U/I fails to render when `options` is passed to `IndexHtmlCreator#createIndexHtml`
@phillip-kruger and @MikeEdgar 

Hi, gents. 

I'm creating a Helidon Maven artifact that depends on the SmallRye OpenAPI U/I component. 

When I let the U/I use the default `index.html`, the U/I renders just right.

But, if I invoke `IndexHtmlCreator.createIndexHtml(options)`--even with a null `options` which defaults everything--the browser shows an empty screen (Safari and Firefox). Using Safari's developer mode, I see 
```
ReferenceError: Can't find variable: oar
```
And, in fact, the generated `index.html` contains the code below, referring to `oar` without declaring it.

I believe the problem is at https://github.com/smallrye/smallrye-open-api/blob/ab0d9912fbb4c7998cea713f62cc145ec17c6117/ui/open-api-ui/src/main/java/io/smallrye/openapi/ui/IndexHtmlCreator.java#L71-L73  `variableOption` is an `Option` but that line uses `.equals` against a string which never matches. It looks as if the comparison should be against `Option.oauth2RedirectUrl`.

But that change alone is not enough. I think instead of returning "-" perhaps the code should assign `replacement = "-"` and then continue on with the rest of the method, returning `null` at that point only if the replacement is null and the option in play is not `oauth2RedirectUrl`.

I tried these changes locally and it seems to work better, but I certainly haven't tested the change thoroughly.

Here's the full `index.html` delivered to the browser:
```
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="UTF-8">
        <title>Helidon OpenAPI U/I</title>
        <link rel="stylesheet" type="text/css" href="swagger-ui.css" >
        <link rel="stylesheet" type="text/css" href="theme-feeling-blue.css" >

        <link rel="shortcut icon" href="favicon.ico" type="image/x-icon">
        <link rel="icon" href="favicon.ico" type="image/x-icon">
        <link rel="stylesheet" type="text/css" href="style.css" >
    </head>

    <body>
        <div id="swagger-ui"></div>
        <script src="swagger-ui-bundle.js" charset="UTF-8"></script>
        <script src="swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
        <script>

            window.onload = function() {
                
                // If Url is relative, append the host and port
                if(oar.startsWith("/")){
                    oar = window.location.protocol + "//" + window.location.host + oar;   
                }
                
                console.log(oar);
                
                var ui = SwaggerUIBundle({
                            url: '/openapi',
                            dom_id: '#swagger-ui',
                            deepLinking: true,
                            presets: [SwaggerUIBundle.presets.apis,SwaggerUIStandalonePreset],
                            plugins: [SwaggerUIBundle.plugins.DownloadUrl],
                            layout: 'StandaloneLayout',
                            oauth2RedirectUrl: oar,
                          })
                
                
                window.ui = ui;

                document.getElementsByClassName("link")[0].innerHTML = "<a id='swaggerUiLogoLink' href='/openapi-ui'><img src='/frank.svg' alt='Helidon OpenAPI U/I' height='44' align='middle'></a><a id='swaggerUiTitleLink' href='/openapi-ui'></a>";
            }
        </script>
    </body>
</html>
```

tjquinno:
Resolves #1304 

This PR changes the special handling of the `oauth2Redirect` token replacement to:

1. Fix the comparison.
2. Instead of returning right away with the token's default value, continue processing to use the replacement value to substitute for the token in the line being processed.

Also adds a test.

### DIFF
--- a/ui/open-api-ui/src/main/java/io/smallrye/openapi/ui/IndexHtmlCreator.java
+++ b/ui/open-api-ui/src/main/java/io/smallrye/openapi/ui/IndexHtmlCreator.java
@@ -68,38 +68,38 @@ public class IndexHtmlCreator {
                     String replacement = options.get(variableOption);
                     if (replacement == null) {
                         // Special case for oauth2RedirectUrl
-                        if (variableOption.equals("oauth2RedirectUrl")) {
-                            return "-";
+                        if (!variableOption.equals(Option.oauth2RedirectUrl)) {
+                            // You want to remove this line
+                            return null;
                         }
-                        // You want to remove this line
-                        return null;
-                    } else {
-                        // Some properties can be boolean or String, if String we need to add '
-                        replacement = replacement.trim();
-                        if (BOOLEAN_OR_STRING_KEYS.contains(variableOption)) {
-                            if (!replacement.equals("true") && !replacement.equals("false")) {
-                                replacement = "'" + replacement + "'";
-                            }
-                        }
-                        // Some properties can be a String or a function, if String we need to add '
-                        replacement = replacement.trim();
-                        if (STRING_OR_FUNCTION_KEYS.contains(variableOption)) {
-                            if (!replacement.startsWith("function")) {
-                                replacement = "'" + replacement + "'";
-                            }
-                        }
-                        // Some properties are string arrays, and we need to add the ' per element
-                        if (STRING_ARRAY_KEYS.contains(variableOption)) {
-                            List<String> newArray = new ArrayList<>();
-                            String[] parts = replacement.replace("[", "").replace("]", "").split(",");
-                            for (String part : parts) {
-                                newArray.add("'" + part.trim() + "'");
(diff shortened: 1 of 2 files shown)


phillip-kruger (approved):
LGTM

phillip-kruger:
Thanks @tjquinno !

phillip-kruger:
@MikeEdgar we probably need to backport this ?

MikeEdgar:
> we probably need to backport this ?

Yeah, probably. I'm hoping to get away from the 2.x branch by having 3.1.x pass the older TCKs. If for some reason that turns out to be a problem we can do another 2.3.x/3.0.x release.

PR #1347, opened 2023-01-11
Do not set `type` for schemas generated from `java.lang.Object`

### LINKED ISSUE #1022: Support for Any schema
Currently if we expose an object, the schema is

```
@Schema(description = "The value of the characteristic", required = true)
public Object getValue() {
    return value;
}
```
gives

```
value:
    description: The value of the characteristic
    type: object
```

Is it possible to not have the type attribute or just `{}` when no other attribute are present to express an Any schema ?

`io.smallrye:smallrye-open-api-core:jar:2.1.17`

MikeEdgar:
Fixes #1022

This change to the scanner will interpret `java.lang.Object` (and equivalent wildcards) as any schema type, dropping the `type: object` from the schema that is generated today. Properties of type `Object` could hold objects any schema `type` (e.g. `String`s, `Integer`s, etc.) and it can not be assumed they are schema `object`s.

Reference: https://swagger.io/docs/specification/data-models/data-types/#any


phillip-kruger (approved):
LGTM. I wonder if we should bump a majon or minor version for this change ? How breaking is it ?

phillip-kruger:
Maybe 3.2.0  ?

MikeEdgar:
> LGTM. I wonder if we should bump a majon or minor version for this change ? How breaking is it ?

Yeah, we can do that. Historically, most of the feature PRs should have done that really :-) I think adopting semver for real might be a good thing.

PR #1352, opened 2023-01-13
Use resolved configuration and artifacts for Gradle plugin dependencies

### LINKED ISSUE #1321: Schemas for classes in dependency JARs never generated
Hello,

Not sure if this is a bug or a configuration issue, but I am trying to use the gradle plugin with a vanilla microprofile project (simple WAR generated and loaded into a TomEE instance). I have run into the following issue:

Let's assume a library A is included as compileOnly and some object MyObject from A is used either as a request body or response. Basically:

```
public Response appendNetwork(@RequestBody MyObject myObject) throws Exception {
        //do stuff with my object here.
        return Response.ok().build();
    }
```     

When I run the gradle task below everything gets generated correctly except the schema for MyObject which results in the following errors in the debug ``` [io.smallrye.openapi.runtime.scanner] SROAP04005: Could not find schema class in index: MyObject``` (assume that MyObject is included correctly as some ```com.mypackage.whatever.MyObject```)

```
generateOpenApiSpec {
    outputDirectory = new File(this.getProjectDir().toString() + "/src/main/resources/openapi")
    scanDependenciesDisable = false
}
```
Is this a limitation or am I missing some configuration here?

As a possible hint when using the 'io.swagger.core.v3.swagger-gradle-plugin' version '2.1.11' the same problem surfaces but it can be mitigated with the following configuration of the plugin, in which case MyObject is indeed found and correctly included as a schema in the OpenAPI specification:

```
resolve {
    outputFileName = 'api'
    outputFormat = 'YAML'
    prettyPrint = 'TRUE'
    classpath = sourceSets.main.output + configurations.compileClasspath
    buildClasspath = classpath
}
```

MikeEdgar:
Fixes #1321

### DIFF
--- a/tools/gradle-plugin/src/main/java/io/smallrye/openapi/gradleplugin/GradleDependencyIndexCreator.java
+++ b/tools/gradle-plugin/src/main/java/io/smallrye/openapi/gradleplugin/GradleDependencyIndexCreator.java
@@ -11,12 +11,12 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.gradle.api.artifacts.PublishArtifact;
-import org.gradle.api.artifacts.PublishArtifactSet;
+import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.logging.Logger;
 import org.jboss.jandex.CompositeIndex;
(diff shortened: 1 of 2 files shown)


MikeEdgar:
@sidohin-felix, @Hes-Siemelink - I think this PR fixes it for real this time :rofl: 

Let me know how it looks in your projects.

Hes-Siemelink:
Works for me! It now finds the domain classes defined in other projects that are linked by source and generated the OpenAPI accordingly.

Thanks Mike!

MikeEdgar:
@phillip-kruger, please take a look when you have some time. The other two PRs I have open should be ready for review also, but I expect those to need some minor modifications for conflicts after this one.

phillip-kruger (approved):
Nice !

PR #1369, opened 2023-02-05
Add configuration option to specify resource classes

anarsultanov:


core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
```diff
@@ -175,6 +175,10 @@ default Set<String> getScanExcludeProfiles() {
         return new HashSet<>();
     }
 
+    default Map<String, String> getScanResourceClasses() {
```
MikeEdgar, line 178: It might be better to have this class return the class-name/path rather than the full configuration key. You'll want to strip out the prefix in your MavenConfig implementation of this and the PR will also need an implementation in `OpenApiConfigImpl` for the non-Maven case.

extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsAnnotationScanner.java
```diff
@@ -611,6 +614,22 @@ private Collection<ClassInfo> getJaxRsResourceClasses(AnnotationScannerContext c
...
+     * Use the Jandex index to find all resource classes specified in the configuration.
+     *
+     * @param context current scanning context
+     * @return Collection of ClassInfo's
+     */
+    private Collection<ClassInfo> getConfigurationResourceClasses(AnnotationScannerContext context) {
+        return context.getConfig().getScanResourceClasses()
+                .keySet()
+                .stream()
+                .map(key -> key.substring(OpenApiConstants.SCAN_RESOURCE_CLASS_PREFIX.length()))
```
MikeEdgar, line 627: As mentioned above, `getScanResourceClasses` should return just the class names as the keys. That means the part here with `substring` won't be necessary.

extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsAnnotationScanner.java
```diff
@@ -216,7 +216,10 @@ private OpenAPI processApplicationClass(final AnnotationScannerContext context,
 
     private void processResourceClasses(final AnnotationScannerContext context, OpenAPI openApi) {
         // Now find all jax-rs endpoints
-        Collection<ClassInfo> resourceClasses = getJaxRsResourceClasses(context);
+        Collection<ClassInfo> resourceClasses = new ArrayList<>();
+        resourceClasses.addAll(getJaxRsResourceClasses(context));
+        resourceClasses.addAll(getConfigurationResourceClasses(context));
```
MikeEdgar, line 221: In addition to this, once the class is being processed, you'll need to again look at the config's `getScanResourceClasses` to determine the path (since there is no `@Path`)

MikeEdgar (commented):
@anarsultanov this looks like a good start. I left a few comments on how to continue.

core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
anarsultanov: I agree, will update it to return class name without prefix.
I'm not sure how should it be added to `OpenApiConfigImpl` since it gets properties from `org.eclipse.microprofile.config.Config` and I do not know how it gets populated.

extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsAnnotationScanner.java
anarsultanov: This is exactly what I didn't figure out how to do.
As I understand in `JaxRsParameterProcessor` we have to create a parameter from the path specified in the configuration, so I tried something like this, but without success:
```
    @Override
    protected void processPathParameters(ClassInfo resourceClass, MethodInfo resourceMethod, ResourceParameters parameters) {
        ClassInfo resourceMethodClass = resourceMethod.declaringClass();

        // This is what I added
        readConfigurationPathParameter(resourceClass);
        
        /*
         * Phase I - Read class fields, constructors, "setter" methods not annotated with JAX-RS
         * HTTP method. Check both the class declaring the method as well as the resource
         * class, if different.
         */
        readParametersInherited(resourceMethodClass, null, false);

        if (!resourceClass.equals(resourceMethodClass)) {
            /*
             * The resource class may be a subclass/implementor of the resource method class. Scanning
             * the resource class after the method's class allows for parameter details to be overridden
             * by annotations in the subclass.
             */
            readParameters(resourceClass, null, true);
        }

        parameters.setPathItemParameters(getParameters(resourceMethod));
    }

    private void readConfigurationPathParameter(ClassInfo resourceClass) {
        FrameworkParameter frameworkParam = JaxRsParameter.PATH_PARAM.parameter;
        String path = scannerContext.getConfig().getScanResourceClasses()
                .get(OpenApiConstants.SCAN_RESOURCE_CLASS_PREFIX + resourceClass.name().toString());
        if (path != null) {
            readParameter(
                    new ParameterContextKey(path, frameworkParam.location, frameworkParam.defaultStyle),
                    null,
                    frameworkParam,
                    path,
                    resourceClass,
                    false);
        }
    }
```

core/src/main/java/io/smallrye/openapi/api/OpenApiConfig.java
MikeEdgar: I think the method in `OpenApiConfigImpl` will be very similar to the `getSchemas` method.

extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsAnnotationScanner.java
MikeEdgar: I think you'll need to change `pathOf` [1] to get the new configuration value in the `CLASS` case of the switch.

[1] https://github.com/smallrye/smallrye-open-api/blob/2cd874153e0d48f9946ed6d4190b6488cf53c59e/extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsParameterProcessor.java#L236-L245

anarsultanov: Thanks, it seems to work. Which do you think should take precedence, the path annotation or the configuration property if both are present?

MikeEdgar: I think overriding the annotation with the configuration would make sense.

anarsultanov: Agree. I've pushed a working solution, just need to add some tests.

anarsultanov: I wanted to add tests today, but to be honest I'm not quite sure where and how it should be tested since I didn't find any tests for the existing options. In addition, sonarcloud complains about uncovered lines in `JaxRsAnnotationScanner`, but as far as I can see this class is not covered by tests at all, not just the lines that I changed / added.

MikeEdgar: @anarsultanov the tests are generally code to use the main scanner entry point `OpenApiAnnotationScanner`. See the tests in `io.smallrye.openapi.runtime.scanner` in the jaxrs module's test directory.

You could add a new test class there and follow the pattern from the others to set your desired configuration, call the `scan` method, and compare the result against some known static JSON file. The class should extend `IndexScannerTestBase` where you can find methods to create simple implementations of the `OpenApiConfig` interface, for example `dynamicConfig(String key, Object value)`.

extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsParameterProcessor.java
```diff
@@ -32,21 +32,21 @@
...
- * javax.ws.rs.core.PathSegment PathSegment are not currently supported.
+ * Note, javax.ws.rs.PathParam PathParam targets of javax.ws.rs.core.PathSegment PathSegment are not currently supported.
  *
  * @author Michael Edgar {@literal <michael@xlate.io>}
- *
  */
 public class JaxRsParameterProcessor extends AbstractParameterProcessor {
 
+    private AnnotationScannerContext scannerContext;
+
```
MikeEdgar, line 42: This shouldn't be needed. The `scannerContext` in `AbstractParameterProcessor` should be visible.

extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsAnnotationScanner.java
anarsultanov: Added some simple tests, please let me know if you think more cases need to be tested.

MikeEdgar (approved):
Looks good to me, thank you @anarsultanov

Issue #1379, opened 2023-02-15
OpenAPI scanner reports duplicate operation id for exactly the same methods of the same class


mswiderski:
I started to see a warning about duplicate operation id

````
SROAP07903: Duplicate operationId: abortTask_results_recivers_0 produced by Class: org.acme.mq.ReciversResource, Method: javax.ws.rs.core.Response abortTask_results_recivers_0(javax.ws.rs.core.HttpHeaders, java.lang.String, java.lang.String, java.lang.String, java.lang.String, java.util.List<java.lang.String>, boolean) and Class: org.acme.mq.ReciversResource, Method: javax.ws.rs.core.Response abortTask_results_recivers_0(javax.ws.rs.core.HttpHeaders, java.lang.String, java.lang.String, java.lang.String, java.lang.String, java.util.List<java.lang.String>, boolean)
````

but it actually points to the same class/method so it is the same thing. 

This class is being generated as part of quarkus extension and maybe because of that it gets bit confused. 

Would it be possible to add extra check in the scanner to ignore if the class and method are the same?

MikeEdgar:
I think checking if it's the same method previously encountered would be fine. Is the resource class mentioned the parent of other resource classes?

mswiderski:
> I think checking if it's the same method previously encountered would be fine

great to hear

> Is the resource class mentioned the parent of other resource classes?

Not sure I understand this one, do you mean if there are other resource classes in the same package? If so then yes, there might be but non of them have any conflicting method names.

MikeEdgar:
I just mean 

> > Is the resource class mentioned the parent of other resource classes?
> 
> Not sure I understand this one, do you mean if there are other resource classes in the same package? If so then yes, there might be but non of them have any conflicting method names.

I am wondering if you have AbstractResource with the conflicting method and then `Resource1 extends AbstractResource` and `Resource2 extends AbstractResource`. I'd like to understand why the one method is scanned twice.

mswiderski:
ah in that case no, there is no parent class that this one extends.

MikeEdgar:
Do you have a small reproducer case or app to demonstrate the problem?

mswiderski:
that's the thing it is not a simple reproducer but let me try to debug it more to figure out why it has duplicates. 

As mentioned, the classes are generated at build time so not easy to give a simple reproducer.

mswiderski:
[...]

PR #1382, opened 2023-02-23
remove duplicates in JAXRS scanner by comparing thier fqcn (dotname)

mswiderski:
replaces `distinct` with truest and comparator based on `DotName` to address duplicates issue #1379


MikeEdgar (approved):
Nice catch @mswiderski , thanks for the PR

PR #1384, opened 2023-02-23
Switch to smallrye-parent version 39

MikeEdgar:


MikeEdgar:
This PR has a large number of files impacted by an update to the formatter plugin that now removes whitespace in JavaDoc comments that was not removed previously.

phillip-kruger (approved):

PR #1389, opened 2023-03-01
Allow for alternate properties file to determine MP OpenAPI spec level

### LINKED ISSUE #1333: Don't require the pom.properties file to be present on the classpath
When we implement MP OpenAPI in open liberty, we repackage the org.eclipse.microprofile.openapi jar and generate new maven metadata so it does not contain the pom.properties file in the expected location.

Can we adapt the code added in #1269 to default to the latest version if this file is not found? (A different solution would also be fine, but this one seems the most straightforward.)

MikeEdgar:
Fixes #1333

Allow MP OpenAPI specification version to be given via properties file `io.smallrye.openapi.api.util.VersionUtil.properties` on the classpath.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/util/VersionUtil.java
+++ b/core/src/main/java/io/smallrye/openapi/api/util/VersionUtil.java
@@ -2,9 +2,13 @@ package io.smallrye.openapi.api.util;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.UncheckedIOException;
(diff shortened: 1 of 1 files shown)


MikeEdgar:
@Azquelt , please review and let me know if this option will work for #1333.

Azquelt (approved):
Yes, this looks good to me, thank you.

phillip-kruger (approved):

PR #1407, opened 2023-03-16
Map java.io.File to binary-formatted string in generated schema

MikeEdgar:
Originates from https://github.com/quarkusio/quarkus/discussions/31807


phillip-kruger (approved):
Nice one !

PR #1425, opened 2023-03-24
Gradle plugin: Add mandatory dependency to smallrye-config

### LINKED ISSUE #1426: Gradle plugin broken in v3.3.1
PR with a fix in #1425, reproducer with #1424

snazy:
Using the Gradle plugin (v3.3.1) otherwise leads to
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':nessie-model:generateOpenApiSpec'.
> io/smallrye/config/SmallRyeConfigBuilder
```
which is a `ClassNotFoundException`.

Fixes #1426

### DIFF
--- a/tools/gradle-plugin/pom.xml
+++ b/tools/gradle-plugin/pom.xml
@@ -36,6 +36,11 @@
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-open-api-vertx</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.smallrye.config</groupId>
+            <artifactId>smallrye-config</artifactId>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
 
     <build>


MikeEdgar:
Thank you for fixing this. Why is it that the Gradle dependency for config is not enough?

snazy:
(Fat-finger close/open issue)

snazy:
Gradle would publish its own Pom to plugins.gradle.org, but it’s still better to keep both in sync, since users could prefer to pull Gradle plug-ins from Maven Central (we did that when Gradle‘s plugin repo was a bit unreliable).

MikeEdgar:
> Gradle would publish its own Pom to plugins.gradle.org, but it’s still better to keep both in sync, since users could prefer to pull Gradle plug-ins from Maven Central (we did that when Gradle‘s plugin repo was a bit unreliable).

I'm admittedly ignorant about the Gradle ecosystem, unfortunately. I didn't realize there was crossover between the two like that. I'll go ahead any merge this one, thanks again.

MikeEdgar (approved):

PR #1435, opened 2023-04-07
Gradle-Plugin: prepare for Gradle configuration cache

snazy:
Gradle's configuration cache aims to prevent running (expensive) code to configure the build (tasks, et al). Preventing configuration can save a (comparatively) huge amount of time, especially on small/tiny CI machines (looking at you, GH standard hosted runner).

But to be able to use the configuration cache, the build itself and all plugins must respect a couple of rules. Gradle types that must not be referenced during task _execution_ are for example: `Project`, `Configuration`, `Task` and some more. Those need to be replaced with either more specialized types or with different mechanisms.

Test note: The Quarkus plugin in all released versions (at least up to 3.0.0.CR1) cannot be used with Gradle's configuration cache. The tests that use the Quarkus plugin do not enable Gradle's configuration cache.

See [Gradle build cache requirements](https://docs.gradle.org/8.0.2/userguide/configuration_cache.html#config_cache:requirements).

### DIFF
--- a/tools/gradle-plugin/build.gradle
+++ b/tools/gradle-plugin/build.gradle
@@ -49,6 +49,9 @@ test {
     // multiple Gradle invocations executed from the same testing JVM cause problems, so fork
     // for each test, and make sure we only run one Gradle execution in each test
     forkEvery 1
+    // For Gradle testing :(
+    jvmArgs("--add-opens=java.base/java.lang.invoke=ALL-UNNAMED",
+            "--add-opens=java.base/java.util=ALL-UNNAMED")
 }
 
 gradlePlugin {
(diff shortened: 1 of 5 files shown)


tools/gradle-plugin/build.gradle
```diff
@@ -49,6 +49,9 @@ test {
     // multiple Gradle invocations executed from the same testing JVM cause problems, so fork
     // for each test, and make sure we only run one Gradle execution in each test
     forkEvery 1
+    // For Gradle testing :(
+    jvmArgs("--add-opens=java.base/java.lang.invoke=ALL-UNNAMED",
```
snazy, line 53: This won't work with Java < 9, hope that's okay or shall I add a version-gate?

MikeEdgar: I think this is fine - we're only running CI with 11 and 17 anymore.

MikeEdgar (approved):
Looks good to me, thank you @snazy

Issue #1370, opened 2023-02-07
Allow (de)activating annotation scanners via the maven plugin config


Spindl:
## Summary 
When using the maven plugin for generating the openapi.json files, it's not possible to deactivate scanners that are not required. 
For example, we do not use spring or vert.x, but the scanners are running still. 
Also, this prevents us from extending the generated document with additional information.

The method `OpenApiAnnotationScanner#scan` already provides the means of filtering the scanners that are executed, but the maven plugin does not offer a way of setting the filter property. So it would be nice if that could be added.

## Rationale
I found [this discussion](https://github.com/smallrye/smallrye-open-api/issues/1255 ) and we have the same problem basically, so I want to elaborate further:

We also want to extend the generated document with additional information, like special query parameters which are handled by jax-rs filters and therefore aren't present in the endpoint method signatures. 
We tried to achieve that by subclassing the `JaxRsAnnotationScanner`, and adding the information after e.g. `processResponse` has been called since the context, the method info, and the operation are available there, and the information scanned from the annotations is already there. And generally, this does work as expected.
The only problem is that the added information gets overwritten if the original `JaxRsAnnotationScanner` is executed after our custom one. The order in which the scanners are called varies since they are loaded via the java `ServiceLoader` and stored in a hash map, and we found no reasonable way of making sure that our added information persists. 
If we just could disable the other scanners, that would make things easier 😇.

Issue #1255, opened 2022-09-13
Adding a custom AnnotationScanner


Raptoer:
Is is possible to add a custom AnnotationScanner in the project we're working with?

I know I could add it as a java service if I packaged it into a jar, but I want it to load from the project that the maven plugin is currently running on.

MikeEdgar:
Hi @Raptoer, what is your requirement? There may be better-supported alternate approaches that rely on standard interfaces intended for application use. The `AnnotationScanner` interface is an internal type and shouldn't necessarily be considered for use by applications.

Raptoer:
I’m trying to read nonstandard annotations in order to alter the open api output. I’ve tried using filters, but by the time you’re at that stage you’ve lost the association to the original method you’re reading from.

phillip-kruger:
There are ways in that you can get the original method. We do it in Quarkus. See https://github.com/quarkusio/quarkus/blob/b60ad38c41ed42d34ec1fe86cc5934647c92440f/extensions/smallrye-openapi/deployment/src/main/java/io/quarkus/smallrye/openapi/deployment/filter/AutoTagFilter.java#L51

as an example

MikeEdgar:
@Raptoer , what type of annotations are you dealing with? Something entirely custom to your application domain?

Raptoer:
So I'm trying to read @PreAuthorize, the spring security annotation, to somehow indicate on the openAPI document about the permissions required.

Raptoer:
So I'm trying to read @PreAuthorize, the spring security annotation, to somehow indicate on the openAPI document about the permissions required

phillip-kruger:
This sounds similar to https://github.com/quarkusio/quarkus/blob/main/extensions/smallrye-openapi/deployment/src/main/java/io/quarkus/smallrye/openapi/deployment/filter/AutoRolesAllowedFilter.java

If you create a small reproducer and the expected entries in the schema document, I can maybe build it in the above filter.

Raptoer:
The problem with these filters is that I'm using the maven plugin, and I'd rather not duplicate the expansive SmallRyeOpenApiProcessor that the quarkus extension uses.
The way the maven plugin works I don't get a constructor to hand in extra data.
I tried just using the class loader in the filter when it get activated, but I don't seem to have access to my application classes at that time.

MikeEdgar:
> I tried just using the class loader in the filter when it get activated, but I don't seem to have access to my application classes at that time.

[...]

PR #1475, opened 2023-05-16
Support for readers and filters to access the Jandex index

MikeEdgar:
Allows implementations of `OASModelReader` and `OASFilter` to optionally provide a single-argument constructor accepting a Jandex `IndexView` to be used internally.

Relates to #1255 and #1370

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,14 +86,14 @@ jobs:
           java-version: 11
 
       - name: maven cache
-        uses: actions/cache@v3
+        uses: actions/cache/restore@v3
         with:
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
 
       - name: generate javadocs
-        run: mvn -B package javadoc:javadoc -DskipTests
+        run: mvn -B install javadoc:javadoc -DskipTests
 
   tck-reporting:
     runs-on: ubuntu-latest
--- a/core/src/main/java/io/smallrye/openapi/runtime/OpenApiProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/OpenApiProcessor.java
@@ -2,9 +2,11 @@ package io.smallrye.openapi.runtime;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Constructor;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.Callable;
 
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigProvider;
@@ -12,6 +14,7 @@ import org.eclipse.microprofile.openapi.OASFilter;
 import org.eclipse.microprofile.openapi.OASModelReader;
 import org.eclipse.microprofile.openapi.models.OpenAPI;
 import org.jboss.jandex.IndexView;
+import org.jboss.jandex.Indexer;
 
 import io.smallrye.openapi.api.OpenApiConfig;
 import io.smallrye.openapi.api.OpenApiDocument;
@@ -26,6 +29,8 @@ import io.smallrye.openapi.runtime.scanner.OpenApiAnnotationScanner;
  * @author eric.wittmann@gmail.com
  */
 public class OpenApiProcessor {
+
+    static final IndexView EMPTY_INDEX = new Indexer().complete();
 
     private OpenApiProcessor() {
     }
@@ -71,8 +76,8 @@ public class OpenApiProcessor {
         }
         // Filter and model
         if (config != null && classLoader != null) {
-            OpenApiDocument.INSTANCE.modelFromReader(modelFromReader(config, classLoader));
-            OpenApiDocument.INSTANCE.filter(getFilter(config, classLoader));
+            OpenApiDocument.INSTANCE.modelFromReader(modelFromReader(config, classLoader, index));
+            OpenApiDocument.INSTANCE.filter(getFilter(config, classLoader, index));
         }
 
         OpenApiDocument.INSTANCE.initialize();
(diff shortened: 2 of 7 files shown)


MikeEdgar:
@Raptoer, this PR is for a concept that would allow you to use the Jandex index from an instance of an `OASFilter` in order to extract the Spring security information you mentioned on #1255. You could associate the `Operation`s in the `OpenAPI` via operation Id or with some other custom `@Extension` of your choosing. Wdyt?

@Spindl, this is also meant to assist with your issue #1370. Using an `OASFilter` provided with an `IndexView` that runs following the JAX-RS annotation scanner should allow you fill in or replace the pieces in the `OpenAPI` model. Please share your thoughts if that could work.

For both issues, I would say implementing an `AnnotationScanner` is somewhat discouraged since it's meant to be an internal component and could change in ways that break external implementations, even in non-major releases of the scanner.

Spindl:
Hi @MikeEdgar,

thank you for considering helping us with our problem.
I fear that having the index and the operation does not do the trick for us. 
We need the `MethodInfo` object of the method the operation refers to, e.g. because we look for additional annotations or thrown exceptions to extend the openapi documentation.

I understand that the Scanners are considered internal API, and would also prefer not to hook into the scanning process like that 😅 
But if there is no way to get to the `MethodInfo` object from the operation I fear we have no other options.

Spindl:
Being able to filter the active scanners would allow us to get rid of another ugly workaround though, which we need for now since we can't influence the order in which the scanners run. 😬

MikeEdgar:
> We need the `MethodInfo` object of the method the operation refers to, e.g. because we look for additional annotations or thrown exceptions to extend the openapi documentation.

@Spindl, having the index would allow that. Suppose you have an endpoint method:
```java
package com.example;

@Path("/widgets")
public class WidgetsResource {

    @GET
    @Path("{widgetId}")
    @Operation(operationId = "getWidget")
    @com.example.MyAnnotation(value = "Something Important")
    public Response getWidgets(String widgetId) {
       ...
    }

}
```
You could do something like in the filter:
```java
public class MyFilter implements OASFilter {

    private IndexView index;
    
    public MyFilter(IndexView index) {
        this.index = index;
    }

    @Override
    public Operation filterOperation(Opertion operation) {
        if ("getWidget".equals(operation.getOperationId()) {
            MethodInfo getWidgets = index
                .getClassByName("com.example.WidgetsResource")
                .method("getWidget", Type.create(
                    DotName.createSimple("java.lang.String"), 
                    Type.Kind.CLASS));
            AnnotationInstance myAnnotation = getWidgets.annotation("com.example.MyAnnotation");
            // Do something with `myAnnotation` and the `operation`
        }
    }
}
```

MikeEdgar:
> I guess we could maybe try to "put the cart before the horse" and guess the class the method belongs to from the operation ID (which is `{tag}_{methodName}` in our case) and all classes having the `@Tag` annotation or something like that. But that does not really seem any better than what we have now.

@Spindl , there are other options to access the `MethodInfo` that might work for you. For example, given an `Operation` model in the filter, you can find the matching `@Operation` annotation (assuming you have them to set the values to `{tag}_{methodName}`) like this:
```java
index.getAnnotations(org.eclipse.microprofile.openapi.annotations.Operation.class)
    .stream()
    .filter(annotation -> operation.getOperationId().equals(annotation.value("operationId").asString()))
    .map(AnnotationInstance::target)
    .map(AnnotationTarget::asMethod)
    .findFirst()
    .ifPresent(methodInfo -> {
        // Do something with the methodInfo
    });
```

Something similar may work for schemas if there is a way to navigate back from the annotations you are interested in.

Spindl:
Hi @MikeEdgar 
sorry for the late reply, I did get sidetracked a little bit at work 😅 

I assume we could do something like that for the operations, even though we still couldn't get rid of our custom scanner since we don't annotate our endpoints with the @Operation annotation, but rather also calculate the operation_id in the scanner.

And for the Schema I see no way this could work. 
We don't explicitly annotate any of our model classes either, and without the fully qualified class name, it would be difficult to find the correct ClassInfo object for a Schema.

ChMThiel:
I have same problems like @Spindl : I want to add [BeanValiation-Infos](https://quarkusio.zulipchat.com/#narrow/stream/294206-smallrye/topic/custom.20bean.20validation.20in.20openApi) annotated at classes/methods/fields to OpenApi. so i need the relation between an OpenApi-Schema and the class/method/field.

ChMThiel:
any progress on this? Only the low test-coverage prevents the merge  :thinking:

MikeEdgar:
> any progress on this? Only the low test-coverage prevents the merge thinking

If this is useful to you as-is, I'm happy to fix up this PR and merge it. However, adding in some relation between the OpenAPI model pieces and the source that generated them will require something more involved.

ChMThiel:
I hope it will suffice for me, to get access to the jandex. 
We have a standard name-schema in our classes so i hope to get the releation between OpenApi <-> annotated field/method etc.

Nevertheless: my goal is to document custom BeanValidations in OpenApi. 
The best solution for this would be to have a kind of 'OpenApiFilter for BeanValidations': a way to filter/enrich the results of `io.smallrye.openapi.runtime.scanner.dataobject.BeanValidationScanner`

phillip-kruger (approved):
Thanks @MikeEdgar

ChMThiel:
@MikeEdgar 
Hello Mike, i was wondering why you provide only an EMPTY_INDEX in OpenApiDocumentService? (see [commit)](https://github.com/quarkusio/quarkus/commit/75e841ec7ff0191b5c7c45c08a507403e95ac7f3).
So the index is available during build-time, but during runtime it's always empty.

phillip-kruger:
The Jandex index is not available at runtime, the openapi schema document created from the index in build time is available.

MikeEdgar:
Right - that is obviously specific to Quarkus, so if you need to do something at run time that you could not have done at build time, your reader/filter will need to fall back to some other mechanism. I suppose you could also build your own Jandex index at build time with your app classes and load it yourself in a filter.

ChMThiel:
My problem is that my openApiFilter is triggered twice: once during build time and once during runtime.
Even after setting the `config quarkus.smallrye-openapi.store-schema-directory`. Each time it's given a fresh input OpenApi (second run seems not to get the results of the first run). And the second time (runtime) creates the resulting OpenApi.

My hope was to find a configuration so that quarkus creates and filters OpenApi during build-time and stores the result in the jar. So during runtime this openApi could be used without any further filtering/processing.

MikeEdgar:
> Each time it's given a fresh input OpenApi (second run seems not to get the results of the first run). And the second time (runtime) creates the resulting OpenApi.

That doesn't seem correct. I would expect the runtime input (loaded in `OpenApiDocumentService`) to have the output from the build time OpenAPI model. Are you certain of that?

MikeEdgar:
The output of the build should be placed in `META-INF/quarkus-generated-openapi-doc.JSON`. That would be in `target/quarkus-app/quarkus/generated-bytecode.jar`. How does the OpenAPI look there?

ChMThiel:
The openApi unter `META-INF/quarkus-generated-openapi-doc.JSON` has no tags at all. I seems it does not contain the result of my filter.

MikeEdgar:
Ok, I see. We only execute the app's filter on the version written to the `store-schema-directory` [1]. The copy placed in `META-INF` was not run with the filter. @phillip-kruger do you see any reason we can't just move the processing of the app filter earlier in the process, from `storeDocument` to `loadDocument`?

[1] https://github.com/quarkusio/quarkus/blob/1e3fbb99d7e9c3db994ea6ffe32be0f831386054/extensions/smallrye-openapi/deployment/src/main/java/io/quarkus/smallrye/openapi/deployment/SmallRyeOpenApiProcessor.java#L1102-L1104

MikeEdgar:
@ChMThiel , another option may be to use an `OASModelReader` instead. That would run at build time and you would construct an `OpenAPI` with just the pieces you want. It would later be merged with the model from annotations.

phillip-kruger:
The document that is stored and the one we pass to runtime might not be the same. Except if you have a reason for storing it, there should be no reason to do so. That feature was a request from users that store and version control the stored schema, so that they can compare between releases. The reason we run the filter at build time in that case is that it only runs during build, and not at runtime. For the normal flow of events the user's filters are run at runtime.

We have two options here I think. 1) Allow users to define the stage a Filter should run, with the default being runtime as it is now. 2) Add built-in support for what you are trying to do (BeanValidationScanner).

Maybe we should do both ?

ChMThiel:
I do not only filter/adjust the OpenApi in my OASFilter, i want to add infos. To get these infos (that are not present in given OpenApi) i had to use some tricks/hacks: During runtime i had access to CDI and was able to request classes implementing certain marker-interface.
Example: I have certain entity envers-audited. I want to add infos about that to the schema-description. In Order to do so i had to add a marker-interface, etc.

Now that the jandex is available, i can replace the CDI-hack with the an elegant Jandex-request (For my audit-example `jandex.getAnnotations(Audited.class)`). But because jandex is only available during build-time, i need an option to avoid running the filter during runtime again but keep/use the result of the build-time filter instead.

For the special case of adding _BeanValidation-infos_ imho the optimal solution would be do make the BeanValidatioScanner more flexible/generic. But with something like
```java
jandex.getKnownClasses().stream()
             .filter(ClassInfo::isAnnotation)
             .filter(i -> i.hasAnnotation(Constraint.class))
```
I can get all BeanValidation-annotations in the OASFilter, connect those to the matchinng  OpenApi-Schema  and add the info by myself.

I would really appreciate if you support both options because both options have it's use cases, but to _define the run-stage of the filter is what i need most_.

phillip-kruger:
Cool, I can look at making the change to allow running user filters during build. That is a Quarkus change I think. /cc @MikeEdgar

PR #1505, opened 2023-06-30
Expose annotated XML attributes/elements when XmlAccessorType is NONE

### LINKED ISSUE #1504: XML attributes/elements ignored when class is annotated with XmlAccessType.NONE
See https://github.com/quarkusio/quarkus/issues/34402

MikeEdgar:
Fixes #1504

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/constants/JaxbConstants.java
+++ b/core/src/main/java/io/smallrye/openapi/api/constants/JaxbConstants.java
@@ -1,7 +1,11 @@
 package io.smallrye.openapi.api.constants;
 
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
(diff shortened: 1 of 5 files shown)


phillip-kruger (approved):

troosan:
@MikeEdgar There is an issue when the name in the XmlResponse does not match the method name
image

it then returns
image

troosan:
Also the XmlType does not seem to be taken into account, for the name of the type and the order of the elements
image

troosan:
the good new though is that the change seems to be compatible with 2.x (except for the Annotations class that was added later and which I had to change back to TypeUtil

MikeEdgar:
Thanks for reporting this @troosan . It seems the XML support can be improved a bit. Would you mind putting your findings about the missing name support into a new issue?

troosan:
> Thanks for reporting this @troosan . It seems the XML support can be improved a bit. Would you mind putting your findings about the missing name support into a new issue?

done, see #1510

PR #1506, opened 2023-06-30
Lower severity for missing framework parameter logging to DEBUG

### LINKED ISSUE #1264: Standalone `@Parameter` annotation and warning logs
@lthasle , this issue is to further discuss and understand your requirements for using a "detached" `@Parameter` annotation. How do you access the header that is being documented? Are you using the JAX-RS `HttpHeaders` object or something else where there is no matching JAX-RS parameter?

See for original issue https://github.com/quarkusio/quarkus/issues/28006
Relates to changes made in #1197

MikeEdgar:
Closes #1264 

Log a slightly modified message at DEBUG when a `@Parameter` is found without an equivalent framework parameter, e.g. `@PathParam`. Using `@Parameter` as a standalone annotation has valid use cases (see #1264) and should not be logged as a warning.


phillip-kruger (approved):
LGTM

PR #1522, opened 2023-07-28
Adding a configuration property for the Maven and Gradle plugin to filter output file types

### LINKED ISSUE #1511: Option to skip generation of json files
In our team we need only yaml file generation via maven-plugin but there is no option to define we dont need json files.
Is there already a way how to do it?

fabiobrz:
Adding the "outputFileTypeFilter" property to the Maven and Gradle plugin configuration, so that a user can decide whether to generate `*.yaml` ("YAML"), `*.json` ("JSON") file types, or both ("ALL").

Fix #1511

### DIFF
--- a/tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/GenerateSchemaMojo.java
+++ b/tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/GenerateSchemaMojo.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.maven.artifact.DependencyResolutionRequiredException;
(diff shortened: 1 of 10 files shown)


tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/GenerateSchemaMojo.java
```diff
@@ -419,6 +432,7 @@ private Map<String, String> getProperties() throws IOException {
         addToPropertyMap(cp, OpenApiConstants.SCAN_PROFILES, scanProfiles);
         addToPropertyMap(cp, OpenApiConstants.SCAN_EXCLUDE_PROFILES, scanExcludeProfiles);
         addToPropertyMap(cp, OpenApiConstants.SCAN_RESOURCE_CLASS_PREFIX, scanResourceClasses);
+        addToPropertyMap(cp, OpenApiConstants.INFO_TITLE, infoTitle);
```
MikeEdgar, line 435: Is this intended? Same as line 422.

MikeEdgar (commented):
Changes look good, just one question on something maybe unexpected in the Maven plugin.

tools/maven-plugin/src/main/java/io/smallrye/openapi/mavenplugin/GenerateSchemaMojo.java
fabiobrz: :man_facepalming: Definitely not, a copy/paste leftover of mine.
Thanks for catching it, should be fixed now

MikeEdgar (dismissed):

MikeEdgar:
Thanks for the PR @fabiobrz . Ok to merge?

fabiobrz:
> Thanks for the PR @fabiobrz . Ok to merge?

Welcome! Sure thing, go ahead 🚀

fabiobrz:
> > Thanks for the PR @fabiobrz . Ok to merge?
> 
> Welcome! Sure thing, go ahead 🚀 

Or maybe the README? WDYT?

MikeEdgar:
Good idea, that would be helpful to make the new options more discoverable.

MikeEdgar (approved):
Thanks again @fabiobrz , merging :+1:

PR #1533, opened 2023-08-07
Fix issue 1532, Gradle plugin wrongly generates a YAML file with JSON contents

### LINKED ISSUE #1532: Gradle task writes out JSON in the YAML file
The following line writes out the json serialized value instead of the serialized yaml that is generated at the top of this method. 

https://github.com/smallrye/smallrye-open-api/blob/7abb44784ba4d9bda5e90188e40f8a600d850a0d/tools/gradle-plugin/src/main/java/io/smallrye/openapi/gradleplugin/SmallryeOpenApiTask.java#L319

fabiobrz:
Subject says it all :)

Fix #1532


MikeEdgar (approved):
Thanks @fabiobrz

PR #1568, opened 2023-09-11
Fix gradle multithread issue

mslavik-leidos:
My team and I have run into a race condition on generating the schema for our composite gradle build. This is due to gradle trying to build multiple projects at the same time and this plugin's use of the OpenApiDocument.INSTANCE singleton. Using the newInstance method instead gives a fresh document to each task, this lets us continue to build with multithreading on and avoid the race condition.


MikeEdgar (approved):
Looks good to me, thank you for the PR @mslavik-leidos

PR #1574, opened 2023-09-20
Register full schema of array items for known types

### LINKED ISSUE #1573: Incorrect open api document generation when a resource uses an array of ZonedDateTime
Given 
```
@ApplicationScoped
@Path("a")
public class ARestResource {

    @GET
    @Operation(summary = "a")
    public ZonedDateTimeArrayWrapper get() {
        return new ZonedDateTimeArrayWrapper();
    }

}
```
with 

```
public class ZonedDateTimeArrayWrapper {

    ZonedDateTime[] now;

    public ZonedDateTime[] getNow() {
        return now;
    }

    public void setNow(ZonedDateTime[] now) {
        this.now = now;
    }

}
```

Then the generated open api yaml file does not contain the schema for ZonedDateTime. Worse, since the schema is registered in the schema registry (but not in the open api document), if another resource which also uses a ZonedDateTime is processed afterwards (e.g. BRestResource), then the type will not be reprocessed. In the end, you get an invalid open api yaml file.

Tested with smallrye-open-api version 3.5.1 via quarkus-smallrye-openapi version 3.3.3

MikeEdgar:
Fixes #1573

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeProcessor.java
@@ -164,13 +164,16 @@ public class TypeProcessor {
         // Only use component (excludes the special name formatting for arrays).
         TypeUtil.applyTypeAttributes(componentType, itemSchema);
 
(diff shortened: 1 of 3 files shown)


phillip-kruger (approved):
Nice !

PR #1575, opened 2023-09-20
Guard against self-referencing schemas, stabilize scan order with sort

### LINKED ISSUE #1565: @Schema causes self-reference on field's type
## Problem

The addition of `@Schema` on a class with value object members randomly causes a self-reference on the value object's open API definition. 

This problem is different from https://github.com/smallrye/smallrye-open-api/issues/1466 and https://github.com/smallrye/smallrye-open-api/issues/1507 in that the owning class's `@Schema` annotation causes this issue.

## How to replicate

The following code 
```java
@ApplicationScoped
@Path("/")
public class Resource {
    @GET
    public MyClass getMyClass(){
       return new MyClass(new MyValueClass());
    }

    @Schema(description = "some description") // <-- Removal of this schema annotation yields expected type of value the object.
    public record MyClass(MyValueClass alongerRecord1) {
    }

    @Schema(type = SchemaType.STRING)
    public record MyValueClass() {
    }
}
```

yields the `openapi.yaml`:

```yaml
---
openapi: 3.0.3
info:
  title: openapi-swaggerui-quickstart API
  version: 1.0.0-SNAPSHOT
paths:
  /:
    get:
      tags:
      - Resource
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/MyClass'
components:
  schemas:
    MyClass:
      description: some description
      type: object
      properties:
        alongerRecord1:
          $ref: '#/components/schemas/MyValueClass'
    MyValueClass:
      $ref: '#/components/schemas/MyValueClass'  # <--- self-reference
  securitySchemes:
    SecurityScheme:
      type: http
      description: Authentication
      scheme: basic
```

There seems to be some randomness  to the presence of the self-reference dependent on the classes' names. Most of the times there is a self reference but I have received the expected string type of the value class. Changing a value object's class name may cause the correct schema to appear on the value-object type.

We use quarkus version 3.3.1 and the example above is based on https://quarkus.io/guides/openapi-swaggerui.

MikeEdgar:
Fixes #1565

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiAnnotationScanner.java
@@ -242,6 +242,7 @@ public class OpenApiAnnotationScanner {
                 .stream()
                 .filter(this::annotatedClasses)
                 .map(annotation -> Type.create(annotation.target().asClass().name(), Type.Kind.CLASS))
+                .sorted(Comparator.comparing(Type::name)) // Process annotation classes in predictable order
                 .forEach(type -> SchemaFactory.typeToSchema(context, type, null, context.getExtensions()));
     }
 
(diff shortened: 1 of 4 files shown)


MikeEdgar:
@phillip-kruger please take a look when you get a chance. This will conflict with #1574 , so leaving it as draft for now. I'll fix the conflicts after that's merged if you're OK with this one.

phillip-kruger (dismissed):
LGTM

PR #1584, opened 2023-09-28
Only make json the default for POJO

phillip-kruger:
Related to https://github.com/quarkusio/quarkus/issues/34700


phillip-kruger:
Sorry, something not right, will recreate this later

PR #1585, opened 2023-09-28
Default content type

phillip-kruger:
This PR allows different default content type for pojo return and primitives

Related to https://github.com/quarkusio/quarkus/issues/34700


MikeEdgar (approved):
:+1:

PR #1590, opened 2023-10-03
Default content type for wrapped types primitive

phillip-kruger:
Adding on to #1585  supporting wrapped responses

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractAnnotationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractAnnotationScanner.java
@@ -12,6 +12,7 @@ import org.jboss.jandex.Type;
 
 import io.smallrye.openapi.api.OpenApiConfig;
 import io.smallrye.openapi.api.constants.OpenApiConstants;
+import io.smallrye.openapi.runtime.util.TypeUtil;
 
 /**
  * Abstract base class for annotation scanners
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractAnnotationScanner.java
```diff
@@ -104,18 +105,22 @@ private static boolean profileIncluded(OpenApiConfig config, Set<String> profile
     }
 
     public String[] getDefaultConsumes(AnnotationScannerContext context, MethodInfo methodInfo) {
-        if (methodInfo.returnType().kind().equals(Type.Kind.PRIMITIVE)
-                || methodInfo.returnType().name().equals(DotName.createSimple(String.class))) {
+        if (isPrimimive(methodInfo.returnType())) {
```
MikeEdgar, line 108: I didn't noticed this last time, but shouldn't it be looking at one of the method's args to determine the "consumes" type?

phillip-kruger: Good point. I'll have to check with the RESTEasy maintainers as to what JAX-RS does, as we are trying to align the defaults there. Can we do that as a separate effort ?

MikeEdgar: Yeah, separate effort is fine with me. I'd suggest leaving out the ability to configure primitive consumes for now though since it definitely won't work based on the return type. Wdyt?

MikeEdgar (approved):
LGTM :+1:

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractAnnotationScanner.java
phillip-kruger: As discussed - remove consumes from this text/plain check. consumes default stays json as it was.

PR #1603, opened 2023-10-07
Bring your own `ObjectWriter` to serialize the OpenAPI model

### LINKED ISSUE #1600: Quarkus: Custom ObjectMapper
cc @phillip-kruger cc @MikeEdgar 

Phillip asked me to open this from a Quarkus Zulip conversation.   Let me explain what is happening.

1. I have a JAR of already generated POJO's from OpenAPI Generator I am re-using.  It has fields declared like this.

```java
private JsonNullable<String> directorySize = JsonNullable.<String>undefined();
```

2. When I generate using Microprofile automatic generation in Quarkus it generates the spec like this.

```json
 "directorySize" : {
            "$ref" : "#/components/schemas/JsonNullableString"
},

"JsonNullableString" : {
        "type" : "object",
        "properties" : {
          "value" : {
            "type" : "string"
          },
          "isPresent" : {
            "type" : "boolean"
          },
          "present" : {
            "type" : "boolean"
          }
        }
},
```

3. What I expected was my custom ObjectMapper which is produced uses the `JsonNullable` module to decode these.

```java
import io.quarkus.jackson.ObjectMapperCustomizer;
import jakarta.inject.Singleton;
import org.openapitools.jackson.nullable.JsonNullableModule;

@Singleton
public class JacksonModuleCustomizer implements ObjectMapperCustomizer {

    public void customize(ObjectMapper mapper) {
        mapper.registerModule(new JsonNullableModule());
    }
}
```

4. I was expecting this to be used and it to deserialize properly to:

```json
"directorySize" : {
       "type" : "string"
},
```

5. However https://github.com/smallrye/smallrye-open-api/blob/main/core/src/main/java/io/smallrye/openapi/runtime/io/OpenApiSerializer.java#L41
Smallrye OpenApi always creates a `new ObjectMapper()` for each schema that should be written to disk.
This new ObjectMapper does not take any settings from the Quarkus ObjectMapper into account.

gastaldi:
- Relates to #1600

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/OpenApiSerializer.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/OpenApiSerializer.java
@@ -6,6 +6,7 @@ import org.eclipse.microprofile.openapi.models.OpenAPI;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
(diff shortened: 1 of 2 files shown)


MikeEdgar:
Unfortunately I don't think this will solve the issue. The main problem is that the scanner is introspecting a POJO with a field of type `JsonNullable` and it doesn't know how RestEasy will serialize that with a customized `ObjectMapper`. The scanner is building the OpenAPI model based on the Jandex types and the serialization of that model using a one-off `ObjectMapper` is separate from the RestEasy serialization of runtime instances of it.

I think a solution will need to either (1) make the annotation scanner aware of how a `JsonNullable<T>` should look or (2) make it something a user can configure. 

For (1), it may just be that `JsonNullable` behaves like an `Optional`, which we already support.

For (2), we already have the `mp.openapi.schema.` config prefix, but I don't think it would work correctly with a generic type argument - that could be changed perhaps.

gastaldi:
@MikeEdgar the solution I proposed here is to reuse an ObjectMapper that already has the required modules installed. The next step after this is released would be to change in Quarkus to pass the managed ObjectMapper that in @melloware's scenario is the one to be used.

Feel free to close this PR if that doesn't make sense

MikeEdgar (approved):
@gastaldi I think it makes sense to have a way for Quarkus to pass its ObjectMapper like this, so let's include this PR. I'll unlink the issue though because at the point the OpenAPI model is being serialized, it already contains the incorrect schema model introspected from the Jandex structure. In other words, the `JsonNullable` isn't in the model for serialization at this stage so having the Jackson module present for serialization won't impact the output.

PR #1604, opened 2023-10-08
Maven plugin: index output directory instead of project artifact, logs #1538 (2.3.x backport)

tuffnatty:

PR #1605, opened 2023-10-09
maven-plugin: Document JDK 1.8 compatibility hack

tuffnatty:
As discussed in #1604, JDK 1.8 is supported when a couple of dependency versions are overridden.


MikeEdgar (approved):
Looks good, thank you @tuffnatty !

PR #1606, opened 2023-10-09
Expose Swagger configuration "tryItOutEnabled" property

brunolellis:
This feature exposes Swagger `tryItOutEnabled` configuration property ([doc](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/))

### DIFF
--- a/ui/open-api-ui/src/test/java/io/smallrye/openapi/ui/IndexCreatorTest.java
+++ b/ui/open-api-ui/src/test/java/io/smallrye/openapi/ui/IndexCreatorTest.java
@@ -32,7 +32,7 @@ class IndexCreatorTest {
         assertTrue(s.contains("dom_id: '#swagger-ui',"));
         assertTrue(s.contains("deepLinking: true,"));
         assertFalse(s.contains("queryConfigEnabled"));
-
+        assertFalse(s.contains("tryItOutEnabled"));
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


MikeEdgar (approved):
LGTM, thanks @brunolellis . 

@phillip-kruger , I'll leave for you to merge in case you spot anything I didn't with the UI.

ui/open-api-ui/src/test/java/io/smallrye/openapi/ui/IndexCreatorTest.java
```diff
@@ -240,4 +240,17 @@ void testCreateWithSyntaxHighlightObject() throws IOException {
 
         assertTrue(s.contains("syntaxHighlight: { activated: true, theme: \"monokai\" },"));
     }
+
+    @Test
+    void testCreateWithTryItOutEnabledBoolean() throws IOException {
```
phillip-kruger, line 245: The default for `tryItOut` is enabled (see https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/) so I would rather test `false` or `disabled`.

phillip-kruger (changes requested):

ui/open-api-ui/src/test/java/io/smallrye/openapi/ui/IndexCreatorTest.java
phillip-kruger: The test only test that the value is set, so maybe this is actually fine.

phillip-kruger (approved):

PR #1609, opened 2023-10-11
Support configuration of schemas for complex types

### LINKED ISSUE #1600: Quarkus: Custom ObjectMapper

MikeEdgar:
Fixes #1600 

Support more complex types (besides simply a fully-qualified class name required by the MP TCK) when using the `mp.openapi.schema.` configuration prefix.

For example, to support a field of type `JsonNullable<String>` mentioned in the issue, the following could be configured:

```
mp.openapi.schema.org.openapitools.jackson.nullable.JsonNullable<java.lang.String>={ \
  "name": "JsonNullableString", \
  "type": "string", \
  "nullable": true, \
  "description": "A string that may be nullable!" \
}
```

This configuration would result in all usages of `JsonNullable<String>` being set to a `$ref` of `#/components/schemas/JsonNullableString`.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/ScannerLogging.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/ScannerLogging.java
@@ -5,6 +5,7 @@ import org.jboss.jandex.DotName;
 import org.jboss.jandex.Type;
 import org.jboss.logging.BasicLogger;
 import org.jboss.logging.Logger;
+import org.jboss.logging.annotations.Cause;
 import org.jboss.logging.annotations.LogMessage;
 import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageLogger;
@@ -36,5 +37,9 @@ interface ScannerLogging extends BasicLogger {
     @LogMessage(level = Logger.Level.WARN)
     @Message(id = 4005, value = "Could not find schema class in index: %s")
     void schemaTypeNotFound(DotName className);
+
+    @LogMessage(level = Logger.Level.WARN)
+    @Message(id = 4006, value = "Configured schema type %s is not a valid type signature")
+    void configSchemaTypeInvalid(String typeSignature, @Cause Throwable cause);
 
 }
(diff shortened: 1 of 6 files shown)


MikeEdgar:
PR needs some cleanup and a few more tests, but generally done. 

@melloware, please comment if you do not think this would cover your use case.

MikeEdgar:
That's right. New-lines are only to make it easier to read and aren't required, just FYI

melloware (approved):

phillip-kruger (approved):

PR #1615, opened 2023-10-18
More changes to default content type

phillip-kruger:
As part of https://github.com/quarkusio/quarkus/pull/36514

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractAnnotationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractAnnotationScanner.java
@@ -1,5 +1,11 @@
 package io.smallrye.openapi.runtime.scanner.spi;
 
+import java.io.File;
+import java.io.InputStream;
+import java.io.Reader;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -12,6 +18,7 @@ import org.jboss.jandex.Type;
 
 import io.smallrye.openapi.api.OpenApiConfig;
 import io.smallrye.openapi.api.constants.OpenApiConstants;
+import io.smallrye.openapi.runtime.scanner.ResourceParameters;
 import io.smallrye.openapi.runtime.util.TypeUtil;
 
 /**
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractAnnotationScanner.java
```diff
@@ -116,8 +116,19 @@ public String[] getDefaultProduces(AnnotationScannerContext context, MethodInfo
...
+        if (type != null) {
+            return type.kind().equals(Type.Kind.PRIMITIVE)
+                    || type.name().equals(DotName.createSimple(String.class))
+                    || type.name().equals(DotName.createSimple(Integer.class))
+                    || type.name().equals(DotName.createSimple(Short.class))
+                    || type.name().equals(DotName.createSimple(Long.class))
+                    || type.name().equals(DotName.createSimple(Float.class))
+                    || type.name().equals(DotName.createSimple(Double.class))
+                    || type.name().equals(DotName.createSimple(Boolean.class))
+                    || type.name().equals(DotName.createSimple(Character.class))
```
MikeEdgar, line 128: These could be in a `Set<DotName>`

MikeEdgar (approved):

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractAnnotationScanner.java
phillip-kruger: Yes true ! I'll adjust :) Thanks

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
```diff
@@ -1015,6 +1016,18 @@ default String[] getConsumes(final AnnotationScannerContext context) {
...
+    default String[] getConsumesForRequestBody(final AnnotationScannerContext context) {
+        String[] currentConsumes = context.getCurrentConsumes();
+        if (currentConsumes == null || currentConsumes.length == 0) {
+            currentConsumes = context.getDefaultConsumes();
+        }
+        return currentConsumes;
+    }
+
+    default boolean isRequestBody(MethodParameterInfo mip) {
+        return mip.annotations().isEmpty();
```
MikeEdgar, line 1028: This could be a problem for cases like a request body annotated with the `@RequestBody`, `@Valid`, etc. I think the Spring override may also support `@Valid` and possibly other or custom annotations.

It may be safer to do something like what the request body scanner is doing: 
https://github.com/smallrye/smallrye-open-api/blob/6d4eecfa2133fd8741f79e488bd33b332998a8e8/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java#L969-L972

Of course it won't be exactly the same unless the processing order is changed and the request body scanning is first.

MikeEdgar (commented):
I commented on the type lookup that I think could be a source of bug reports as-is.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
phillip-kruger: `@RequestBody` is JaxRS right ? I think the JaxRS scanner override that default method

MikeEdgar: That's the MP OpenAPI one. I think there could be some number of annotations on a request body though.

phillip-kruger: Ok, I am with you, some might be better to exclude the others (If NOT `@PathParam`, `@QueryParam` etc ?

phillip-kruger:
Thanks @MikeEdgar ! Now getting the type from the method you suggested. Lets merge and release

MikeEdgar (approved):
LGTM. I wonder if we can get away with only calling `getResourceParameters` once in the JAX-RS scanner.

PR #1619, opened 2023-10-20
Improve `@Tag` support for JAX-RS sub-resources

### LINKED ISSUE #1618: Subresource Paths are not grouped under Resource Tags
Hello!

We are using the following Pattern in our Application:

```java
@Tag("My Tag")
@Path("/myparent")
@ApplicationScoped
class MyResource {
  @Inject MySubResource r;
  @Path("/mysub")
  public MySubResource getMySubResource() { return r; }
}
```

```java
@ApplicationScoped
class MySubResource {
  @GET
  @Path("/hello")
  @Operation(summary = "myop")
  public String getHello() { return "hi"; }
}
```

So we have just a single Endpoint '/myparent/mysub/hello'.
I would expect the EP to have the "My Tag"-Tag but instead it has none.

I suspect the Tag is missing because the Tags are not forwarded to the function that processes the subresources:
https://github.com/smallrye/smallrye-open-api/blob/2f9273551c90825f50700f2a6e9290e25886e364/extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsAnnotationScanner.java#L294

Is that the intended behavior?

MikeEdgar:
Fixes #1618 

`@Tag` order of priority:

1. Directly on resource method
2. On nearest subresource producer method (when operation is in a subresource)
3. On nearest subresource class (when operation is in a subresource)
4. On root resource class

### DIFF
--- a/extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsAnnotationScanner.java
+++ b/extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsAnnotationScanner.java
@@ -227,7 +227,9 @@ public class JaxRsAnnotationScanner extends AbstractAnnotationScanner {
         for (ClassInfo resourceClass : resourceClasses) {
             TypeResolver resolver = TypeResolver.forClass(context, resourceClass, null);
             context.getResolverStack().push(resolver);
-            processResourceClass(context, openApi, resourceClass, null);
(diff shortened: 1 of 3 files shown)


MikeEdgar:
FYI @jbuvv, @gquintana .

gquintana:
@MikeEdgar Awesome! Thanks a lot

phillip-kruger (approved):

PR #1628, opened 2023-10-26
Remove test dependency on commons-io

MikeEdgar:
Replaces #1627 

One less dependency to think about for a simple file read


phillip-kruger (approved):
Nice ! I like PRs that remove dependencies :)

PR #1636, opened 2023-11-04
Add reproducer test self-referencing (#1565)

pe-st:
Additional test for #1565 ; the issue was fixed already by PR #1575 (version 3.6.0).

See also discussion thread at https://github.com/smallrye/smallrye-open-api/issues/1565#issuecomment-1787405103


MikeEdgar (approved):
Thanks for the PR @pe-st

PR #1655, opened 2023-11-22
Preserve record accessor method names, update CI for Java 17/21

### LINKED ISSUE #1631: Duplicate properties when (not) using bean convention
Given the following record (the same issue happens with classes)

```java
	record NotABean(boolean isNotABean, String getReal) {}
```

the following schema is generated:

```yaml
    NotABean:
      type: object
      properties:
        isNotABean:
          type: boolean
        getReal:
          type: string
        notABean:
          type: boolean
        real:
          type: string
```

Which is of course undesirable. As we don't want anything to be interpreted according to the "Java Beans Convention": Can this be configured somehow?

MikeEdgar:
Fixes #1631 

- Update CI to use Java 17 + 21 build matrix (build release w/Java 17, run TCK w/Java 11)
- Bumps Gradle to 8.5 (to support Java 21)


phillip-kruger (approved):

MikeEdgar:
I hope to get back to this later this week. The Quarkus Maven plugin build mojo doesn't support toolchains, so I'm looking at alternate approaches to getting this to work nicely with the CI.

MikeEdgar:
@radcortez , @phillip-kruger - with both Quarkus and Wildfly moving to JDK 17 very soon, are there any qualms about dropping JDK 11 (and adding 21) from the CI here? We can continue to create JDK 8 compatible artifacts, but using JDK 17 as the minimum in CI would help avoid doing extra gymnastics to get the embedded test application building in JDK <17.

MikeEdgar:
Finally got this working with a Java 17 + 21 build matrix. @phillip-kruger can you please take another look before I merge this?

phillip-kruger (approved):
LGTM

PR #1684, opened 2023-12-12
Annotation composition and empty array type determination

### LINKED ISSUE #1153: [feat] Annotation composition (for @Consumes)
Could you add support for annotation composition?
Example:
```java
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.METHOD)
@Consumes({"image/jpeg", "image/png"})
public @interface ConsumesImages {
}

@Path("/hello")
public class GreetingsResource {

    @GET
    @ConsumesImages
    @Produces(MediaType.TEXT_PLAIN)
    public String hello() {
        return "Hello RESTEasy";
    }
}
```
In this example OpenAPI falls back to `Content-Type: application/json` instead of a dropdown with values `image/jpeg` and `image/png`. Haven't tested this with other annotations (only `@Consumes`).

Btw. an issue template would be nice.

### LINKED ISSUE #1682: Inherited Annotations support
I'm migrating from the swagger-maven-plugin to smallrye-open-api, but it looks like there is a feature missing, Inherited Annotations.

I have an annotation like:
```java
@APIResponse(responseCode = "400", description = "Bad Request",
    content = {@Content(
        mediaType = "application/json",
        schema = @Schema(implementation = ErrorResponse.class))})
@APIResponse(responseCode = "401", description = "Unauthorized",
    content = {@Content(
        mediaType = "application/json",
        schema = @Schema(implementation = ErrorResponse.class))})
@APIResponse(responseCode = "403", description = "Forbidden",
    content = {@Content(
        mediaType = "application/json",
        schema = @Schema(implementation = ErrorResponse.class))})
@APIResponse(responseCode = "500", description = "Internal Server Error",
    content = {@Content(
        mediaType = "application/json",
        schema = @Schema(implementation = ErrorResponse.class))})
@Target({ElementType.METHOD})
@Retention(RetentionPolicy.RUNTIME)
public @interface CommonApiResponses {}
```

Then use it on the REST API method:
```java
  @APIResponse(responseCode = "200", description = "OK",
      content = {@Content(
          mediaType = MediaType.APPLICATION_JSON,
          schema = @Schema(type = SchemaType.OBJECT))})
  @Tag(name = "mytag")
  @Operation(summary = "Get something", description = """
      My description...
      """)
  @CommonApiResponses
  @GET
  public MyObject getResponse(@PathParam String publisher, @PathParam String name) {...}
```

But the openapi.yaml output does not contain the CommonApiResponses.

MikeEdgar:
This change allows for users to provide their own annotations composed with any annotation supported by SmallRye OpenAPI. The custom annotations that carry otherwise supported annotations may only occur in the locations where the supported annotations are allowed currently.

```java
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.METHOD)
@Consumes({"image/jpeg", "image/png"})
@Produces(MediaType.TEXT_PLAIN)
public @interface ConsumesImagesAndProducesText {
}

@Path("/images")
public class ImagesResource {
    @POST
    @ConsumesImagesAndProducesText   
    public String save(InputStream image) {
        return "Hello World :-)";
    }
}
```

To avoid checking a large number of annotations commonly present, a default set of packages is excluded from the check for composition. Users may override the packages excluded using the new `mp.openapi.extensions.smallrye.scan.composition.exclude.packages` configuration property (unlikely, but it's available anyway). The default packages excluded are `java.*`, `jakarta.*`, `kotlin.*`, `com.fasterxml.jackson.*`, `io.quarkus.*`, `org.eclipse.microprofile.openapi.*`, and `org.jetbrains.annotations.*`.

Includes:
* big refactoring of `Annotations` class which now must be accessed via `AnnotationScannerContext#annotations`.
* move test cases from core and JAX-RS extension to the testsuite/data application
* update testsuite/data to depend on core and JAX-RS extension to ensure it builds with the latest version of the annotation scanner
* New testsuite/coverage module to handle test coverage report aggregation. Previously this was handled by testsuite/extra

Closes #1153 
Closes #1682


phillip-kruger (approved):
Nice

phillip-kruger (approved):
LGTM

PR #1706, opened 2024-01-02
Fix false-positive duplicate operation-ID w/ javax+jakarta annotations

### LINKED ISSUE #1688: Regression: OpenAPI scanner reports duplicate operation id for exactly the same methods of the same class
I see issue #1379 happening (again) in [our project](https://github.com/projectnessie/nessie), via Quarkus 3.6.3, using smallrye-open-api-core 3.7.0:

```
SROAP07903: Duplicate operationId: %s produced by Class: %s, Method: %s and Class: %s, Method: %s
SROAP07903: Duplicate operationId: %s produced by Class: %s, Method: %s and Class: %s, Method: %s
SROAP07903: Duplicate operationId: %s produced by Class: %s, Method: %s and Class: %s, Method: %s
SROAP07903: Duplicate operationId: %s produced by Class: %s, Method: %s and Class: %s, Method: %s
SROAP07903: Duplicate operationId: %s produced by Class: %s, Method: %s and Class: %s, Method: %s
SROAP07903: Duplicate operationId: %s produced by Class: %s, Method: %s and Class: %s, Method: %s
SROAP07903: Duplicate operationId: %s produced by Class: %s, Method: %s and Class: %s, Method: %s
```
(Unfortunately, logging via Gradle eats up all the message formatting, but that's a different issue.)

Our REST classes basically all use the same inheritance model:
1. The implementation class [example](https://github.com/projectnessie/nessie/blob/64e5b458a877c5470ed56fa6cf30e3ebe3edaa5c/servers/services/src/main/java/org/projectnessie/services/impl/DiffApiImpl.java), which implements a ...
2. REST/OpenAPI annotated interface [example](https://github.com/projectnessie/nessie/blob/64e5b458a877c5470ed56fa6cf30e3ebe3edaa5c/api/model/src/main/java/org/projectnessie/api/v1/http/HttpDiffApi.java), which extends a ...
3. base interface (w/o) the REST/OpenAPI annotations ([example](https://github.com/projectnessie/nessie/blob/64e5b458a877c5470ed56fa6cf30e3ebe3edaa5c/api/model/src/main/java/org/projectnessie/api/v1/DiffApi.java)).

Looking into this issue in a debugger, I see in [`AnnotationScanner.processOperation`](https://github.com/smallrye/smallrye-open-api/blob/89b367a66a7cdca0956f701a0eaafbb02282cc3b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java#L353) that `conflictingMethod` points to the exact same (not equal) object as `method` (so: `method == conflictingMethod`). This means, that the `Duplicate operationId` warning is a false-positive w/o a negative effect (for us).

Looking into this a bit deeper, in `AnnotationScanner.getResourceMethods()`, I see the following behavior:
1. It looks up the methods via `classInfo`, getting one in this example:
image
2. It then includes the interfaces, adding the base methods as well (same signature!):
image

IIUC to fix this issue, it would be the better approach to change `AnnotationScanner.getResourceMethods()` to be aware of overrides and not "blindly" add those (again). WDYT?

snazy:
Fixes #1688

### DIFF
--- a/extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/JaxRsAnnotationScannerTest.java
+++ b/extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/JaxRsAnnotationScannerTest.java
@@ -4,6 +4,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.UUID;
 
 import jakarta.ws.rs.GET;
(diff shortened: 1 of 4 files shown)


extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/JaxRsAnnotationScannerTest.java
```diff
@@ -61,6 +62,31 @@ void testJakartaHiddenOperationNotPresent() throws IOException, JSONException {
...
+
+        Map<String, String> cfg = new HashMap<>();
+        // fail hard to trigger the failure in this test
+        cfg.put(OpenApiConstants.DUPLICATE_OPERATION_ID_BEHAVIOR, OpenApiConfig.DuplicateOperationIdBehavior.FAIL.toString());
+        // method-strategy needed for this test
+        cfg.put(OpenApiConstants.OPERATION_ID_STRAGEGY, OpenApiConfig.OperationIdStrategy.METHOD.toString());
+        OpenApiConfig config = dynamicConfig(cfg);
+
+        OpenApiAnnotationScanner s = new OpenApiAnnotationScanner(config, index);
+        OpenAPI result = s.scan("JAX-RS");
```
MikeEdgar, line 86: Maybe something like this will make explicit that nothing will be thrown even with the `FAIL` DuplicateOperationIdBehavior.
```suggestion
        OpenAPI result = assertDoesNotThrow(() -> s.scan("JAX-RS"));
```

MikeEdgar (approved):
Looks good, but maybe add an assertion to the test method. That will also quiet Sonar which flags it as a "bug" :smile:

MikeEdgar (approved):
Thanks @snazy

snazy:
YW :)

PR #1736, opened 2024-02-01
Annotations#getComposedAnnotation should only process runtime annotations

manovotn:
This came up in a Quarkus issue - https://github.com/quarkusio/quarkus/issues/38506
In that case Lombok could add a non-runtime annotation `@Generated` which you'd then attempt to index, leading to a WARN.
With this change I no longer saw the WARN logged in the Quarkus reproducer.

I am rather unfamiliar with open-api so feel free to take this and tweak whichever way it makes sense :)


Ladicek:
I guess that a few more places in `Annotations` could use the same change, but I'm no expert on this codebase either.

MikeEdgar:
Is the issue that only runtime visible annotations are in the index?

core/src/main/java/io/smallrye/openapi/runtime/util/Annotations.java
```diff
@@ -117,6 +117,7 @@ private Stream<AnnotationInstance> getComposedAnnotation(Collection<AnnotationIn
 
         return declaredAnnotations
                 .stream()
+                .filter(annotationInstance -> annotationInstance.runtimeVisible())
```
MikeEdgar, line 120:
```suggestion
                .filter(AnnotationInstance::runtimeVisible)
```

Ladicek:
Since Jandex 3.0, class-retained annotations are also in the index -- but it's normal and legal for annotations to actually not exist on the classpath. (Regardless of their retention. But it is especially common for class-retained annotations to be missing.)

manovotn:
> I guess that a few more places in `Annotations` could use the same change, but I'm no expert on this codebase either.

Yes, I was thinking the same but I am not familiar with this codebase.

> Is the issue that only runtime visible annotations are in the index?

Hm good question; I was rather thinking along the lines of other than non-runtime annotations not being very useful for inspection anyway. I know we had to take similar precautions in Arc/Quarkus several times already as other than runtime retention annotations were missing in some cases.

MikeEdgar:
This section of the code is dealing with nested/composed annotations, but there isn't any attempt to load the classes - only looking in the index.

> I was rather thinking along the lines of other than non-runtime annotations not being very useful for inspection anyway

We actually requested the runtime invisible feature in Jandex so that the scanner could find certain Kotlin annotations to influence the OpenAPI generated :smile:

MikeEdgar (approved):

MikeEdgar:
I'll go ahead and merge this and do a release for 3.9.0 as well.

manovotn:
> This section of the code is dealing with nested/composed annotations, but there isn't any attempt to load the classes - only looking in the index.

There's an automatic additional indexing if the class you're looking for is missing.
See for instance this - https://github.com/quarkusio/quarkus/blob/main/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanArchives.java#L128-L134

MikeEdgar:
Well that's good to know - TIL! There are other locations where we _do_ check the classpath (non annotations) so I suppose that code is never executing in the case of Quarkus/ArC.

> the several layer of IndexView implementations from SmallRye OpenAPI

Yea, that should be collapsed into one that does the filtering required by the spec.

PR #1759, opened 2024-02-29
Refactor package `io.smallrye.openapi.runtime.io` and subpackages

MikeEdgar:
This PR refactors the I/O functionality from a collection of static methods to a structured object tree. The aim is to decouple the underlying JSON library in use from the construction and serialization of the OpenAPI model and also to get more code reuse between commonly structured models.


MikeEdgar:
I think this is in a pretty good state at this point. Following the merge of this I will put up a PR for #1762

phillip-kruger (approved):

PR #1774, opened 2024-03-14
Add new SmallRyeOpenAPI public builder API

### LINKED ISSUE #1762: Introduce more usable public API
Currently, platforms integrating with smallrye-open-api make use of some combination of at least the following classes:

- `io.smallrye.openapi.api.OpenApiConfig`
- `io.smallrye.openapi.api.OpenApiDocument`
- `io.smallrye.openapi.runtime.OpenApiProcessor`
- `io.smallrye.openapi.runtime.OpenApiStaticFile`
- `io.smallrye.openapi.runtime.io.OpenApiParser`
- `io.smallrye.openapi.runtime.io.OpenApiSerializer`
- `io.smallrye.openapi.runtime.scanner.AnnotationScannerExtension`
- `io.smallrye.openapi.runtime.scanner.spi.AnnotationScanner`

Many of the classes provide overloaded static methods that makes it hard to make modifications without studying each client.

We should have a single unified public "builder" style API that users access to generate OpenAPI models and de-/serialize them to/from YAML or JSON. It is a goal to limit exposing internal smallrye-open-api types in this new API.

MikeEdgar:
Closes #1762


Azquelt:
I'll have a read through this, but for reference the place where we call smallrye OpenAPI in OpenLiberty is here: https://github.com/OpenLiberty/open-liberty/blob/a0f4167ee6dfad2d80a420d32ea410d4df239542/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/ApplicationProcessor.java#L301

phillip-kruger (approved):
Let get this in and roll forward. Should we do a major/minor bump ? Maybe branch the main first ?

PR #1793, opened 2024-04-09
OAS 3.1 changes for schema model

Azquelt:
Implements the spec changes in eclipse/microprofile-open-api#598

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/models/MapBasedModelImpl.java
+++ b/core/src/main/java/io/smallrye/openapi/api/models/MapBasedModelImpl.java
@@ -106,9 +106,12 @@ public abstract class MapBasedModelImpl implements ModelImpl {
     }
 
     protected <T> void addToListProperty(String propertyName, T value) {
-        List<T> list = (List<T>) data.get(propertyName);
         if (value != null) {
-            if (list == null) {
+            Object existing = data.get(propertyName);
+            List<T> list;
+            if (existing instanceof List) {
+                list = (List<T>) existing;
+            } else {
                 list = new ArrayList<>();
                 data.put(propertyName, list);
             }
@@ -117,8 +120,11 @@ public abstract class MapBasedModelImpl implements ModelImpl {
     }
 
     protected <T> void removeFromListProperty(String propertyName, T toRemove) {
-        List<T> list = (List<T>) data.get(propertyName);
-        ModelUtil.remove(list, toRemove);
+        Object existing = data.get(propertyName);
+        if (existing instanceof List) {
+            List<T> list = (List<T>) existing;
+            ModelUtil.remove(list, toRemove);
+        }
     }
 
     protected <T> void setMapProperty(String propertyName, Map<String, T> value) {
@@ -141,8 +147,11 @@ public abstract class MapBasedModelImpl implements ModelImpl {
 
     protected <T> void addToMapProperty(String propertyName, String key, T value) {
         if (value != null) {
-            Map<String, T> map = (Map<String, T>) data.get(propertyName);
-            if (map == null) {
+            Object existing = data.get(propertyName);
+            Map<String, T> map;
+            if (existing instanceof Map) {
+                map = (Map<String, T>) existing;
+            } else {
                 map = new HashMap<>();
                 data.put(propertyName, map);
             }
@@ -151,8 +160,11 @@ public abstract class MapBasedModelImpl implements ModelImpl {
     }
 
     protected <T> void removeFromMapProperty(String propertyName, String key) {
-        Map<String, T> map = (Map<String, T>) data.get(propertyName);
-        ModelUtil.remove(map, key);
+        Object existing = data.get(propertyName);
+        if (existing instanceof Map) {
+            Map<String, T> map = (Map<String, T>) existing;
+            ModelUtil.remove(map, key);
+        }
     }
 
     /**
--- a/core/src/main/java/io/smallrye/openapi/api/models/MapBasedModelImpl.java
+++ b/core/src/main/java/io/smallrye/openapi/api/models/MapBasedModelImpl.java
@@ -1,0 +1,167 @@
+package io.smallrye.openapi.api.models;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+
+import io.smallrye.openapi.api.util.MergeUtil;
+import io.smallrye.openapi.runtime.util.ModelUtil;
+
+/**
+ * Base implementation for extensible model objects which wrap a map of properties
+ */
+public abstract class MapBasedModelImpl implements ModelImpl {
+
+    protected final HashMap<String, Object> data = new HashMap<>();
+
+    protected MapBasedModelImpl() {
+    }
+
+    /**
+     * Merge all properties from another map-based model object into this one
+     * <p>
+     * Usually this method will return {@code this}, but it may return {@code other} or a new object.
+     *
+     * @param other the other map-based model object
+     * @return the merged object
+     */
+    public MapBasedModelImpl mergeFrom(MapBasedModelImpl other) {
+        mergeMap(data, other.data, getNonMergableCollections());
+        return this;
+    }
+
+    public Map<String, Object> getDataMap() {
+        return data;
+    }
+
+    private static <T> void mergeMap(Map<String, T> into, Map<String, T> from, Set<String> nonMergableNames) {
+        for (Entry<String, T> entry : from.entrySet()) {
+            String name = entry.getKey();
+            T value = entry.getValue();
+            T oldValue = into.get(entry.getKey());
+            if (oldValue == null || oldValue.getClass() != value.getClass()) {
+                into.put(name, value);
+            } else {
+                if (oldValue instanceof Map && !nonMergableNames.contains(name)) {
+                    mergeMap((Map<String, Object>) oldValue, (Map<String, Object>) value, nonMergableNames);
+                } else if (oldValue instanceof List && !nonMergableNames.contains(name)) {
+                    mergeList((List<Object>) oldValue, (List<Object>) value);
+                } else if (oldValue instanceof ModelImpl) {
+                    into.put(name, MergeUtil.mergeObjects(oldValue, value));
+                } else {
+                    into.put(name, value);
+                }
+            }
+        }
+    }
+
+    private static <T> void mergeList(List<T> oldValue, List<T> value) {
+        Set<T> contents = new HashSet<>(oldValue);
+        for (T element : value) {
+            if (!contents.contains(element)) {
+                oldValue.add(element);
+            }
+        }
+    }
+
+    protected <T> void setProperty(String propertyName, T value) {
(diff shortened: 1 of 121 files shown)


Azquelt:
Build failures are expected because this depends on spec changes.

Locally I have tests passing up to this point:
```
SmallRye: OpenAPI Test Data ........................ FAILURE [ 10.751 s]
```

With the failure:
```
[ERROR] 	[error]: Build step io.quarkus.smallrye.openapi.deployment.SmallRyeOpenApiProcessor#build threw an exception: java.lang.NoSuchFieldError: org/eclipse/microprofile/openapi/models/media/Schema$SchemaType.NULL
[ERROR] 	at io.smallrye.openapi.api.models.media.SchemaImpl.setNullable(SchemaImpl.java:330)
[ERROR] 	at io.smallrye.openapi.runtime.io.schema.SchemaFactory.readSchema(SchemaFactory.java:188)
[ERROR] 	at io.smallrye.openapi.runtime.io.schema.SchemaFactory.readSchema(SchemaFactory.java:119)
[ERROR] 	at io.smallrye.openapi.runtime.io.schema.SchemaFactory.readSchema(SchemaFactory.java:94)
[ERROR] 	at io.smallrye.openapi.runtime.scanner.OpenApiDataObjectScanner.readKlass(OpenApiDataObjectScanner.java:376)
[ERROR] 	at io.smallrye.openapi.runtime.scanner.OpenApiDataObjectScanner.depthFirstGraphSearch(OpenApiDataObjectScanner.java:251)
```
which looks like this part of the build is running with the built smallrye-open-api, but using an older version of the spec APIs without `SchemaType.NULL`.

Azquelt:
Found the issue and updated the dependency for the testdata project and now my local build runs cleanly.

core/src/main/java/io/smallrye/openapi/api/models/JsonWrappingImpl.java
```diff
@@ -0,0 +1,167 @@
...
+        if (result instanceof Map) {
+            return Collections.unmodifiableMap((Map<String, T>) data.get(propertyName));
+        } else {
+            return null;
+        }
+    }
+
+    protected <T> void addToMapProperty(String propertyName, String key, T value) {
+        if (value != null) {
+            Map<String, T> map = (Map<String, T>) data.get(propertyName);
```
Azquelt, line 144: The add and remove methods for maps and lists risk a `ClassCastException` if the data stored in the property is not actually a map or a list.

We may want to improve this by:
* providing a better exception
* for add: overwriting the value with a new list or map
* for remove: doing nothing

core/src/main/java/io/smallrye/openapi/api/models/MapBasedModelImpl.java
```diff
@@ -0,0 +1,167 @@
...
+        if (result instanceof Map) {
+            return Collections.unmodifiableMap((Map<String, T>) data.get(propertyName));
+        } else {
+            return null;
+        }
+    }
+
+    protected <T> void addToMapProperty(String propertyName, String key, T value) {
+        if (value != null) {
+            Map<String, T> map = (Map<String, T>) data.get(propertyName);
```
Azquelt, line 144: The add and remove methods for maps and lists risk a `ClassCastException` if the data stored in the property is not actually a map or a list.

We may want to improve this by:
* providing a better exception
* avoiding an exception
  * for add: overwriting the value with a new list or map
  * for remove: doing nothing

core/src/main/java/io/smallrye/openapi/api/models/media/SchemaImpl.java
```diff
@@ -82,61 +131,79 @@ public static void addTypeObserver(Schema observable, Schema observer) {
...
-        clone.externalDocs = copy(clone.externalDocs,
-                () -> MergeUtil.mergeObjects(new ExternalDocumentationImpl(), clone.externalDocs));
-
-        clone.oneOf = copy(clone.oneOf, () -> clone.oneOf
-                .stream()
-                .map(SchemaImpl::copyOf)
-                .collect(Collectors.toList()));
+        if (other == null) {
+            return new SchemaImpl();
+        }
```
Azquelt, line 140: The logic of `copyOf` is very similar to `MapBasedModelImpl.mergeFrom`, we may be able to remove one of them.

MikeEdgar:
@Azquelt I've updated the target branch to `main-4.0`

MikeEdgar:
@Azquelt , the target branch did not have CI enabled, so I've added it. We should get the test jobs running against the SNAPSHOT MP OpenAPI 4.0 build on your next rebase/push.

core/src/main/java/io/smallrye/openapi/api/models/media/SchemaImpl.java
Azquelt: It's slightly different - where `from` has a property but `to` does not, the value from `from` is used directly. When the method returns, both `from` and `to` reference the same object. This matches the behaviour of `MergeUtil.mergeObjects` and means `mergeFrom` is only really suitable if you're going to throw both of the source objects away and just use the result object.

For this reason I recommend keeping both, even though the logic is similar.

core/src/main/java/io/smallrye/openapi/api/models/MapBasedModelImpl.java
Azquelt: I've implemented the second option.

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
```diff
@@ -430,7 +464,7 @@ boolean fieldSpecifiesAnnotation(Schema fieldSchema) {
             Schema::getDeprecated,
             Schema::getDescription,
             Schema::getExample,
-            Schema::getExtensions,
+            //            Schema::getExtensions,
```
Azquelt, line 467: Should this be reverted?

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
```diff
@@ -470,8 +471,22 @@ default void createResponseFromRestMethod(final AnnotationScannerContext context
...
+                        Schema refSchema = new SchemaImpl().ref(schema.getRef());
+                        schema.setRef(null);
+                        if (schema.getAnyOf() == null) {
+                            schema.addAnyOf(refSchema).addAnyOf(nullSchema);
+                        } else {
+                            Schema anyOfSchema = new SchemaImpl().addAnyOf(refSchema).addAnyOf(nullSchema);
+                            schema.addAllOf(anyOfSchema);
+                        }
+                    }
                 }
```
Azquelt, line 490: @MikeEdgar you were correct. The old code here did not consider that `schema` might have `$ref` set. If `$ref` was set, then value for `nullable` set here got discarded when the schema was serialized.

It looked like the intent of the old code was to add `nullable = true` when a resource method returned `Optional<T>` so I made the new code handle this case whether or not the response schema is a reference.

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/AnnotationTargetProcessor.java
Azquelt: I think it should be, but uncommenting this line alone causes test failures in the TCK because `getExtensions()` doesn't currently behave like other getter methods. In particular it will never return `null` and the code that uses this list of getters is explicitly checking for whether they return `null`.

To fix this, I had to update the extensions methods on `SchemaImpl` to behave like other getter/setter sets and differentiate between being `null` and being empty. [`ModelConstructionTest.processExtensible`](https://github.com/eclipse/microprofile-open-api/blob/972a655efc638079c88e2491076d0c179e48230a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java#L1750) from the TCK also expects this behaviour, but the exact way it was testing it wasn't tripped up by our current implementation.

MikeEdgar:
@Azquelt I just fast-forwarded `main-4.0` to the current `main`. There are conflicts, but nothing looks terrible.

Azquelt:
@MikeEdgar Thanks, I'll have a look

Azquelt:
@MikeEdgar I've rebased this on the updated `main-4.0` branch to resolve the conflicts and I think it's ready for review

core/src/main/java/io/smallrye/openapi/api/models/media/SchemaImpl.java
```diff
@@ -891,7 +1078,421 @@ public Schema addOneOf(Schema oneOf) {
...
+        });
+        return extensions.isEmpty() && !explicitSetExtensions ? null : extensions;
+    }
+
+    @Override
+    public Schema addExtension(String name, Object value) {
+        setProperty(name, value);
+        explicitSetExtensions = true;
+        return this;
+    }
```
MikeEdgar, line 1100: The new mutator methods are missing calls to `incrementModCount`. I wonder if it would be better to move `modCount` up to `MapBasedModelImpl` as protected and consolidate usage of `incrementModCount`. Either way is fine with me.

MikeEdgar (commented):
This looks good to me, as expected. I left a few comments, but nothing major.

@phillip-kruger, this is for the MP OpenAPI 4.0 spec update. Please take a pass through if you have a chance.

phillip-kruger:
Let go, we can roll forward. Would this cause a breaking change in Quarkus, or is it backwards compatible ?

MikeEdgar:
> Let go, we can roll forward. Would this cause a breaking change in Quarkus, or is it backwards compatible ?

This and other MP OpenAPI changes are going into the main-4.0 branch for now, so we won't release that until everything is completed and we'll bump the major version at that point.

phillip-kruger:
Just branch of current main into a tag for in case we need to do work on that branch.

core/src/main/java/io/smallrye/openapi/api/models/media/SchemaImpl.java
Azquelt: Good spot, I've pushed up changes moving the `incrementModCount()` calls into the `setProperty` methods on `SchemaImpl`.

I've kept it all in `SchemaImpl` for the moment since we already have additional logic we do before getting and setting properties on a schema.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
```diff
@@ -400,7 +400,8 @@ boolean templateParameterPatternEligible(Parameter param) {
         return Parameter.In.PATH.equals(param.getIn())
                 && !Style.MATRIX.equals(param.getStyle())
                 && param.getSchema() != null
-                && SchemaType.STRING.equals(param.getSchema().getType())
+                && param.getSchema().getType() != null
+                && param.getSchema().getType().contains(SchemaType.STRING)
```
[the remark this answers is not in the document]
Azquelt: Possibly, though I note that only string instances are constrained by `pattern`. If `pattern` was set and type was `["number", "string"]`, any numeric value would be valid, but a string value would only be valid if it conformed to `pattern`.

MikeEdgar (approved):
LGTM, merging to `main-4.0`

PR #1801, opened 2024-04-16
Schema annotations 3.1

Azquelt:
Implement the `@Schema` and `@SchemaProperty` changes from eclipse/microprofile-open-api#601.

I haven't duplicated cases which are already covered by the TCK.


Azquelt:
In e83724a I made some of the scanner tests output the actual JSON document in the event of a failure because it was more helpful to me during development than configuring logging and finding the logs, but that part is not relevant to the actual changes and can be dropped if desired.

Azquelt:
CI will not run cleanly until eclipse/microprofile-open-api#601 is merged but I have built and run this change locally with no failures.

MikeEdgar (approved):
Looks good. Let's leave it open until there's a good run of the TCK against a new MP snapshot.

Azquelt:
@MikeEdgar we now have a good build

Issue #1802, opened 2024-04-17
Enhancement Request: Provide a fixed order for responses


dgf:
We've encountered some challenges while using definitions generated by Quarkus in our CI workflow. Constantly/randomly it changes the order of the responses in the generated OpenAPI JSON and YML files. This has various side effects and requires manual reviews in our processes.

The cause is the generation of the responses section itself. The function `io.smallrye.openapi.runtime.io.response.ResponseWriter#writeAPIResponses` loops over the entry set of responses without any sorting.

We fixed it for now in our env by deploying a SNAPSHOT that sorts the set with `Map.Entry.comparingByKey()`

Its tested and running well already for some dozen builds and we would like to see the stabilization for the export of definitions in the standard.

To do that we/you need to decide what should be the fixed order. For now I'm seeing two different ones in the tests. Some JSON based expectation files like `core/src/test/resources/io/smallrye/openapi/runtime/io/_everything.json` containing the responses in alphabetical order "default", "200", "404" and the some YML assertions like in the test `io.smallrye.openapi.runtime.io.OpenApiParserAndSerializerTest#testParsingBigStaticYamlFile` are reversed: "204", "200" and "404", "200"

What is the expected order? Are there any notes about in the specs or a best practice?

We can easily create a PR, as we have a fix ready (just need to decide if forward or backwards sorted).

phillip-kruger:
Please provide that PR and we can discuss there. Thanks

dgf:
oh dear, I've just discovered that the entire core package has been refactored. I created the fix based on 3.10.0 . So I'll have to start again, but first I'll have to familiarize myself with the new JSON handling.

MikeEdgar:
Hi @dgf , how large were your changes? Even if you can post a link to a git commit here, we can at least discuss the approach before re-basing on the latest main branch.

dgf:
I have rebuilt the changes to the current version. Unfortunately, I can't test this with Quarkus at the moment because the class references have changed. But that should be sufficient as a basis for further exchanges about that topic.

dgf:
Solved on Quarkus itself by using a sorted map, see https://github.com/quarkusio/quarkus/issues/40222

In general, I still think it's a good idea to have a predefined order for responses, but no longer necessary, at least, for our specific case.

PR #1803, opened 2024-04-18
sort responses alphabetically

dgf:


dgf:
refs #1802

phillip-kruger:
@MikeEdgar This seems ok to me. w.d.y.t ?

MikeEdgar:
One thing to think about with doing it this way is that it affects the responses object no matter how it was created. For example, users may set some custom order in a static file or filter today (using a `TreeMap` or some custom insertion-order `LinkedHashMap`) and this overrides that.

I am not sure how likely that is. Generally, I would say that stuff that changes the model would be better placed outside of the serializer, and keep that concerned with just dumping whatever the model already is. 

Maybe the sorting of `APIResponses` keys could be done in the annotation scanner after all the paths are scanned? That would limit the impact to just scanned responses. I think we do something like this for tag names already.

Thoughts?

phillip-kruger:
Mmm, yes. I wonder should all of this not be done in a Filter ? We can provide the filter but let users opt in to use it ?

MikeEdgar:
> Mmm, yes. I wonder should all of this not be done in a Filter ? We can provide the filter but let users opt in to use it ?

Maybe something like the `io.smallrye.openapi.api.util.UnusedSchemaFilter`

dgf:
Just to add some context. 

Our specific issue is caused by the way how Quarkus is adding the default security responses for authenticated endpoints. This results in random outputs of the response entries in generated files, depending on the machine or time of the day or whatever is seeding the loop access of maps in Java.

Quarkus creates a Map of additional responses: https://github.com/quarkusio/quarkus/blob/37fbf8acd65ad2ae1884272d60e8f0408d3f7bb6/extensions/smallrye-openapi/deployment/src/main/java/io/quarkus/smallrye/openapi/deployment/filter/AutoRolesAllowedFilter.java#L148

and merges it with the existing ones: https://github.com/quarkusio/quarkus/blob/37fbf8acd65ad2ae1884272d60e8f0408d3f7bb6/extensions/smallrye-openapi/deployment/src/main/java/io/quarkus/smallrye/openapi/deployment/filter/AutoRolesAllowedFilter.java#L139

Additional note / finding: the order is constant in the generated responses per run, so e.g. for the simple cases all responses are "200" "401" "403" or "200" "403" "401"

phillip-kruger:
Yea this should be easy to fix in the Quarkus auto security filter

MikeEdgar:
Closing for now since it was fixed in Quarkus. If we want to revisit again here, let's consider doing it as a filter or some other approach that users can opt-in to enable, but have it disabled by default.

PR #1818, opened 2024-04-28
Update openapi version string

benjamin-confino:
This change will keep us in line with https://github.com/eclipse/microprofile-open-api/issues/434


MikeEdgar:
@benjamin-confino I think this should target the `main-4.0` branch.

benjamin-confino:
I'm marking this as a draft until #1801 is delivered, otherwise it will cause a merge conflict in that more important PR. When 1801 is in I'll deal with the merging myself.

core/src/main/java/io/smallrye/openapi/runtime/io/schema/SchemaConstant.java
```diff
@@ -22,7 +22,7 @@
 /**
  * Constants related to Schema
  *
- * @see <a href="https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#schemaObject">schemaObject</a>
+ * @see <a href="https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#schemaObject">schemaObject</a>
```
MikeEdgar, line 25:
```suggestion
 * @see <a href="https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#schema-object">schemaObject</a>
```

tools/gradle-plugin/src/test/java/io/smallrye/openapi/gradleplugin/SmallryeOpenApiPluginTest.java
```diff
@@ -71,7 +71,7 @@ void taskPropertiesInheritance() {
         ext.applicationPathDisable.set(true);
         ext.configProperties.set(new File("/foo/bar"));
         ext.encoding.set("ISO-8859-1");
-        ext.openApiVersion.set("3.0.0");
+        ext.openApiVersion.set("3.1.0");
```
MikeEdgar, line 74: It might be good to keep this different from the default. Maybe `3.1.1` or something. Even if it's not valid, we can test that the configuration is actually taking effect.

MikeEdgar (changes requested):
First pass with some comments. For the files removed, were some of them simply not used in the tests?

tools/gradle-plugin/src/test/java/io/smallrye/openapi/gradleplugin/SmallryeOpenApiPluginTest.java
benjamin-confino: If we're going to do a non-valid one, can I suggest something like 999.999.999. That way is unambiguously invalid.

benjamin-confino:
> First pass with some comments. For the files removed, were some of them simply not used in the tests?

Indeed they were not. I did a text search for the filenames and found no reference to them, then I tested without the files and nothing complained.

tools/gradle-plugin/src/test/java/io/smallrye/openapi/gradleplugin/SmallryeOpenApiPluginTest.java
MikeEdgar: That's fine with me, thanks.

tools/gradle-plugin/src/test/java/io/smallrye/openapi/gradleplugin/SmallryeOpenApiPluginTest.java
```diff
@@ -193,7 +193,7 @@ void smokeProject(Path buildDir, boolean withQuarkus, String taskName, String ou
                         "}",
                         "",
                         "smallryeOpenApi {",
-                        "  openApiVersion.set(\"3.0.2\")",
+                        "  openApiVersion.set(\"3.1.0\")",
```
MikeEdgar, line 196:
```suggestion
                        "  openApiVersion.set(\"999.999.999\")", //intentionally invalid
```

tools/gradle-plugin/src/test/java/io/smallrye/openapi/gradleplugin/SmallryeOpenApiPluginTest.java
```diff
@@ -287,7 +287,7 @@ private static void checkGeneratedFiles(Path buildDir, String expectedOutputFile
             JsonNode root = new ObjectMapper().readValue(
                     targetOpenapiDir.resolve("my-openapi-schema-file.json").toUri().toURL(),
                     JsonNode.class);
-            assertThat(root.get("openapi").asText()).isEqualTo("3.0.2");
+            assertThat(root.get("openapi").asText()).isEqualTo("3.1.0");
```
MikeEdgar, line 290:
```suggestion
            assertThat(root.get("openapi").asText()).isEqualTo("999.999.999");
```

core/src/main/java/io/smallrye/openapi/runtime/util/TypeUtil.java
```diff
@@ -110,7 +110,7 @@ public class TypeUtil {
     public static final IndexView jdkIndex;
     private static final Set<DotName> wrapperTypes = new HashSet<>();
 
-    // https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#dataTypeFormat
+    // https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#data-type-format
```
MikeEdgar, line 113: This one didn't change in the spec ¯\_(ツ)_/¯
```suggestion
    // https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#dataTypeFormat
```

MikeEdgar (changes requested):
A few more comments

benjamin-confino:
The latest push was rebaseing on the latest main-4.0

MikeEdgar (approved):

PR #1821, opened 2024-05-03
OASFactoryResolverImpl clinit registry lazy init

franz1981:
According to the quarkus start stop tests (at https://github.com/quarkus-qe/quarkus-startstop/pull/371):

image

https://github.com/quarkusio/quarkus/blob/4eaa586672bbc18a25f3dc82f18201260ee8ce0a/extensions/smallrye-openapi/runtime/src/main/java/io/quarkus/smallrye/openapi/runtime/OpenApiRecorder.java#L98

(which is in the flamegraph, at `io/quarkus/smallrye/openapi/runtime/OpenApiRecorder.classLoaderHack`)

cause SPI loading and allocating `OASFactoryResolverImpl` , causing its registry allocation (with the many class loading events it brings...).

By default just hitting 

```
$ curl http://localhost:8080/q/openapi
```
nor any of the endpoints, I'm not able to get `OASFactoryResolverImpl::createObject` to be used, but please correct me if I have some naive assumption on how this is supposed to work.

WIth this PR the registry allocation/population, would happen lazily, making it lighter and faster application startup if this registry map isn't used.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/spi/OASFactoryResolverImpl.java
+++ b/core/src/main/java/io/smallrye/openapi/spi/OASFactoryResolverImpl.java
@@ -78,43 +78,44 @@ import io.smallrye.openapi.api.models.tags.TagImpl;
  */
 public class OASFactoryResolverImpl extends OASFactoryResolver {
 
-    private static final Map<Class<? extends Constructible>, Supplier<? extends Constructible>> registry = new HashMap<>();
+    private static class OASFactoryResolverRegistry {
 
-    static <K extends Constructible, V extends K> void put(Class<K> key, Supplier<V> value) {
-        registry.put(key, value);
-    }
+        private static final Map<Class<? extends Constructible>, Supplier<? extends Constructible>> REGISTRY;
 
-    static {
-        put(APIResponse.class, APIResponseImpl::new);
-        put(APIResponses.class, APIResponsesImpl::new);
-        put(Callback.class, CallbackImpl::new);
-        put(Components.class, ComponentsImpl::new);
-        put(Contact.class, ContactImpl::new);
-        put(Content.class, ContentImpl::new);
-        put(Discriminator.class, DiscriminatorImpl::new);
-        put(Encoding.class, EncodingImpl::new);
-        put(Example.class, ExampleImpl::new);
-        put(ExternalDocumentation.class, ExternalDocumentationImpl::new);
-        put(Header.class, HeaderImpl::new);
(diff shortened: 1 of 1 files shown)


franz1981:
Regardless, I'm a bit concerned on the lack of concurrency protection while writing/reading on `HashMap`, but is a different problem...

Sanne:
Good idea!

core/src/main/java/io/smallrye/openapi/spi/OASFactoryResolverImpl.java
```diff
@@ -78,43 +78,48 @@
...
-        put(Schema.class, SchemaImpl::new);
-        put(SecurityRequirement.class, SecurityRequirementImpl::new);
-        put(SecurityScheme.class, SecuritySchemeImpl::new);
-        put(Server.class, ServerImpl::new);
-        put(ServerVariable.class, ServerVariableImpl::new);
-        put(Tag.class, TagImpl::new);
-        put(XML.class, XMLImpl::new);
+    static <K extends Constructible, V extends K> void put(Class<K> key, Supplier<V> value) {
+        OASFactoryResolverRegistry.REGISTRY.put(key, value);
     }
```
MikeEdgar, line 124: With your changes this put method can be removed also.

```suggestion
```

MikeEdgar (approved):
Thanks @franz1981 , this looks good. The `put` method previously used to load the map can be taken out now too.

core/src/main/java/io/smallrye/openapi/spi/OASFactoryResolverImpl.java
franz1981: wait a sec, so the `REGISTRY` cannot be modified?

MikeEdgar: Yes, that's right.

MikeEdgar: Does that change your thinking around the concurrency protection you mentioned earlier?

MikeEdgar:
@franz1981 I'm considering doing a release this week and updating the version used by Quarkus (which requires some code changes as well). I think this PR would be good to get in for the release. Are you planning to push another commit to remove the `put` on the `REGISTRY`?

franz1981:
Sure thing, I was 100% sure to have pushed it already, but is likely that I didn't, sorry for that!

MikeEdgar (approved):

Issue #1819, opened 2024-04-29
Support for (some) Springdoc annotations in the openapi spring module


phillip-kruger:
See https://github.com/quarkusio/quarkus/issues/40328

/cc @mrickly

@MikeEdgar - w.d.y.t ?

MikeEdgar:
This makes sense to me. Probably best to isolate it to the Spring module.

phillip-kruger:
@mrickly are you keen to do a PR for this ?

mrickly:
@phillip-kruger : We are open to giving it a try. But we will need some guidance, e.g. the correct location/package for such a change. We would probably start with the ParameterObject and if it works as expected add things as the needs arise.

phillip-kruger:
Ok No worries. Most of the work is done in core (https://github.com/smallrye/smallrye-open-api/tree/main/core) and core then calls out to one of jax-rs, spring or vertx. So this should be done somewhere in spring module (https://github.com/smallrye/smallrye-open-api/tree/main/extension-spring). We just need to see if the "hooks" exist in core. I would assume it will work the same way as `BeanParam` in jax-rs. (see https://github.com/smallrye/smallrye-open-api/blob/e47033dca7fe383ebfcc64b4c8abf64156966a3e/extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsParameter.java#L25). @MikeEdgar, anything to add ?

MikeEdgar:
This will likely be in the [SpringParameterProcessor](https://github.com/smallrye/smallrye-open-api/blob/main/extension-spring/src/main/java/io/smallrye/openapi/spring/SpringParameterProcessor.java). I'm not entirely sure if `@ParameterObject` is analogous to `@BeanParam` or is it more like the MP OpenAPI `@Parameter` (i.e. documentation only, not functional).

flyinfish:
Hi @MikeEdgar @phillip-kruger 

i posted PR https://github.com/smallrye/smallrye-open-api/pull/1823 for @mrickly .

this would solve our needs. situation is somehow "shady" as quarkus does not seem to behave exactly as spring needing @QueryParam on fields while spring does not.

thanks for your review. tell us if you need more tests whatever

PR #1823, opened 2024-05-06
feat(spring): support SpringDoc `@ParameterObject`

flyinfish:
proposal for support of @ParameterObject as discussed in 
https://github.com/smallrye/smallrye-open-api/issues/1819

### DIFF
--- a/extension-spring/src/main/java/io/smallrye/openapi/spring/SpringParameter.java
+++ b/extension-spring/src/main/java/io/smallrye/openapi/spring/SpringParameter.java
@@ -17,8 +17,8 @@ public enum SpringParameter {
     QUERY_PARAM(SpringConstants.QUERY_PARAM, Parameter.In.QUERY, null, Parameter.Style.FORM),
     HEADER_PARAM(SpringConstants.HEADER_PARAM, Parameter.In.HEADER, null, Parameter.Style.SIMPLE),
     COOKIE_PARAM(SpringConstants.COOKIE_PARAM, Parameter.In.COOKIE, null, Parameter.Style.FORM),
-    PARAMETER_OBJECT(SpringConstants.PARAMETER_OBJECT, null, null, null),
-    JAKARTA_QUERY_PARAM(SpringConstants.JAKARTA_QUERY_PARAM, Parameter.In.QUERY, null, Parameter.Style.FORM);
+    // SpringDoc annotation to indicate a bean with parameters (like Jakarta @BeanParam)
+    PARAMETER_OBJECT(SpringConstants.PARAMETER_OBJECT, null, null, null);
 
     //BEAN_PARAM(SpringConstants.BEAN_PARAM, null, null, null),
     //FORM_PARAM(SpringConstants.FORM_PARAM, null, Parameter.Style.FORM, Parameter.Style.FORM),
--- a/extension-spring/src/main/java/io/smallrye/openapi/spring/SpringParameter.java
+++ b/extension-spring/src/main/java/io/smallrye/openapi/spring/SpringParameter.java
@@ -16,7 +16,9 @@ public enum SpringParameter {
     MATRIX_PARAM(SpringConstants.MATRIX_PARAM, Parameter.In.PATH, Parameter.Style.MATRIX, Parameter.Style.MATRIX),
     QUERY_PARAM(SpringConstants.QUERY_PARAM, Parameter.In.QUERY, null, Parameter.Style.FORM),
     HEADER_PARAM(SpringConstants.HEADER_PARAM, Parameter.In.HEADER, null, Parameter.Style.SIMPLE),
-    COOKIE_PARAM(SpringConstants.COOKIE_PARAM, Parameter.In.COOKIE, null, Parameter.Style.FORM);
+    COOKIE_PARAM(SpringConstants.COOKIE_PARAM, Parameter.In.COOKIE, null, Parameter.Style.FORM),
+    PARAMETER_OBJECT(SpringConstants.PARAMETER_OBJECT, null, null, null),
+    JAKARTA_QUERY_PARAM(SpringConstants.JAKARTA_QUERY_PARAM, Parameter.In.QUERY, null, Parameter.Style.FORM);
 
     //BEAN_PARAM(SpringConstants.BEAN_PARAM, null, null, null),
     //FORM_PARAM(SpringConstants.FORM_PARAM, null, Parameter.Style.FORM, Parameter.Style.FORM),
--- a/extension-spring/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/GreetingParam.java
+++ b/extension-spring/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/GreetingParam.java
@@ -1,0 +1,16 @@
+package test.io.smallrye.openapi.runtime.scanner.entities;
+
+import jakarta.ws.rs.QueryParam;
+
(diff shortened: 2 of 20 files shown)


extension-spring/src/main/java/io/smallrye/openapi/spring/SpringParameter.java
```diff
@@ -16,7 +16,9 @@ public enum SpringParameter {
     MATRIX_PARAM(SpringConstants.MATRIX_PARAM, Parameter.In.PATH, Parameter.Style.MATRIX, Parameter.Style.MATRIX),
     QUERY_PARAM(SpringConstants.QUERY_PARAM, Parameter.In.QUERY, null, Parameter.Style.FORM),
     HEADER_PARAM(SpringConstants.HEADER_PARAM, Parameter.In.HEADER, null, Parameter.Style.SIMPLE),
-    COOKIE_PARAM(SpringConstants.COOKIE_PARAM, Parameter.In.COOKIE, null, Parameter.Style.FORM);
+    COOKIE_PARAM(SpringConstants.COOKIE_PARAM, Parameter.In.COOKIE, null, Parameter.Style.FORM),
+    PARAMETER_OBJECT(SpringConstants.PARAMETER_OBJECT, null, null, null),
+    JAKARTA_QUERY_PARAM(SpringConstants.JAKARTA_QUERY_PARAM, Parameter.In.QUERY, null, Parameter.Style.FORM);
```
phillip-kruger, line 21: Why is this needed ?

extension-spring/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/GreetingParam.java
```diff
@@ -0,0 +1,16 @@
+package test.io.smallrye.openapi.runtime.scanner.entities;
+
+import jakarta.ws.rs.QueryParam;
+
+public class GreetingParam {
+    @QueryParam("nameQuery") // "real" spring does not require this, but quarkus-spring-web does
```
MikeEdgar, line 6: Does this mean neither the OpenAPI nor the actual web application work correctly without using the Jakarta REST `@QueryParam` annotation in this way?

flyinfish: yes that is right

@phillip-kruger this is the reason for having `JAKARTA_QUERY_PARAM` in `SpringParameter.java`

extension-spring/src/main/java/io/smallrye/openapi/spring/SpringParameter.java
flyinfish: @see https://github.com/smallrye/smallrye-open-api/pull/1823#discussion_r1591279541

extension-spring/src/test/java/test/io/smallrye/openapi/runtime/scanner/entities/GreetingParam.java
flyinfish: i did play around with https://github.com/flyinfish/spring-web-quickstart/compare/feature/adapt-to-patch?expand=1

`HelloParamCheatedWithQueryParam` has the same role there as `GreetingParm` above

phillip-kruger: That does not seem right. At least the actual endpoint should work, even if openapi can not generate correctly. @geoand ?

geoand: AFAIR, we don't support Spring's ParamObject, although we certainly could

MikeEdgar: The issue (I haven't tried using Spring in Quarkus myself) is that using `GreetingParam` as a REST method parameter on a `GET` request doesn't automatically take the bean's fields to be query parameters without adding a Jakarta annotation to the fields. If that's how Spring really behaves (I don't know), then I think both Quarkus and the OpenAPI scanner should understand that whether the springdoc `@ParameterObject` annotation is present or not.

phillip-kruger: Agree!

flyinfish: i agree that `@ParameterObject` is only for documentation.

there is an issue that spring obviously does not depend on jakarta-query-param beeing present on query-class. but quarkus does - so what is your suggestion?

flyinfish: Hi @MikeEdgar @phillip-kruger 

can you please approve or say what more you need?

in a perfect situation quarkus would support objects on get-methods as spring does. since quarkus needs @jakarka-query-param, this pr is imho a suitable solution as it supports on openapi-docs the features quarkus support at runtime at least as far as i understand it.

were migrating like one hundred spring-microservices to quarkus and would be very happy to do it with the least possible change of code, which means migrating to quarkus-spring-web rather than "pure" quarkus.

thank you very much

phillip-kruger: @geoand - can you comment on this comment: https://github.com/smallrye/smallrye-open-api/pull/1823#discussion_r1592151911

geoand: I don't remember how Spring behaves in this case, I would have to check and unfortunately I am short on time at the moment

flyinfish: @MikeEdgar if we were talking spring i would agree shat scanner should not use jakarta-annotations.

since we are talking quarkus (do we? is this project used outside quarkus?) i would not be too happy if the scanner generates an api which then does not work until i add the missing jakarta-annotations.

phillip-kruger: This is being used in many other projects, not only Quarkus. I agree with @MikeEdgar, the fix should be that we handle the un-annotated case.

MikeEdgar: I've re-based this one and pushed another commit to remove the reliance on the Jakarta annotation for Spring. As we discussed earlier in this thread, Quarkus itself may still not handle this scenario (I haven't looked at the changes in Quarkus Spring since the last comment), but this at least will scan the way it should work with native Spring.

@flyinfish, would you mind reviewing the current state of this PR and let us know if it should work as needed for your OpenAPI documentation?

flyinfish: @MikeEdgar 
sorry for late reply - Yes ist works as needed - Thank you very much!

nevertheless we moved away from spring-approach and convert our legacy apps now to proper quarkus/microprofile annotations.

anyhow for spring-users this should help!

MikeEdgar:
@phillip-kruger please give this another look since the latest changes.

phillip-kruger (approved):
LGTM

PR #1834, opened 2024-05-09
Support identifier field in licence objects

benjamin-confino:
This keeps smallrye up to date with the spec change implemented here: https://github.com/eclipse/microprofile-open-api/issues/436

And this PR depends on https://github.com/eclipse/microprofile-open-api/pull/611

I investigated to see if [PathItem](https://spec.openapis.org/oas/v3.1.0.html#path-item-object] parameters had any code to emit a warning or exception if the constraint is violated and could not find one so I did not add anything to enforce the exclusivity of URL and Identifier. Let me know if I missed something.

For the test I used the maven plugin test to ensure one plugin tests each of the URL and identifier without violating the constraint, let me know if you'd prefer both test both


MikeEdgar (commented):
@benjamin-confino I think this needs to be rebased, then it should be good.

MikeEdgar (approved):

PR #1836, opened 2024-05-10
Implement webhooks model

Azquelt:
- Note that I haven't made a `WebhookIO` class.
  - For the model, the new `webhooks` property just holds a `Map<String, PathItem>`
  - For the annotation, a `@Webhook` annotation represents one entry in that map. Since the value is a `PathItem`, I added it to `PathItemIO`. I could make a new IO class for it instead, though there's no model object to read and write to JSON.
- I haven't added any tests here because I think the TCK tests are sufficient.

Implements eclipse/microprofile-open-api#612

~In order to pass the TCK locally, I also have to merge in #1818 because that implements eclipse/microprofile-open-api#604 which has already been merged.~

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/api/models/OpenAPIImpl.java
+++ b/core/src/main/java/io/smallrye/openapi/api/models/OpenAPIImpl.java
@@ -1,11 +1,14 @@
 package io.smallrye.openapi.api.models;
 
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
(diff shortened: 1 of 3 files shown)


Azquelt:
Rebased on `main-4.0`

Azquelt:
The annotation has been removed from the upstream PR. Removed the implementation here to match.

MikeEdgar (approved):

PR #1858, opened 2024-05-31
Require requestBody by default

Azquelt:
When generating a requestBody for a resource method or from an annotation, it should be required by default, unless something explicitly indicates that it's not required.

The change affects generation of `requestBody` only when scanning code. If there's an operation which is annotated `@Consumes`, the generated `requestBody` will have `required = true`, unless the user annotates the method with `@RequestBody(required = false)`.

However, if the user calls `OASFactory.createRequestBody` in a filter or reader and doesn't call `setRequired`, then the output will not include `required = true`.

Implements eclipse/microprofile-open-api#627

Also delete a bunch of test json files which weren't being used.


Azquelt:
This was a bit tricky to implement.

There are two code paths where we read create a request body during code scanning:
* When `@RequestBody` is used as a child of `@CallbackOperation` or `@Components`
* When scanning a resource method

The first case is a simple read of the annotation with no other changes made.

The second case reads the annotation if it's present and then combines it with more information that can be discovered from the method. This scanning code assumes that if a property of `RequestBody` is not `null`, then it's been set by an earlier part of the scanning process and shouldn't be overwritten.

If we set `required = true` when reading the annotation, we lose the ability to query whether something else has set the property.

If we try to set `required = true` at the end of code scanning, we end up having to do that everywhere the annotation is read - in `AnnotationScanner`, `ComponentsIO` and `CallbackOperationIO` which feels like something those IO classes shouldn't really have to deal with.

That's why I ended up extending `RequestBodyImpl` to allow a default value to be set and to query whether any earlier code has called `setRequired()`.

I'm still not super happy about this approach, it wouldn't scale very well if there were lots of properties with a default value and it means `required` has to be treated separately. I do wonder whether the current approach of "has this been set already" is really the right one. Some of the discussion in eclipse/microprofile-open-api#356 suggests that having a property on an annotation set to the default value should be treated the same as not setting it at all which is at odds with the current scanning code.

Azquelt:
I'm a bit surprised that the 4.0 TCK is passing without eclipse/microprofile-open-api#627 having been merged. I guess that just shows that there was a gap in the TCK coverage.

Azquelt:
Rebased on `main-4.0`.

MikeEdgar (approved):

Issue #1892, opened 2024-07-02
Allow 4.0 to pass the older MP OpenAPI TCKs


Azquelt:
At the moment, the `main-4.0` branch passes the MP OpenAPI 4.0 TCK.

Ideally, we would keep compatibility with older versions and continue to pass those TCKs.

This is quite tricky because there are breaking API changes in MP OpenAPI 4.0 in the models packages.

PR #1918, opened 2024-07-12
Support generation of OpenAPI v3.0

Azquelt:
Support for OpenAPI v3.0 is implemented as a transformation on input and
output. The changes between the two formats are minimal and mostly in
the Schema object.

When reading or writing the OpenAPI version is read from the document
or model respectively.

When generating a document for an application, config is read and if a
verison is configured the OpenAPI version in the model will be updated
before it is output.

Also fix `VersionTest` so that it tests what it's meant to.


Azquelt:
Draft for the moment because I realised we need to do a smarter transform for the schema we create when we want to reference a schema but make it optional:

```json
                  "anyOf": [
                    {
                      "$ref" : "#/components/schemas/Greeting"
                    },
                    {
                      "type" : "null"
                    }
                  ]
```

I think this can either remain exactly as-is, or we need to change it back to what we used to create:

```json
                  "$ref" : "#/components/schemas/Greeting",
                  "allOf": [
                    {
                      "nullable" : true
                    }
                  ]
```

What I don't think works is what the code currently produces:

https://github.com/smallrye/smallrye-open-api/pull/1918/files#diff-57214ee0f8afb8375d24dc188e3ab57c9284512982e7a8aee6e54cb48b711d5dR24-R31

Azquelt:
Sonarcloud also shows a few conditions in `SchemaIO` not tested which it should be easy to hit.

MikeEdgar:
I think the second option is the only valid choice for OAS 3.0 (unless there is another, better option).

```json
                  "$ref" : "#/components/schemas/Greeting",
                  "allOf": [
                    {
                      "nullable" : true
                    }
                  ]
```

Azquelt:
Discussed this on the call today and came to a few conclusions.

1. We have to translate both ways. Since our model always conforms to the 3.1 spec, we need to be able to translate 3.0 to 3.1 when necessary when reading, and translate 3.1 to 3.0 when necessary when writing.

2. `"type": "null"` is equivalent to `"enum" : [null]`, which is valid in 3.0.

3. Ideally, we would produce the same output as older versions if the user configures us for 3.0 output. This means that where possible, we should produce

Azquelt:
I think this is now ready for review.

What is covered:
- Dropping fields which were new in 3.1 when outputting in 3.0 format
  - Mostly tested in `testEverything30`
- Transforming Schemas between 3.0 and 3.1 with some exceptions for `nullable`
  - I've covered the straightforward case where `nullable` is used with `type`
  - I've covered the `nullable` + `$ref` cases we're likely to create when doing annotation scanning
  - There are other `nullable` +`$ref` cases which aren't handled (e.g. `nullable` with `allOf` containing multiple `$ref` schemas, `nullable` with `anyOf` containing `$ref` schemas)
    - I've included unit tests for some of these cases which don't work, asserting the current behaviour with a note that the current behaviour isn't correct.
    - While I have noted cases we could handle better, I don't think we can handle all `nullable` and `$ref` cases correctly

MikeEdgar (approved):
This looks reasonable to me. I'll re-base my local work for #1892 on this and see if that reveals any gaps or cases not already covered here. Thank you for working on this.

MikeEdgar:
@Azquelt do you want to do one more re-base before I merge this?

Azquelt:
Yes, I'll do that now

Azquelt:
Rebased on `main-4.0`

PR #2012, opened 2024-10-07
Fix scanning of empty SecurityRequirements on resource class

### LINKED ISSUE #2011: Empty @SecurityRequirements annotation is only honored on methods but not on classes
The behavior of the smallrye-open-api-maven-plugin to generate an OpenAPI schema from Jax-RS sources has changed.

Up until (including) version 3.10.0, putting an empty \@SecurityRequirements annotation on a **class** lead to a schema, where each operation in the resulting schema was annotated with "security: []". This is important for us during validation of the schema.

Since version 3.11.0, this attribute is missing from the generated schema.

Putting an empty  \@SecurityRequirements on a **method** still results in an attribute being added to that operation. So our current workaround is to put the annotation on every single method. But, of course, annotating the whole class is much easier, and the change in behavior looks like a regression and not like a intended feature.

I have attached a small test case that illustrates the problem.

[smallrye-security-requirements.zip](https://github.com/user-attachments/files/17280959/smallrye-security-requirements.zip)

MikeEdgar:
Fixes #2011

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
@@ -665,6 +665,7 @@ public interface AnnotationScanner {
 
         if (securityRequirements.isEmpty() && !emptyContainerPresent) {
             securityRequirements = context.io().securityIO().readRequirements(resourceClass);
+            emptyContainerPresent = isEmptySecurityRequirements(context, resourceClass);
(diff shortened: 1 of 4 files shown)


MikeEdgar:
@phillip-kruger please take a look. One-liner fix and the rest is testing. I'll do another alpha release for 4.0 after this. Thanks!

phillip-kruger (approved):

phillip-kruger:
Sorry, I missed this one

PR #2022, opened 2024-10-22
Split and Reuse the Release workflow

MikeEdgar:


MikeEdgar:
@radcortez how does this PR to use the new release process look? Note that OpenAPI (and GraphQL) publish a Gradle plugin. I didn't think that was yet part of the new common release process so it's still left here. I suppose it would have some of the same issues as the publish to Maven Central.

radcortez:
> @radcortez how does this PR to use the new release process look? Note that OpenAPI (and GraphQL) publish a Gradle plugin. I didn't think that was yet part of the new common release process so it's still left here. I suppose it would have some of the same issues as the publish to Maven Central.

Yes, we need to figure something out for that case.

Everything else looks good. Did you set up the ruleset described here https://github.com/smallrye/smallrye/wiki/Release-Process#pull-request-model?

MikeEdgar:
> Did you set up the ruleset described here https://github.com/smallrye/smallrye/wiki/Release-Process#pull-request-model?

No, I haven't. I'll do that next. Once that's done, it's safe to merge this and try a trial release?

radcortez:
Yes!

PR #2030, opened 2024-10-23
Additional type checks in models and Json/YAML reader

MikeEdgar:


MikeEdgar:
@Azquelt - fyi

core/src/main/java/io/smallrye/openapi/runtime/io/ModelIO.java
```diff
@@ -330,7 +307,7 @@ protected Object readJson(V node, DataType desiredType) {
             } else if (desiredType.clazz == Object.class) {
                 return jsonIO().fromJson(node);
             } else {
-                // Log node dropped?
+                IoLogging.logger.invalidJsonType(desiredType.clazz.getName(), String.valueOf(node));
```
Azquelt, line 310: `desiredType.clazz` could be `null` here

Azquelt: This is true of both logging statements. If `desiredType.type` is `LIST` or `MAP`, then `desiredType.clazz` will be `null`.

Azquelt:
This is definitely better.

It does still have one change in behaviour for `Schema`.

If I read in a schema like this:
```json
{
    "type": "object",
    "properties": ["id", "name"]
}
```

The `properties` field is dropped when reading because it's not a map, which wasn't the case in alpha6.

Using the `Schema.set(String, Object)`, I can set `properties` to whatever I want, it only gets dropped during reading.

This does slightly affect our ability to pass through a schema object we don't understand untouched. I've written some tests to demonstrate the change in behaviour here: https://github.com/Azquelt/smallrye-open-api/commit/0e17e6b9eb70adb0ff9a505129f84aecef82b602 - with these changes `testSchemas_InvalidTypes` is the only one that doesn't pass.

core/src/main/java/io/smallrye/openapi/runtime/io/ModelIO.java
MikeEdgar: Fix for this is pushed - added a `DataType#toString` method.

core/src/main/java/io/smallrye/openapi/runtime/io/JacksonJsonIO.java
```diff
@@ -334,7 +339,7 @@ public Object fromJson(JsonNode value) {
     @SuppressWarnings("unchecked")
     @Override
     public <T> T fromJson(JsonNode object, Class<T> desiredType) {
-        if (desiredType == String.class) {
+        if (desiredType == String.class && object.isValueNode()) {
```
MikeEdgar, line 342: I added this so that `{}` and `[]` are not coerced to empty strings whereas any value type will be. This matches the Jakarta JSON reader (which I don't think anyone uses yet), but also seems more intuitive. wdyt?

MikeEdgar:
> The `properties` field is dropped when reading because it's not a map, which wasn't the case in alpha6.
> 
> Using the `Schema.set(String, Object)`, I can set `properties` to whatever I want, it only gets dropped during reading.
> 
> This does slightly affect our ability to pass through a schema object we don't understand untouched. I've written some tests to demonstrate the change in behaviour here: [Azquelt@0e17e6b](https://github.com/Azquelt/smallrye-open-api/commit/0e17e6b9eb70adb0ff9a505129f84aecef82b602) - with these changes `testSchemas_InvalidTypes` is the only one that doesn't pass.

I've added an additional fallback for `Schema` where it will read any type for the property. The last push also includes your commit cherry-picked to this PR (if that's ok :smiley:)

core/src/main/java/io/smallrye/openapi/runtime/io/JacksonJsonIO.java
Azquelt: Yeah, that makes sense to me.

Azquelt (approved):
These changes look good to me. There were a few other small changes in behaviour which needed tweaks in OpenLiberty (noted on #2021), but now I have all my tests passing.

phillip-kruger (approved):

PR #2050, opened 2024-10-30
Make some ModelIO instance fields constants

Azquelt:
The initialization of these two fields showed up in performance testing, since we create quite a few ModelIO-based objects when parsing or serializing and `SmallRyeOASModels` populates a large map in its constructor.

An alternative would be to make `SmallRyeOASModels` a singleton, but since this is the only place it's used, making the field a constant should fix the performance issue.


MikeEdgar (approved):

MikeEdgar (commented):
Thanks @Azquelt . I'll probably release 4.0.1 later today or tomorrow once I have the other fixes verified.

PR #2053, opened 2024-11-03
Allow use of `@JsonView` on class

### LINKED ISSUE #1764: OpenAPI generation of JsonView annotated classes or Ignored fields does not work
The issue [here](https://github.com/smallrye/smallrye-open-api/issues/1724) shows a previous problem where OpenAPI was not respecting fields annotated with `@JsonView`. The applied fix seems to work for simple cases, but does not work when a _class_ is annotated with `@JsonView` or if the ObjectMapper is set not to add fields to views by default. Examples are in Kotlin/Quarkus, using version 3.10.0 of smallrye-open-api via Quarkus 3.8.1.

### Example 1 (works as expected):
```kotlin
    @JsonView(View.Public::class)
    @GET
    @Path("/public")
    fun userPublic(): User? {
        return User().also {
            it.id = 123
            it.name = "Person"
        }
    }

    class User {
        @JsonView(View.Private::class)
        var id = 0

        @JsonView(View.Public::class)
        var name: String? = null
    }

    class View {
        open class Public
        class Private : Public()
    }
```
Endpoint /public returns `{"name":"Person"}`
Generated type schema (as expected):
```yaml
    User_Public:
      type: object
      properties:
        name:
          type: string
          nullable: true
```

### Example 2 (annotate class definition with `@JsonView`):

```kotlin
    @JsonView(View.Public::class)
    @GET
    @Path("/public")
    fun userPublic(): User? {
        return User().also {
            it.id = 123
            it.name = "Person"
        }
    }

    @JsonView(View.Private::class)
    class User {
        var id = 123

        @JsonView(View.Public::class)
        var name: String? = null
    }

    class View {
        open class Public
        class Private : Public()
    }
```

Endpoint /public returns `{"name":"Person"}` (as expected)
Generated type schema (**Note: id field should not be shown here**):
```yaml
    User_Public:
      type: object
      properties:
        id:
          format: int32
          type: integer
        name:
          type: string
          nullable: true
```

### Example 3 (change ObjectMapper to exclude fields from `@JsonView` by default):

ObjectMapper can be setup as follows to exclude fields by default from `@JsonView`:

```kotlin
@Singleton
class RegisterCustomModuleCustomizer : ObjectMapperCustomizer {
    override fun customize(mapper: ObjectMapper) {
        mapper.apply {
            disable(MapperFeature.DEFAULT_VIEW_INCLUSION)
        }
    }
}
```

Neither the `id` field or the `User` class now need to be annotated to be ignored by Jackson:

```kotlin
    @JsonView(View.Public::class)
    @GET
    @Path("/public")
    fun userPublic(): User? {
        return User().also {
            it.id = 123
            it.name = "Person"
        }
    }

    class User {
        var id = 123

        @JsonView(View.Public::class)
        var name: String? = null
    }

    class View {
        open class Public
        class Private : Public()
    }
```

Endpoint /public returns `{"name":"Person"}` (as expected)
Generated type schema (**Note: id field should not be shown here**):
```yaml
    User_Public:
      type: object
      properties:
        id:
          format: int32
          type: integer
        name:
          type: string
          nullable: true
```

MikeEdgar:
Fixes #1764

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
@@ -42,6 +42,7 @@ import io.smallrye.openapi.api.constants.JaxbConstants;
 import io.smallrye.openapi.api.constants.JsonbConstants;
 import io.smallrye.openapi.runtime.io.schema.SchemaConstant;
 import io.smallrye.openapi.runtime.scanner.spi.AnnotationScannerContext;
+import io.smallrye.openapi.runtime.util.Annotations;
 import io.smallrye.openapi.runtime.util.JandexUtil;
 import io.smallrye.openapi.runtime.util.TypeUtil;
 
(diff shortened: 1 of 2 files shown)


phillip-kruger (approved):

phillip-kruger:
Nice !

PR #2057, opened 2024-11-05
fix: check value type before Jackson getJsonBigDecimal

### LINKED ISSUE #2054: JacksonJsonIO may throw exception when trying to read a number
[JacksonJsonIO.getJsonBigDecimal](https://github.com/smallrye/smallrye-open-api/blob/ecb1ae514d64a642673b4fb9e58fdeab138c0631/core/src/main/java/io/smallrye/openapi/runtime/io/JacksonJsonIO.java#L153-L157) does not check that the node contains a number before trying to parse it as a big decimal.

This results in a `NumberFormatException` rather than a `null` [as the Javadoc would suggest](https://github.com/smallrye/smallrye-open-api/blob/ecb1ae514d64a642673b4fb9e58fdeab138c0631/core/src/main/java/io/smallrye/openapi/runtime/io/JsonIO.java#L308-L315).

Admittedly, the Javadoc was written after the implementation, so arguably it's the Javadoc that's wrong here, but having this method return a `null` would be consistent with the other methods in this class.

This can result in throwing an exception when parsing an OpenAPI 3.0 schema rather than ignoring the invalid field.

Ideally we should unit test the `JsonIO` implementations. There may be other similar bugs.

MikeEdgar:
Fixes #2054 

Also avoid re-parsing BigDecimal/BigInteger values in fromJson

### DIFF
--- a/core/src/test/java/io/smallrye/openapi/runtime/io/JacksonJsonIOTest.java
+++ b/core/src/test/java/io/smallrye/openapi/runtime/io/JacksonJsonIOTest.java
@@ -1,0 +1,19 @@
+package io.smallrye.openapi.runtime.io;
+
+import org.junit.jupiter.api.BeforeEach;
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+class JacksonJsonIOTest extends JsonIOTest<JsonNode, ArrayNode, ObjectNode, ArrayNode, ObjectNode> {
+
+    @BeforeEach
+    void setup() {
+        super.target = new JacksonJsonIO(new ObjectMapper()
+                .enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS));
+    }
+
+}
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/JacksonJsonIO.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/JacksonJsonIO.java
@@ -153,7 +153,7 @@ class JacksonJsonIO implements JsonIO<JsonNode, ArrayNode, ObjectNode, ArrayNode
     @Override
     public BigDecimal getJsonBigDecimal(ObjectNode object, String key) {
         JsonNode value = object.get(key);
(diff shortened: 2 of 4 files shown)


MikeEdgar:
@Azquelt FYI

core/src/test/java/io/smallrye/openapi/runtime/io/JacksonJsonIOTest.java
```diff
@@ -0,0 +1,19 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+class JacksonJsonIOTest extends JsonIOTest<JsonNode, ArrayNode, ObjectNode, ArrayNode, ObjectNode> {
+
+    @BeforeEach
+    void setup() {
+        super.target = new JacksonJsonIO(new ObjectMapper()
+                .enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS));
```
Azquelt, line 16: Is `USE_BIG_DECIMAL_FOR_FLOATS` just needed to get predictable accuracy for the tests? We don't appear to use it anywhere else in the code.

Azquelt: Hmm, I assume this will change the value that `JsonIO.fromJson(JsonNode)` returns for a decimal number from a `double` to a `BigDecimal`. I guess usually we'll call `JsonIO.fromJson(JsonNode, DataType)` instead though.

It seems odd to be testing with an option that we don't enable at runtime, though the user or integrator could provide their own `ObjectMapper` and I suppose we're trying to test the path where `fromJson(JsonNode)` does return a `BigDecimal`.

MikeEdgar: > It seems odd to be testing with an option that we don't enable at runtime, though the user or integrator could provide their own `ObjectMapper` and I suppose we're trying to test the path where `fromJson(JsonNode)` does return a `BigDecimal`.

Right - it's contrived for the test. As you said we call the overload with the expected type usually, and this is the fallback. It would also be used for things like examples or extensions. It's not really possible for an integrator to provide the `ObjectMapper` cleanly yet, but that is something that could be available at some point. It would be useful in cases like the test here where a `double` would truncate the user's value in an example/extension.

Azquelt (approved):

phillip-kruger (approved):

Issue #2076, opened 2024-11-20
Schema example should not auto convert to examples


dhoffer:
In our upgrade from Swagger libraries to smallrye openapi it is converting all our 'example' to 'examples' in the OpenAPI schema.  

This is not desirable as we need to be able provide an example that matches the data type.  We use arrays extensively and even double arrays occasionally.  With smallrye openapi wrapping the example in another array, will be very confusing for users.

I understand the microprofile-openapi-api has:

A free-form property to include an example of an instance for this schema.
Deprecated: use examples()
`@Deprecated(since = "4.0")
 String example() default "";`

But for now, example and examples are both valid per the OpenAPI spec.

https://github.com/oai/openapi-specification/blob/main/versions/3.1.0.md#schema-object

Deprecated: The example property has been deprecated in favor of the JSON Schema examples keyword. Use of example is discouraged, and later versions of this specification **may** remove it.

I would like to see smallrye openapi support example until it is actually removed which may never happen.   And also support examples.

MikeEdgar:
@Azquelt do you have any thoughts on this request? I suppose it would be simple enough to keep `example` and `examples` separated.

FYI @michalvavrik. I remember you mentioned some impact to testing around the change from `example: any` to `examples: []`.

michalvavrik:
> FYI @michalvavrik. I remember you mentioned some impact to testing around the change from `example: any` to `examples: []`.

yes, this issue describes exactly what I meant, I didn't realize it was only deprecated; I need to be more careful next time

MikeEdgar:
Well, it is deprecated, but the scanner currently combines them all into the array, whether an example is in `example` or `examples` on the `@Schema` annotation. The ask is to map them without combining.

dhoffer:
Yes exactly we think you should support both **example** and **examples** but don't combine them.

MikeEdgar:
The current behavior may be enforced by the MicroProfile OpenAPI 4.0.2 TCK test suite (I only glanced at it). If that's a case and it's agreed that it should be changed, this will need to be changed there first.

MikeEdgar:
Related issue in MicroProfile OpenAPI: https://github.com/eclipse/microprofile-open-api/issues/674

dhoffer:
I'm not following your first point/reason.  I think libraries should code for user's 'least surprise'.  We were not expecting example in the annotations to be examples in the OpenAPI model.

[...]

PR #2079, opened 2024-11-20
fix: preserve property order read from YAML/JSON input

### LINKED ISSUE #2077: OpenAPI component schema generation should match the field order of the Java object(s)
In our upgrade from Swagger libraries to smallrye openapi it is not honoring the order of the fields in the entities it is generating the schema for.  So the resulting OpenAPI schema looks vastly different than the POJO it represents.   All the fields exist but in what appears to be random order which is very undesirable.

We group related fields together in the POJO and need that same order in the OpenAPI schema.   

We just discoveered that this is a regression, it works fine with Quarkus 3.16.3 but it fails when we add these later dependencies that have other fixes we need.

`                    <dependency>
                        <groupId>io.smallrye</groupId>
                        <artifactId>smallrye-open-api-core</artifactId>
                        <version>4.0.3</version>
                    </dependency>
                    <dependency>
                        <groupId>io.smallrye</groupId>
                        <artifactId>smallrye-open-api-jaxrs</artifactId>
                        <version>4.0.3</version>
                    </dependency>
                    <dependency>
                        <groupId>org.eclipse.microprofile.openapi</groupId>
                        <artifactId>microprofile-openapi-api</artifactId>
                        <version>4.0.2</version>
                    </dependency>`

MikeEdgar:
Fixes #2077

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/ModelIO.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/ModelIO.java
@@ -2,7 +2,6 @@ package io.smallrye.openapi.runtime.io;
 
 import java.util.AbstractMap;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


phillip-kruger (approved):

dhoffer:
I can confirm this is fixed in the 4.0.4-SNAPSHOT version.  When do you expect that 4.0.4 will be released?

MikeEdgar:
Let's see the outcome of #2076

Issue #2092, opened 2024-12-05
Parameters and BeanParam being generated in alphabetical order in JSON


mrobertorr1:
Hello, everyone. I'm having difficulty generating the JSON/YAML using the smallrye-open-api-maven-plugin. The properties of the methods annotated with @Parameter are being automatically sorted alphabetically, meaning they are not respecting the order in which the attributes are declared. The same happens when I try to use a @BeanParam. All the DTO attributes are sorted alphabetically, and I need them to follow the declaration order of the class. Can anyone help me? I’ve already tried adding mp.openapi.extensions.smallrye.sorted-properties.enable=false, but it had no effect. I’ve created an example with the issue:

**Obs:** Environment configured with Java 17 and Maven 3.

**Endpoint:**

```
package com.example;

import jakarta.ws.rs.*;
import jakarta.ws.rs.core.MediaType;
import jakarta.ws.rs.core.Response;
import org.eclipse.microprofile.openapi.annotations.Operation;
import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;

@Path("/people")
public class PersonResource {
    @GET
    @Produces(MediaType.APPLICATION_JSON)
    @Operation(summary = "Performs a search with three parameters", description = "Performs a search based on name, age, and city.")
    public Response search(
        @Parameter(description = "Name of the person for the search", required = true)
        @QueryParam("name") String name,
        
        @Parameter(description = "Age of the person for the search", required = false)
        @QueryParam("age") Integer age,
        
        @Parameter(description = "City of the person for the search", required = false)
        @QueryParam("city") String city
    ) {
        // Simulating search logic
        String result = String.format("Search performed successfully! Name: %s, Age: %d, City: %s", 
                                      name, (age != null ? age : "not provided"), 
                                      (city != null ? city : "not provided"));

        return Response.ok(result).build();
    }
}
``` 

**pom.xml:**

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.example</groupId>
  <artifactId>example-app</artifactId>
  <packaging>jar</packaging>
  <version>1.0-SNAPSHOT</version>
  <name>example-app</name>
[...]

Issue #1957, opened 2024-08-22
Maven plugin should be configurable regarding the order of the parameters in a REST endpoint


cristalp:
We use Swagger annotations and a different Maven plugin to generate OpenAPI from our Java code. But we want to migrate to Eclipse MicroProfile annotations and the Smallrye toolchain.

There is a difference in the generated OpenAPI regarding the order of the parameters in a REST endpoint.

For instance:
```java
  public Response endpoint(
      @Parameter(description = "The foo") @PathParam("foo") final String foo,
      @Parameter(description = "The bar") @PathParam("bar") final String bar) {
    return null;
  }
```

- The Swagger toolchain lists parameters as they are defined in the Java code. So OpenAPI first contains `foo` and then `bar`.
- The Smallrye toolchain lists parameters alphabetically. So OpenAPI first contains `bar` and then `foo`.

While this could be called nitpicking, I do believe that the order of the parameters is important. Main parameters (like an ID) should come first and optional parameters should come later.

In order not to break existing behavior, the Maven plugin should  contain a flag defining how parameters are sorted. For instance `orderParametersAlphabetically` which defaults to `true`. If it's `false`, the order is defined by their order in the Java code.

MikeEdgar:
This is a reasonable request. We currently support it when `@Parameter` is on the method rather than the arguments, but that results in duplication because the `name` at least is needed on the `@Parameter` to link it to the correct argument.

cristalp:
Thanks a lot! I will try it as soon as the new release is available.

PR #2116, opened 2024-12-18
fix: natural order of parameters

davidbasto:
Issues #2092 and #1957.

Fixes the natural ordering inconsistency when the @Parameters annotation is not used. Method parameters and BeanParam were sorted in alphabetical order if they were not declared with @Parameter within the @Parameters annotation.


MikeEdgar (commented):
Thanks for the PR @davidbasto . I need some more time to look into this, so it won't be in the next release quite yet.

MikeEdgar:
The solution for sorting/not sorting parameters needs to be configurable. I've taken this into account with #2248. I'll close this one for now. Thank you for the PR.

PR #2117, opened 2024-12-18
Skip IO of fields added in OpenAPI 3.1 when reading or writing OpenAPI 3.0 documents

### LINKED ISSUE #2105: Invalid properties included in OpenAPI v3.0 output document
If output in OpenAPI v3.0 is configured, properties which were added in OpenAPI 3.1 are included in the output document.

Previously, if the output mode was 3.0, we would drop any properties which are not valid for 3.0, I think that changed in #2021.

I don't think OpenAPI allows the addition of arbitrary fields, so I think we should revert this behaviour change to ensure we're always outputting a valid OpenAPI v3.0 document.

Azquelt:
Add a new field to the model metadata to indicate which version of OpenAPI added the field.

I changed `SmallRyeOASModels.Properties` to `PropertyMetadata` so that it doesn't conflict with other uses of `Properties`, but I'd be happy to change the naming here if other people have better ideas.

More detail in the commit messages.

Fixes #2105


MikeEdgar (dismissed):
Very nice. Thank you @Azquelt for working on this. Are you planning any additional changes or is it good to merge? I don't have any concern with the changes here.

MikeEdgar (approved):

Issue #2118, opened 2024-12-19
smallrye-open-api implements @RequestMapping wrong. Multiple paths are not supported.


geniegeist:
## Description

Given 

```
@RequestMapping("/path1", "/path2")
@RestController
class OverviewController() 
```

I expect the following OpenAPI file

```
paths:
  /path1:
    get: ...
  /path2:
    get: ...
```

Instead, it produces

```
paths:
  /path1/path2:
    get: ...
```

I don't think the output is the correct behaviour. Please let me know. If it is indeed incorrect, I can go ahead and fix it.

MikeEdgar:
Yes, I think your expectation is correct and the current handling should be fixed.

PR #2122, opened 2024-12-27
fix: Add support for multiple paths

geniegeist:
Issue #2118

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/ResourceParameters.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/ResourceParameters.java
@@ -145,6 +145,7 @@ public class ResourceParameters {
         return pathItemPaths.stream()
                 .flatMap(pathItemPath -> operationPaths.stream()
                         .flatMap(operationPath -> getPathParameterTemplateName(pathItemPath, operationPath).stream()))
+                .distinct()
                 .collect(Collectors.toList());
     }
 
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/ResourceParameters.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/ResourceParameters.java
@@ -5,6 +5,7 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import org.eclipse.microprofile.openapi.models.media.Content;
 import org.eclipse.microprofile.openapi.models.media.Schema;
@@ -50,10 +51,10 @@ public class ResourceParameters {
 
     static final Pattern TEMPLATE_PARAM_PATTERN = Pattern.compile("\\{(\\w[\\w\\.-]*)\\}");
 
-    private String pathItemPath;
+    private List<String> pathItemPaths;
     private List<Parameter> pathItemParameters;
 
-    private String operationPath;
+    private List<String> operationPaths;
     private List<Parameter> operationParameters;
 
     private Content formBodyContent;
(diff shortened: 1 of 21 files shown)


core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
```diff
@@ -1151,52 +1113,60 @@ protected Object primitiveToObject(Primitive primitive, String stringValue) {
...
+     * class. Method-level targets will include both the paths to the resource and the paths to the method joined with a '/'.
      *
      * @param target target item for which the path is being generated
-     * @return full path (excluding application path) of the target
+     * @return full paths (excluding application path) of the target
      */
-    protected String fullPathOf(AnnotationTarget target) {
-        String pathSegment = null;
+    protected List<String> fullPathsOf(AnnotationTarget target) {
+        List<String> pathSegments = List.of("null");
```
MikeEdgar, line 1151: This looks a little funky, but I realize the current code would effectively do the same thing.

core/src/main/java/io/smallrye/openapi/runtime/scanner/ResourceParameters.java
```diff
@@ -139,7 +142,10 @@ public void sort(List<Parameter> preferredOrder) {
     }
 
     public List<String> getPathParameterTemplateNames() {
-        return getPathParameterTemplateName(this.pathItemPath, this.operationPath);
+        return pathItemPaths.stream()
+                .flatMap(pathItemPath -> operationPaths.stream()
+                        .flatMap(operationPath -> getPathParameterTemplateName(pathItemPath, operationPath).stream()))
+                .collect(Collectors.toList());
```
MikeEdgar, line 148: This may need a `distinct()` before the `collect` since the `pathItemPath` will be processed for each entry in `operationPaths`.

```suggestion
        return pathItemPaths.stream()
                .flatMap(pathItemPath -> operationPaths.stream()
                        .flatMap(operationPath -> getPathParameterTemplateName(pathItemPath, operationPath).stream()))
                .distinct()
                .collect(Collectors.toList());
```

MikeEdgar (commented):
This is looking good, thanks @geniegeist . Could you have a look at my comment about the distinct path names? I think without that it has the potential to add duplicates in some cases.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
geniegeist: Thanks, I'm not sure what an alternative is. Would it make it sense to initialize an empty list instead?

```
    protected List<String> fullPathsOf(AnnotationTarget target) {
        List<String> pathSegments;

        switch (target.kind()) {
            case FIELD:
                pathSegments = pathsOf(target.asField().declaringClass());
                break;
            case METHOD:
                pathSegments = methodPaths(target.asMethod());
                break;
            case METHOD_PARAMETER:
                pathSegments = methodPaths(target.asMethodParameter().method());
                break;
            default:
                pathSegments = List.of("null");
                break;
        }

        return pathSegments.stream()
                .map(pathSegment -> contextPath + '/' + pathSegment)
                .collect(Collectors.toList());
    }
```

MikeEdgar: Probably, but I think the default case is effectively dead code anyway so it's probably fine as you have it for now.

geniegeist: I tried to make it less funky with the latest commit. Let me know what you think.

MikeEdgar: I think that's good, thanks. Ultimately it would probably just omit the `/` and `pathSegment` for the default case, but I doubt that's a real world possibility so we can't really test it anyway.

Thanks for the PR. I'll go ahead and merge once the CI passes and get this into a release.

MikeEdgar (approved):

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
geniegeist: Thank you, I'm happy to contribute :)

PR #2128, opened 2025-01-04
fix: use more accurate formats for byte, short, and char types

### LINKED ISSUE #2124: Generated results for Character Java type are invalid
We are finding that a Java Character field is mapped to OpenAPI via:

"type": "string",
"format": "byte",
"maxLength": 1.0,

We do not think 'byte' is the correct format nor is it documented in the 3.1.0 spec.  Also length types should be integers not floating point.  Also another reason byte does not make sense is a Character is one char or 2 bytes (or could be more under special cases)  in Java.

We think the correct OpenAPI for this would be:

"maxLength": 1,
"minLength": 0,
"type": "string",

Thanks,
-David

### LINKED ISSUE #2126: Generated OpenAPI for Java byte is incorrect
If we have Java code like this:

@JsonView(Views.Abridged.class)
@Schema(description = "Velocity of the sensor in the vertical direction, in decimeters per second.", minimum = "-128", maximum = "127", example="2")
private byte d17;

@JsonView(Views.Abridged.class)
@Schema(description = "Standard deviation of the estimate of the sensor track, in degrees.", minimum = "0", maximum = "45", example="2")
private byte d18;

Currently this generates this OpenAPI:

d17:
          type: string
          format: byte
          description: "Velocity of the sensor in the vertical direction, in decimeters\
            \ per second."
          maximum: 127
          minimum: -128
          example: "2"
d18:
          type: string
          format: byte
          description: "Standard deviation of the estimate of the sensor track, in\
            \ degrees."
          maximum: 45
          minimum: 0
          example: "2"

The best I can find from the OpenAPI spec is here: https://spec.openapis.org/registry/format/

Which leads me to think this should be defined in OpenAPI as:

"d17": {
                        "maximum": 127,
                        "minimum": -128,
                        "type": "number",
                        "description": "Velocity of the sensor in the vertical direction, in decimeters per second.",
                        "format": "int8",
                        "example": 2
},
"d18": {
                        "maximum": 45,
                        "minimum": 0,
                        "type": "number",
                        "description": "Standard deviation of the estimate of the sensor track, in degrees.",
                        "format": "int8",
                        "example": 2
},

or possibly this:

"d17": {
                        "maximum": 127,
                        "minimum": -128,
                        "type": "integer",
                        "description": "Velocity of the sensor in the vertical direction, in decimeters per second.",
                        "format": "int8",
                        "example": 2
},
"d18": {
                        "maximum": 45,
                        "minimum": 0,
                        "type": "integer",
                        "description": "Standard deviation of the estimate of the sensor track, in degrees.",
                        "format": "int8",
                        "example": 2
},

So the type would not be a string.  Also having a maximum & minimum string makes no sense.  Only when the type is a number or integer do those make sense.  

But the key is the "format": "int8" as that makes it clear this is a signed 8 bit number.

Thanks,
-David

MikeEdgar:
- `byte`: `format: int8`
- `short`: `format: int16`
- `char`: `format: char`

See the format registry for reference: https://spec.openapis.org/registry/format/

Fixes #2124 
Fixes #2126


MikeEdgar:
@dhoffer, FYI. Let me know if this change matches your expectations.

dhoffer:
Hi @MikeEdgar 

It still seems a little off:

@JsonView(Views.Abridged.class)
@Schema(description = "Relative velocity to skin line.", example = "1")
private short h29;

Results in:
"h29" : {
            "type" : "integer",
            "description" : "Relative velocity to skin line.",
            "example" : 1
          },

I think it should be:
"h29" : {
            "type" : "integer",
            "format" : "int16",
            "description" : "Relative velocity to skin line.",
            "example" : 1
          },

And 

@JsonView(Views.Abridged.class)
@Schema(description = "Computed object length based upon HRR profile, in meters.", minimum = "0", maximum = "100", example="22")
private byte h30;

Results with:
"h30" : {
            "type" : "string",
            "format" : "byte",
            "description" : "Computed object length based upon HRR profile, in meters.",
            "maximum" : 100,
            "minimum" : 0,
            "example" : "22"
          },

And we think it should be:
"h30" : {
            "type" : "integer",
            "format" : "int8",
            "description" : "Computed object length based upon HRR profile, in meters.",
            "maximum" : 100,
            "minimum" : 0,
            "example" : "22"
          },

byte, short, int, long are all integers with different size specified in the format.

I'm using these dependencies that I built locally.

<dependency>
                <groupId>io.smallrye</groupId>
                <artifactId>smallrye-open-api-core</artifactId>
                <version>4.0.6-SNAPSHOT</version>
</dependency>
 dependency>
                <groupId>io.smallrye</groupId>
                <artifactId>smallrye-open-api-jaxrs</artifactId>
                <version>4.0.6-SNAPSHOT</version>
</dependency>

MikeEdgar:
@dhoffer, the snippets you gave as your expected results are what I see locally. Did you build `main` or this PR for your test?

dhoffer:
Hi @MikeEdgar my bad, yes I was on main.  I guess I thought this was already merged to main but I see its not.  I don't seem to have access to pull down your PR branch.  Given your reply, I say this is ready for a merge.  I will test as soon as I have access.

phillip-kruger (approved):

dhoffer:
Hi @MikeEdgar I have confirmed the 4.0.6-SNAPSHOT code fixes this and the other related issue.  How soon could we get a release?

PR #2143, opened 2025-01-12
fix: prevent binding OAS parameter to unrelated resource method argument

MikeEdgar:
Relates to quarkusio/quarkus#45358


MikeEdgar:
This will conflict with #2138, so leaving it as a draft until that is merged.

phillip-kruger (approved):

PR #2152, opened 2025-01-14
Fix OpenAPI 3.0 output of nullable references

### LINKED ISSUE #2145: OpenAPI 3.0.3 Schema Incorrectly Includes enum: [null] When Description is Present in @Schema Annotation
### Description:

When generating OpenAPI 3.0.3 schema using SmallRye OpenAPI, if a field's @Schema annotation includes a description, the generator incorrectly adds enum: [null] to the schema for nullable fields. This doesn't occur when the description is omitted.

### Example code:

Code: 
```java
@Schema(description = "This is a test record for demonstrating the bug")
public record HelloWorld(
        @Schema(description = "If the field has description, then null is included",
                required = true,
                nullable = true)
        Subject incorrectOutput,
        @Schema(
                // no description, correct output
                required = true,
                nullable = true)
        Subject correctOutput) {}
```

Configuration: `application.properties` force OpenAPI to `3.0.3`
```properties
mp.openapi.extensions.smallrye.openapi=3.0.3
```

### Resulting Schema
```yaml
    HelloWorld:
      description: This is a test record for demonstrating the bug
      required:
      - incorrectOutput
      - correctOutput
      type: object
      properties:
        incorrectOutput:
          description: "If the field has description, then null is included"
          type: string
          anyOf:
          - $ref: "#/components/schemas/Subject"
          - enum:
            - null
          nullable: true
        correctOutput:
          allOf:
          - $ref: "#/components/schemas/Subject"
          nullable: true
```

### Issue:

* The incorrectOutput field (with description) incorrectly includes enum: [null]
* The correctOutput field (without description) generates the correct schema
* Both fields are configured identically except for the presence of a description

I'm using quarkus 3.17.5, which includes io.smallrye:smallrye-open-api-core:4.0.5.

Thank you for your help!

Azquelt:
When a nullable field or parameter references a non-nullable schema, we generate a schema for the field which uses anyOf to permit null. Any informational properties or additional assertions defined on the field are also put on the field schema and the type property is also retained.

When we were transforming this field schema for OpenAPI 3.0, we were only doing certain transformations if the type property is not present, which doesn't handle the case above.

Ensure the transformation to and from OpenAPI 3.0 schemas handles these nullable references which include the type property.

Fixes #2145

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/media/SchemaIO.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/media/SchemaIO.java
@@ -191,7 +191,7 @@ public class SchemaIO<V, A extends V, O extends V, AB, OB> extends MapModelIO<Sc
         }
 
         // Detect {$ref=....,nullable=true} and convert to anyOf[{$ref=...}, {type=null}]
-        if (schema.getRef() != null && schema.getType() == null && SchemaSupport.getNullable(schema) == Boolean.TRUE) {
+        if (schema.getRef() != null && SchemaSupport.getNullable(schema) == Boolean.TRUE) {
             List<Schema> newAnyOfSchemas = new ArrayList<>();
             newAnyOfSchemas.add(OASFactory.createSchema().ref(schema.getRef()));
             newAnyOfSchemas.add(SchemaSupport.nullSchema());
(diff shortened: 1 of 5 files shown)


Azquelt:
Waiting for #2151 to be merged to rebase this.

MikeEdgar:
Looks good, thanks @Azquelt

PR #2156, opened 2025-01-15
Permit Callback.pathItemRef to be any JSON path

### LINKED ISSUE #2155: `@Callback.pathItemRef` always has `#/components/pathItem/` prepended to the value
For most `ref` properties on OpenAPI annotations, we check to see if the value already starts with `#/components/itemtype/` and prepend the annotation value with this prefix if it does not.

However, the value of `@Callback.pathItemRef` has this prefix added unconditionally.

Azquelt:
Change processing of Callback.pathItemRef so that it doesn't always prefix the value with #/components/pathItems/ when adding the reference to the model.

Fixes #2155


MikeEdgar:
Ignore the deprecation issue for now. I'll come back and address those in the JAX-RS module tests separately.

PR #2159, opened 2025-01-15
fix: save parent field information to type resolver's target queue

### LINKED ISSUE #2148: Schema annotations lost when interface methods are implemented in a mapped super class
The following architecture results in schema annotations in the mapped super (parent) class being dropped in the openapi.json:

```
Interface          Parent Class (provides concrete implementation)
     \_______   _______/
             | |
           Child Class (implements interface, extends parent)
```

The parent (mapped super) class contains a concrete implementation of the interface but does not implement the interface directly. Instead a child class implements the interface and uses the parent's implementation to satisfy it. Under these conditions, the schema annotations are dropped.

A reproducer:
[reproducer.zip](https://github.com/user-attachments/files/18402820/reproducer.zip)

MikeEdgar:
- Save a discovered field to the `TypeResolver`'s `targets` queue. This queue will later be used to find the best `AnnotationTarget` that represents the property
- Save fields for previously-known properties that are private when `privatePropertiesEnable` is true

Fixes #2148

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
@@ -284,7 +284,15 @@ public class TypeResolver {
     }
 
     private void setField(FieldInfo field) {
+        if (this.field != null) {
+            targets.remove(this.field);
+        }
+
         this.field = field;
+
+        if (field != null) {
+            targets.add(field);
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


MikeEdgar:
@fastLaugh FYI. You can confirm the expected JSON matches your desired output from #2148

dhoffer:
What is the estimated timeline of a release with this fix?

MikeEdgar:
> What is the estimated timeline of a release with this fix?

Probably later next week.

phillip-kruger (approved):

PR #2161, opened 2025-01-17
Create properties for record accessors

### LINKED ISSUE #2157: Record support with scanning of private fields disabled
At the moment we sort of have support for records by virtue of them being classes with private fields.

However, if `private-properties.enable` is set to `false` (e.g. to match the default behaviour of JSON-B), we no longer generate any schema properties for a record.

The scanner ignores the record's fields because they're private and ignores the record's methods because they're not getters. (A record with a component named `foo` has an accessor method named `foo` rather than `getFoo`.)

The JSON-B spec doesn't mention records at all, though it seems like most of the implementations do support records and there's a PR open to specify support: jakartaee/jsonb-api#365

Azquelt:
Consider record accessor methods as similar to getters so that we create object properties for them.

Fixes #2157


MikeEdgar (approved):
This looks good, thanks @Azquelt . The test location is good for now. As you said, it can be adjusted when the project moves to Java 17.

PR #2167, opened 2025-01-21
fix: improve handling of nullable parameters

### LINKED ISSUE #2144: Nullable Enum query string param and kotlin generates incorrect OAS file
Using Quarkus, if I have a nullable query string param parsed as an enum, like [in this example](https://github.com/brice-laurencin/quarkus-tuto/blob/enum-queryparam-java/src/main/java/org/acme/GreetingResource.java), launching `./gradlew quarkusAppPartsBuild`, I can see the OAS file in `build/generated/resources` and the parameter is documented as:
```yaml
      - name: fred
        in: query
        schema:
          $ref: "#/components/schemas/Fred"
```
Which seems fine.

Using kotlin, as shown [here](https://github.com/brice-laurencin/quarkus-tuto/blob/enum-queryparam-kotlin/src/main/kotlin/org/acme/GreetingResource.kt), I can see two different behaviors whether the parameter is nullable or not.
If not nullable, I get the same schema as the java version.

But if the parameter is nullable, I get this:
```yaml
      - name: fred
        in: query
        schema:
          allOf:
          - $ref: "#/components/schemas/Fred"
          - nullable: true
```
Which seems incorrect.

I'm using quarkus 3.15.2 (LTS), packed with `io.smallrye:smallrye-open-api-core:3.10.0`.

I'm posting this here after a quick chat with @MikeEdgar on [Zulip](https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/OAS.20exporter).

Thanks for your help!

MikeEdgar:
Closes #2144

Since #2144 is using a slightly older pre-4.0 release, the current behavior is not exactly as described. 

With this change, given

```java
enum StatusEnum {
    VAL1,
    VAL2;
}

@jakarta.ws.rs.Path("/status")
static class FruitResource {
    @jakarta.ws.rs.GET
    public String get(@jakarta.ws.rs.QueryParam("status") @org.jetbrains.annotations.Nullable StatusEnum status) {
        return null;
    }
}
```

the expected `status` parameter schema would look like this for 3.0.x
```json
{
  "allOf" : [ {
    "$ref" : "#/components/schemas/StatusEnum"
  } ],
  "nullable" : true
}
```

or this for 3.1.x
```json
{
  "anyOf" : [ {
    "$ref" : "#/components/schemas/StatusEnum"
  }, {
    "type": "null"
  } ]
}
```

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/internal/models/media/SchemaSupport.java
+++ b/core/src/main/java/io/smallrye/openapi/internal/models/media/SchemaSupport.java
@@ -128,6 +128,13 @@ public final class SchemaSupport {
         setTypesRetainingNull(observer, getTypeList(observed));
     }
 
+    public static Boolean isEmpty(Schema schema) {
+        if (schema instanceof BaseModel) {
+            return ((BaseModel<?>) schema).getAllProperties().isEmpty();
+        }
+        return null;
+    }
+
     static void notifyTypeObservers(Schema observed, Consumer<Schema> observerAction) {
         List<Schema> typeObservers = Extensions.getTypeObservers(observed);
 
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
@@ -716,9 +716,21 @@ public abstract class AbstractParameterProcessor {
         setDefaultValue(localSchema, context.defaultValue);
 
         if (localOnlySchemaModified(paramSchema, localSchema, modCount)) {
-            // Add new `allOf` schema, erasing `type` derived above from the local schema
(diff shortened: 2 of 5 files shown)


MikeEdgar:
@brice-laurencin, FYI. Please comment if you have any concern with the expected results.

phillip-kruger (approved):

brice-laurencin:
> @brice-laurencin, FYI. Please comment if you have any concern with the expected results.

Thanks a lot @MikeEdgar, these schemas please the swagger parser lib, hence the generator.
I now have proper client generation with such nullable enum as path param.

Thanks again!

brice-laurencin:
Hello again @MikeEdgar, would you have any insight on whether Quarkus 3.19 (and then 3.20LTS) will rely on this new `4.0.7` release of `smallrye-openapi`?

MikeEdgar:
@brice-laurencin the PR to update Quarkus to `4.0.7` is included in the 3.19 milestone: https://github.com/quarkusio/quarkus/pull/45802

PR #2182, opened 2025-01-29
fix: use UTF-8 for InputStreamReaders relying on platform encoding

MikeEdgar:
Make sure readers do not rely on platform default encoding when reading input streams. This will be back-ported to 4.0.x also.

For quarkusio/quarkus#44569


gsmet (approved):
Awesome, thanks!

I wonder if we should also push it in a micro release compatible with Quarkus 3.15 (which uses 3.10: https://github.com/quarkusio/quarkus/blob/3.15/bom/application/pom.xml#L59).

MikeEdgar:
This bug didn't come until 3.11, so Quarkus 3.15 isn't affected :tada:

gsmet:
Thanks for checking!

phillip-kruger (approved):

PR #2196, opened 2025-02-12
fix: restore core dependency on model-apt for correct build order

MikeEdgar:
Restore `core` dependency on `model-apt` module, removed in #2179 . The build failure was masked by the presence of model-apt in the GitHub Maven cache, which is now missing.


MikeEdgar:
Going ahead to merge this one, the build will fail without it.

PR #1360, opened 2023-01-18
Represent Java BaseStream<T, S> as array of T in generated schema

MikeEdgar:
Fixes #1359

Signed-off-by: Michael Edgar <michael@xlate.io>

PR #2208, opened 2025-02-21
Add test for schema creation for custom collection classes

Azquelt:
These examples work fine in the current version but were broken in previous versions and I couldn't see a test for them.

I think this function was implemented in 3.3.0 with #1360.


MikeEdgar (approved):
Thanks @Azquelt

PR #2246, opened 2025-04-17
Release 4.0.10

MikeEdgar:


Ladicek:
I see you bumped to Jandex 3.3.0 (#2239). Note that this Jandex version has a newer index format version, so if this project includes a persistent Jandex index (`META-INF/jandex.idx`), it won't be readable by older Jandex versions. That might be OK, I don't really know, but it needs to be considered carefully.

MikeEdgar:
@Ladicek the Jandex bump is on `main`, this release is from the `4.0.x` branch. As far as the bump on `main`, that's the release for the next MicroProfile spec update. Are you intending to update Jandex on the downstream runtimes in the near future?

Ladicek:
Ah, I didn't notice that. All should be good then!

As for Jandex 3.3 on the downstream runtimes, Quarkus has Jandex 3.3 since 3.21.2. WildFly didn't update yet (and I don't see a JIRA for that right now).

MikeEdgar:
Thanks! I think we should be fine then. Either way there are no indexes being bundled with any of the artifacts in this project.

PR #2247, opened 2025-04-21
Add ability to produce an unmodifiable model using builder

MikeEdgar:


phillip-kruger (approved):
LGTM, but why do we need this ?

MikeEdgar:
@phillip-kruger , there is a proposed Quarkus extension https://github.com/quarkusio/quarkus/issues/47236 where the extension would push the OpenAPI to Apicurio at Quarkus startup time. We were thinking this should be done only after the user's OASFilters have executed, so an option to make the model available may be as a CDI bean of type `SmallRyeOpenAPI` [1], which includes the MP OpenAPI model and the Jackson model (for string serialization). To support that case, allowing an immutable version of that model would prevent user code from modifying the MP model while the Jackson model is already inaccessible in `SmallRyeOpenAPI`.

Of course, I'm open to other ideas/feedback on that approach. Thoughts?

[1] https://github.com/smallrye/smallrye-open-api/blob/b22b49c76be25f9e1de84ecf4d4000ac4a717c1a/core/src/main/java/io/smallrye/openapi/api/SmallRyeOpenAPI.java

phillip-kruger:
Ok thanks! This sounds good to me, I think having it available in a cdi bean will be useful in other places too

MikeEdgar:
@Azquelt are you interested in taking a look at this as well?

Azquelt (approved):
It looks sensible to me too.

PR #2248, opened 2025-04-21
Add option to disable sorted parameters

### LINKED ISSUE #1957: Maven plugin should be configurable regarding the order of the parameters in a REST endpoint

MikeEdgar:
Closes #1957 

This will also be cherry-picked for 4.0.11

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -95,3 +95,11 @@ Set this value in order to change the maximum threshold for processed static fil
 mp.openapi.extensions.smallrye.merge-schema-examples
 ----
 Set this boolean value to disable the merging of the deprecated `@Schema` `example` property into the `examples` array introduced in OAS 3.1.0. If not set, it will default to `true` the deprecated `example` will be mapped to the `examples` array in the OpenAPI model.
+
+* Sorted Parameters
++
+[source%nowrap]
+----
+mp.openapi.extensions.smallrye.sorted-parameters.enable
+----
+Set this boolean value to enable or disable the sorting of parameter array entries during annotation scanning. When enabled (set to `true`), parameters will be order either by their order within a `@Parameters` annotation on an operation method or (in the absence of that annotation) by their `$ref`, `in`, and `name` attributes. When disabled (set to `false`), parameters will be in the order they are encountered in the Java code.
--- a/README.adoc
+++ b/README.adoc
@@ -102,4 +102,4 @@ Set this boolean value to disable the merging of the deprecated `@Schema` `examp
 ----
 mp.openapi.extensions.smallrye.sorted-parameters.enable
 ----
-Set this boolean value to enable or disable the sorting of parameter array entries during annotation scanning. When enabled (set to `true`), parameters will be order either by their order within a `@Parameters` annotation on an operation method or (in the absence of that annotation) by their `$ref`, `in`, and `name` attributes. When disabled (set to `false`), parameters will be in the order they are encountered in the Java code.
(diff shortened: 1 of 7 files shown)


README.adoc
```diff
@@ -95,3 +95,11 @@ Set this value in order to change the maximum threshold for processed static fil
...
 ----
 Set this boolean value to disable the merging of the deprecated `@Schema` `example` property into the `examples` array introduced in OAS 3.1.0. If not set, it will default to `true` the deprecated `example` will be mapped to the `examples` array in the OpenAPI model.
+
+* Sorted Parameters
++
+[source%nowrap]
+----
+mp.openapi.extensions.smallrye.sorted-parameters.enable
+----
+Set this boolean value to enable or disable the sorting of parameter array entries during annotation scanning. When enabled (set to `true`), parameters will be order either by their order within a `@Parameters` annotation on an operation method or (in the absence of that annotation) by their `$ref`, `in`, and `name` attributes. When disabled (set to `false`), parameters will be in the order they are encountered in the Java code.
```
Azquelt, line 105: Can we include what the default value is here?

Azquelt (changes requested):
I think the doc should say what the default value of the new option is.

I'm less sure about my other comment, it could just be that I've misunderstood what we're doing there.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
```diff
@@ -482,18 +491,20 @@ boolean templateParameterPatternEligible(Parameter param) {
...
-                .filter(Objects::nonNull)
-                .sorted(ResourceParameters.parameterComparator(preferredOrder))
-                .collect(Collectors.toList());
+                .filter(Objects::nonNull);
+
+        if (scannerContext.getConfig().sortedParametersEnable()) {
+            parameterStream = parameterStream.sorted(ResourceParameters.parameterComparator(preferredOrder));
+        }
+
+        List<Parameter> parameters = parameterStream.collect(Collectors.toList());
```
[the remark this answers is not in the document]
MikeEdgar: Yes it does seem to be redundant. I was just replacing the existing sorting, but I'll see if we can just remove the sort here.

README.adoc
MikeEdgar: Added default value to the README documentation.

Azquelt (approved):

PR #2256, opened 2025-05-02
fix: use array item schema ref when available instead of inline schema

MikeEdgar:
Resolves problem described in quarkusio/quarkus#47653

This will also be cherry-picked for 4.0.11

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -82,7 +82,7 @@
         <dependency>
             <groupId>io.xlate</groupId>
             <artifactId>yaml-json</artifactId>
-            <version>0.2.0</version>
+            <version>0.2.1</version>
             <optional>true</optional>
         </dependency>
 


Azquelt:
From a brief glance this looks right, but I'll have a closer look at it on Monday.

core/src/main/java/io/smallrye/openapi/runtime/io/schema/SchemaFactory.java
```diff
@@ -340,7 +340,7 @@ public static Schema includeTypeSchema(AnnotationScannerContext context, Schema
         if (schema.getType() != null && schema.getType().contains(Schema.SchemaType.ARRAY) && implSchema != null) {
             // If the @Schema annotation indicates an array type, then use the Schema
             // generated from the implementation Class as the "items" for the array.
-            schema.setItems(implSchema);
+            schema.setItems(lookupRef(context, type, implSchema));
```
Azquelt, line 343: This change seems contradictory to line 337, where we call `implSchema = readClassSchema(context, type, false);` and `false` here indicates that we don't want a reference.

Similarly, I think passing `false` there will prevent `readClassSchema` from adding the schema to the registry in the first place. Where do we expect the item schema to be added to the registry?

If we're expecting the item schema to be added to the registry some time before this, is it guaranteed that the item schema is processed before this schema?

I note that `readClassSchema` also calls `setItems` - does that also need to look up references?

MikeEdgar: The flag passed to allow schema registration or not will only enable/disable returning a reference schema. The call to `OpenApiDataObjectScanner#process` may still register the schema, but that method will not return the reference itself. It is up to the SchemaRegistry to do that by possibly calling `SchemaRegistry#schemaRegistration`.

> If we're expecting the item schema to be added to the registry some time before this, is it guaranteed that the item schema is processed before this schema?

Yes, that should occur within `OpenApiDataObjectScanner` for `ClassType`s.

> I note that readClassSchema also calls setItems - does that also need to look up references?

The last commit also updates the two other `setItems` locations in `SchemaFactory` and adjusts the tests accordingly.

core/src/main/java/io/smallrye/openapi/runtime/io/schema/SchemaFactory.java
```diff
@@ -309,43 +309,46 @@ SchemaFactory.<Type[], List<Schema>> readAttr(context, annotation, SchemaConstan
...
+                    // generated from the implementation Class as the "items" for the array.
+                    schema.setItems(lookupRef(context, type, implSchema));
+                } else {
+                    // If there is an impl class - merge the @Schema properties *onto* the schema
+                    // generated from the Class so that the annotation properties override the class
+                    // properties (as required by the MP+OAI spec).
+                    schema = MergeUtil.mergeObjects(implSchema, schema);
+                }
+            }
         }
```
Azquelt, line 352: If I'm understanding it correctly, the addition of the `declaredType` parameter here is to change how we interpret `implementation`?

If the user puts `@Schema(type=Array, implementation=Foo.class)`, we treat `Foo` as the item type, whereas if they put `@Schema(implementation=Bar.class)`, we treat `Bar` as the type itself?

So for something like this:
```java
@Schema(implementation=Bar.class)
private List<Foo> listOfFoo;
```
we'd ignore that the field is a `List<Foo>` and instead derive the schema from `Bar`?

I'm guessing your test in `StandaloneSchemaScanTest` with `@Schema(implementation=Greeting[].class)` is related to this.

I'm wondering whether there are situations where this is going to change what we produce? (other than generating a reference instead of a schema in some cases)

MikeEdgar: > If the user puts `@Schema(type=Array, implementation=Foo.class)`, we treat `Foo` as the item type, whereas if they put `@Schema(implementation=Bar.class)`, we treat `Bar` as the type itself?

That's right. In fact, that simplest case would have been handled earlier in the `else if` that precedes the `else` where the `includeTypeSchema` method is called. Here, we handle it where they have added more attributes, e.g. `description`. Similarly for the case where there is only an `implementation`.

In summary, this should make the following equivalent:

```java
@Schema(implementation = Foo[].class)
private List<Foo> listOfFoo;

@Schema(implementation = Foo[].class, description = "My Foo list")
private List<Foo> listOfFoo;

@Schema(type = SchemaType.ARRAY, implementation = Foo.class)
private List<Foo> listOfFoo;

@Schema(type = SchemaType.ARRAY, implementation = Foo.class, description = "My Foo list")
private List<Foo> listOfFoo;
```

I'll update the test to make it clear.

Azquelt (approved):

PR #2259, opened 2025-05-06
Add a test for nested Kotlin collections

Azquelt:
Testing the scenario from microprofile/microprofile-open-api#688

Seems to work as expected.


MikeEdgar (approved):
Thank you, I'll merge this and ask the user on the MP issue to open something in Quarkus with a reproducer.

PR #2312, opened 2025-07-16
fix: recursively resolve parameterized types from type variables

### LINKED ISSUE #2311: Parameterized generic types not resolved
When a generic type variable resolves/maps to a parameterized type containing the variable, type resolution does not further resolve that parameterized type and the resulting schema is missing information.

In the following example, the type of `T` in `GenericRoot` should be `List<Bean>` for the `DataList` class. However, the resulting schema further below can be seen to have `items` of type `object`. The `items` for this schema should be a reference to the `Bean` schema, similar to the singular `Data` schema.

```java
class Bean {
    String name;
    int age;
}

class GenericRoot<T> {
    T data;
}

class GenericListRoot<T> extends GenericRoot<List<T>> {
}

class Data extends GenericRoot<Bean> {
}

class DataList extends GenericListRoot<Bean> {
}
```

Incorrect schema:
```json
{
  "openapi" : "3.1.0",
  "components" : {
    "schemas" : {
      "Bean" : {
        "type" : "object",
        "properties" : {
          "name" : {
            "type" : "string"
          },
          "age" : {
            "type" : "integer",
            "format" : "int32"
          }
        }
      },
      "Data" : {
        "type" : "object",
        "properties" : {
          "data" : {
            "$ref" : "#/components/schemas/Bean"
          }
        }
      },
      "DataList" : {
        "type" : "object",
        "properties" : {
          "data" : {
            "type" : "array",
            "items" : {
              "type" : "object"
            }
          }
        }
      }
    }
  }
}
```

Expected schema:
```json
{
  "openapi" : "3.1.0",
  "components" : {
    "schemas" : {
      "Bean" : {
        "type" : "object",
        "properties" : {
          "name" : {
            "type" : "string"
          },
          "age" : {
            "type" : "integer",
            "format" : "int32"
          }
        }
      },
      "Data" : {
        "type" : "object",
        "properties" : {
          "data" : {
            "$ref" : "#/components/schemas/Bean"
          }
        }
      },
      "DataList" : {
        "type" : "object",
        "properties" : {
          "data" : {
            "type" : "array",
            "items" : {
              "$ref" : "#/components/schemas/Bean"
            }
          }
        }
      }
    }
  }
}
```

MikeEdgar:
Fixes #2311 

This change refactors how generic type resolution handles variables that resolve to a parameterized type that itself contains a type variable. For example `MyType<T> extends AnotherType<List<T>>`. 

Previously, the resolved type of `T` would have been "stuck" as `List<T>` which then makes the schema of `T` simply `type: object` in the resulting OpenAPI. The fix here will allow `List<T>`  or any of a parameterized type's arguments to further be resolved using the type resolution variable mappings.

Also for backport to `4.0.x`

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java
@@ -11,6 +11,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Deque;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -354,7 +355,7 @@ public class TypeResolver {
      * @return the resolved type (if found)
      */
     public Type resolveType() {
-        return getResolvedType(leaf);
+        return resolveType(leaf, resolutionStack);
     }
 
     public boolean isIgnored() {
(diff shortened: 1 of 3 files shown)


Azquelt:
I've started looking at this, but it'll take me a bit longer to get my head round it and understand why the changes you've made are needed.

MikeEdgar:
Thank you. 

Taking the example of `MyType<T> extends AnotherType<List<T>>`, the previous version of `getResolvedType` was able to resolve a variable `T` to `List<T>`, but since there was no further handling for parameterized types, the resulting item type for the schema's array is not determined. The changes allow for a more recursive approach between type variables and parameterized types.

https://github.com/smallrye/smallrye-open-api/blob/f7267f8be23709f7161c00083dbb42129b9c9d85/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/TypeResolver.java#L394-L417

Azquelt (approved):
I'm happy to approve this since it definitely fixes some cases when compared to the old code, however there are a couple of extra cases which might need some extra thought.

1. Should we attempt to resolve wildcards in parameterized types? E.g. `class A<T> extends B<List<? extends T>>`

2. The resolution stack is a linear list and it looks like we put interfaces in the list along with classes. Is it possible to get a name clash because an interface has a type variable with the same name?

e.g.

```java
        interface Extra<T> {}

        class GenericRoot<T> implements Extra<String> {
            T data;
        }

        @Schema(name = "Data")
        class Data extends GenericRoot<Bean> {
        }
```

The `T` in `T data` should be resolved to `Bean`, but could we also have a mapping of `T -> String` in the resolution stack?

I expanded your test to check this and I found that it doesn't break, because in `TypeResolver.getAllFields` we add mappings for the current class, then scan fields, then scan methods, then add mappings for any interfaces, so when we scan the fields we haven't added the interfaces to the stack yet.

As long as that's the intended behaviour I think we're ok.

MikeEdgar:
> 1. Should we attempt to resolve wildcards in parameterized types? E.g. `class A<T> extends B<List<? extends T>>`

I agree this needs to be considered. I actually made that change initially, but wasn't prepared to investigate the resulting changes in several tests depending on the current behavior. There would be some needed change to the `SchemaRegistry` as well because a type could be registered twice - once keyed by a `Type` with the wildcards, and a second time with the wildcards resolved.

> 2. The resolution stack is a linear list and it looks like we put interfaces in the list along with classes. Is it possible to get a name clash because an interface has a type variable with the same name?

This has occurred to me as well, and there are possibly some edge cases that could have an issue. Generally, it works for most scenarios and I don't recall there ever being bug reports on this part.

PR #2324, opened 2025-07-26
Include content and media types w/o schema for opaque response types

### LINKED ISSUE #2315: @Produces annotation does not seems to be used
Hi,

I have an endpoint defined like that :

```java
  @GET
  @Path("/myPath")
  @Produces(MediaType.APPLICATION_OCTET_STREAM)
  Response getOctetStream()
```

this will generate something like that :

```yaml
responses:
        "200":
          description: OK
```

I was using swagger before and I got something like :
```yaml
responses:
        default:
          description: default response
          content:
            application/octet-stream: {}
```

Which seems more coherent with the API.

Is the "@Produces" annotation parsed to define the response type ?

MikeEdgar:
Also add new configuration property `mp.openapi.extensions.smallrye.default-with-internal-response` (name up for discussion) to use `default` response entry in these cases.

To-do:
- add docs to README for new property
- resolve TCK failure due to (incorrect?) assumption by TCK that a media type always has a schema. This does not appear to be required by the OAS spec.

Fixes #2315

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
@@ -466,87 +466,117 @@ public interface AnnotationScanner {
             final MethodInfo method,
             Operation operation) {
 
-        Type returnType = context.getResourceTypeResolver().resolve(method.returnType());
-        APIResponse response = null;
+        final Type returnType = context.getResourceTypeResolver().resolve(method.returnType());
+        final boolean internalResponse = isScannerInternalResponse(returnType, context, method);
         final int status = getDefaultStatus(method);
-        final String code = String.valueOf(status);
-        final String description = getReasonPhrase(status);
+        final String code;
+        final String description;
 
-        if (isVoidResponse(method)) {
-            if (generateResponse(code, operation)) {
-                response = OASFactory.createAPIResponse().description(description);
-            }
-        } else if (generateResponse(code, operation)) {
-            response = OASFactory.createAPIResponse().description(description);
-
-            /*
-             * Only generate content if not already supplied in annotations and the
-             * method does not return an opaque Scanner Response
-             */
-            if (!isScannerInternalResponse(returnType, context, method) &&
-                    (ModelUtil.responses(operation).getAPIResponse(code) == null ||
-                            ModelUtil.responses(operation).getAPIResponse(code).getContent() == null)) {
-
-                Schema schema;
-
-                if (isMultipartOutput(returnType)) {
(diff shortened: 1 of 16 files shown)


core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
```diff
@@ -466,87 +466,126 @@ default void createResponseFromRestMethod(final AnnotationScannerContext context
...
+            mt.setSchema(schema);
+            content.addMediaType(producesType, mt);
+        }
+
+        return content;
+    }
+
+    default void maybeUpdateOptionalResponseSchema(Type returnType, Schema schema) {
+        if (schema != null && SchemaSupport.getNullable(schema) == null
+                && TypeUtil.isOptional(returnType)) {
```
Azquelt, line 571: It's not obvious what this method does from its name alone.

Would it be clearer if the `if` block here was removed, and the method was named `makeSchemaOptional`? The caller would then need to do the check in the `if` block before calling it.

Otherwise I think we should add doc here to say what this method does.

Azquelt (commented):
There are really three changes here:

1. When we don't have any type information because a method returns `Response`, we should still take into account any annotations which set the response content type.
2. Add an option to use `default` as the response code, rather than `200` when a method returns `Response` and we don't know what the response code is.
3. Automatically add an empty schema when we generate an `application/json` response content as a workaround to a TCK issue.

I've made one comment where I think the code could be clearer, but otherwise the changes look good to me.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AnnotationScanner.java
```diff
@@ -466,87 +466,126 @@ default void createResponseFromRestMethod(final AnnotationScannerContext context
...
+            code = "default";
+            description = "Default Response";
+        } else {
+            code = String.valueOf(status);
+            description = getReasonPhrase(status);
+        }
+
+        if (!generateResponse(code, operation)) {
+            return;
+        }
```
Azquelt, line 485: It occurred to me as a potential problem that if an operation had a `200` response specified for the operation already (e.g. from a static file), and we have the new `generic-response-use-default` option set, would this result in us generating a `default` response in addition to the existing `200` response?

We would call `generateResponse("default", operation)`, which would return `true` because there's no `default` response specified yet.

MikeEdgar:
Thanks for reviewing, @Azquelt . I began to look into splitting it into separate PRs for each of the individual changes, and given the closely-related nature of these, I don't really think it's worth the effort. I'll go ahead and merge this as-is.

PR #2332, opened 2025-08-07
fix: ignore duplicated parameters on paths and operations

### LINKED ISSUE #2325: Duplicated query parameters when using the @Parameter annotation on a @BeanParam
With `smallrye-open-api-core` v4.0.11 along with `io.quarkus:quarkus-smallrye-openapi-spi` v3.24.2, when annotating a web service with the `@BeanParam` annotation pointing to a Java record resorting to the `@org.eclipse.microprofile.openapi.annotations.parameters.Parameter`, the generated JSON specifications declares twice the corresponding parameter.

To reproduce the issue, here is a reference example:
```
package com.corp;

import jakarta.enterprise.context.ApplicationScoped;
import jakarta.transaction.Transactional;
import jakarta.validation.Valid;
import jakarta.validation.constraints.NotNull;
import jakarta.ws.rs.*;
import jakarta.ws.rs.core.MediaType;
import org.eclipse.microprofile.openapi.annotations.Operation;
import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;

@SuppressWarnings("unused")
@Path("/test")
@ApplicationScoped
@Consumes(MediaType.APPLICATION_JSON)
public final class TestResource
{

    public record Paging(
            @NotNull @QueryParam("skip") @Parameter(description = "The number of items to skip", required = true) Integer skip,
            @NotNull @QueryParam("take") @Parameter(description = "The number of items to retrieve", required = true) Integer take)
    {

    }

    @GET
    @Path("showcase")
    @Operation(summary = "Showcases an issue")
    @Transactional()
    public void showcase(@Valid @BeanParam Paging paging)
    {

    }

}
```

In that case, the generated, within the OpenAPI `paths` property:
```
    "/test/showcase" : {
      "get" : {
        "summary" : "Showcases an issue",
        "operationId" : "test_showcase",
        "parameters" : [ {
          "description" : "The number of items to skip",
          "required" : true,
          "name" : "skip",
          "in" : "query",
          "schema" : {
            "type" : "integer",
            "format" : "int32"
          }
        }, {
          "description" : "The number of items to retrieve",
          "required" : true,
          "name" : "take",
          "in" : "query",
          "schema" : {
            "type" : "integer",
            "format" : "int32"
          }
        }, {
          "description" : "The number of items to skip",
          "required" : true,
          "name" : "skip",
          "in" : "query",
          "schema" : {
            "type" : "integer",
            "format" : "int32"
          }
        }, {
          "description" : "The number of items to retrieve",
          "required" : true,
          "name" : "take",
          "in" : "query",
          "schema" : {
            "type" : "integer",
            "format" : "int32"
          }
        } ],
        "responses" : {
          "204" : {
            "description" : "No Content"
          }
        },
        "tags" : [ "test" ]
      }
    }
```

As soon as you remove the `@Parameter` annotation from the annotated `Paging` record, the duplication issue disappears.

MikeEdgar:
Fixes #2325

Also re-factor several occurrences of stream processing to simple loops to improve ease of debugging.

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
@@ -17,7 +17,10 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -202,7 +205,7 @@ public abstract class AbstractParameterProcessor {
 
         if (value != null) {
             valueString = value.asString();
-            if (valueString.length() > 0) {
+            if (!valueString.isEmpty()) {
                 return valueString;
             }
         }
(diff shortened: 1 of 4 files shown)


phillip-kruger (approved):

MikeEdgar (commented):
@Azquelt, please let me know if there's anything concerning here from your perspective. Thanks

Azquelt:
Did you work out why we get the duplicates in the first place? I note that in the issue it sounded like the problem was with records specifically, but in the test you're using a class that looks like a record?

I remember we added some changes for records in #2157 so that we pick up the properties even if you don't consider private fields of classes, but based on the tests in this change it sounds like that wasn't the problem here?

MikeEdgar:
> Did you work out why we get the duplicates in the first place? I note that in the issue it sounded like the problem was with records specifically, but in the test you're using a class that looks like a record?

The code that attempts to detect duplicates during the scan requires two annotations to have the same target for a match [1]. Failing that, they need to have a name/style match, which simple `@Parameter` annotations just for a `description` do not have (both name and style are null). In the case of a record, we get the param annotations from the constructor method parameters and the fields (hence two of them). The accessor method is not considered by the scanner, but is in the test to accurately represent a record.

[1] https://github.com/smallrye/smallrye-open-api/blob/ca394e6a2e9f7bfeff99bad2c70b877f1b7a65ce/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java#L1359

Azquelt:
Thanks, that makes sense. This is probably ok then, but maybe there's some cleanup to be done in future so that we're not trying to detect duplicates in two places.

Azquelt (approved):

Issue #2269, opened 2025-05-16
Kotlin Serialization Configurations Not Reflected in OpenAPI Specification


m4rcingsxr:
## Issue Description

Currently, changes made to Kotlin serialization configurations, such as setting a naming strategy to `snake_case` (e.g., via `JsonNamingStrategy.SnakeCase`), are not reflected in the generated OpenAPI specification when using SmallRye OpenAPI. This contrasts with Jackson configurations, where naming strategy changes (e.g., `PropertyNamingStrategy.SNAKE_CASE`) do propagate to the OpenAPI spec. For projects avoiding Jackson in favor of Kotlin serialization, this lack of support creates inconsistencies and fails to meet certain validation requirements, such as those enforced by the IBM API validator.

## Main Goals

To ensure compatibility with Kotlin serialization and compliance with validators like IBM's, we need the following features:

1. **Kotlin Serialization Reflection in OpenAPI**: Annotations (e.g., `@SerialName`) and custom serializers (e.g., for enums) defined in Kotlin serialization should be reflected in the generated OpenAPI specification.
2. **Global Naming Strategy Support**: Setting a global naming strategy (e.g., `snake_case`) in Quarkus configuration (e.g., `quarkus.kotlinx.serialization.naming-strategy=SnakeCase`) should apply to the OpenAPI spec.
3. **Enum Serialization in Snake Case**: Enums should support `snake_case` naming in the OpenAPI spec, either via custom serializers or a naming strategy, to pass IBM validator requirements.

## Current Behavior

- When using Kotlin serialization with `Json { namingStrategy = JsonNamingStrategy.SnakeCase }`, the OpenAPI spec still uses `camelCase` for property names.
- Explicit annotations like `@SerialName("snake_case")` or custom serializers for enums are not honored in the OpenAPI output.
- Jackson-based configurations (e.g., `ObjectMapper` with `PropertyNamingStrategy.SNAKE_CASE`) correctly influence the OpenAPI spec, but we aim to avoid Jackson dependencies.
- This results in validation failures with tools like the IBM API validator, which require `snake_case` for properties and enums.

## Expected Behavior

- Kotlin serialization configurations, including `JsonNamingStrategy.SnakeCase` and annotations like `@SerialName`, should be respected in the generated OpenAPI spec.
- A global configuration option (e.g., in Quarkus) should allow setting a naming strategy (e.g., `snake_case`) that applies to all schemas in the OpenAPI output.
[...]

PR #2343, opened 2025-08-27
Add support for kotlinx.serialization

Nillerr:
Implements initial support for [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization), as requested in https://github.com/smallrye/smallrye-open-api/issues/2269.

This PR does not address all the concerns of the linked issue, but focuses on supporting the `@SerialName` and `@Required` annotations.

With this PR, the `@SerialName` annotation can now be used to specify the name, much like how `@JsonProperty` from Jackson can be used for the same purpose.

Additionally, I added the same logic for `@Required` as is present for Jackson's `@JsonProperty(required = true)`.

To support this, I added the Kotlin Maven Plugin to the test scope in order to define a Kotlin class.

### DIFF
--- a/testsuite/data/src/test/resources/io/smallrye/openapi/testdata/components.schemas.kotlin-value-class-propname.json
+++ b/testsuite/data/src/test/resources/io/smallrye/openapi/testdata/components.schemas.kotlin-value-class-propname.json
@@ -7,7 +7,7 @@
           "description",
           "nestedCollection",
           "actualCustomName",
-          "requiredValue"
+          "requiredNullableValue"
         ],
         "type": "object",
         "properties": {
@@ -35,8 +35,8 @@
           "actualCustomName": {
             "type": "string"
           },
-          "requiredValue": {
-            "type": "string"
+          "requiredNullableValue": {
+            "type": ["string", "null"]
           }
         }
       }
--- a/testsuite/data/src/main/kotlin/io/smallrye/openapi/testdata/kotlin/KotlinBean.kt
+++ b/testsuite/data/src/main/kotlin/io/smallrye/openapi/testdata/kotlin/KotlinBean.kt
@@ -15,7 +15,7 @@ data class KotlinBean (
     @SerialName("actualCustomName")
     val customName: String = "",
     @Required
-    val requiredValue: String = ""
+    val requiredNullableValue: String? = null,
 )
 
 @Deprecated("1.0.0")


core/src/main/java/io/smallrye/openapi/runtime/util/KotlinUtil.java
```diff
@@ -0,0 +1,53 @@
...
+        // Kotlin generates a synthetic method for each property declaring any `AnnotationTarget.PROPERTY` annotations.
+        // The name of the property is the getter-name of the field followed by `$annotations`.
+        return getGetterName(field) + "$annotations";
+    }
+
+    private static Optional<MethodInfo> getSyntheticPropertyAnnotationsMethod(FieldInfo field) {
+        var methodName = getSyntheticPropertyAnnotationsMethodName(field);
+        var clazz = field.declaringClass();
+        return clazz.methods().stream()
+                .filter(m -> Modifiers.isSynthetic(m.flags()) && methodName.equals(m.name()))
```
MikeEdgar, line 37: Use `MethodInfo#isSynthetic` and the `Modifiers` util class can be removed too.

```suggestion
                .filter(m -> m.isSynthetic() && methodName.equals(m.name()))
```

MikeEdgar (commented):
I'm generally OK with this feature, but I'm not thrilled about adding Kotlin compilation to the core module just for the tests. In the past, we've put Kotlin code into the testsuite/data module, which is a Quarkus application.

Nillerr:
> I'm generally OK with this feature, but I'm not thrilled about adding Kotlin compilation to the core module just for the tests. In the past, we've put Kotlin code into the testsuite/data module, which is a Quarkus application.

Good point! In my haste, I didn't notice you already had a Kotlin dependency and testing setup in there. I have moved the Kotlin test logic to the `testsuite/data` module.

MikeEdgar (dismissed):
Thank you for the PR @Nillerr . This looks good to me. If no other reviews have any concerns we can go ahead and merge this for 4.1.x

phillip-kruger (approved):

testsuite/data/src/test/resources/io/smallrye/openapi/testdata/components.schemas.kotlin-value-class-propname.json
```diff
@@ -5,7 +5,9 @@
       "KotlinBean": {
         "required": [
           "description",
-          "nestedCollection"
+          "nestedCollection",
+          "actualCustomName",
+          "requiredValue"
         ],
```
Azquelt, line 11: Given that `description`, `nestedCollection` and `customName` are required without an annotation, does the `@Required` annotation on `requiredValue` actually do anything here?

Nillerr: Good eye! You're right. Because of the generated `@NotNull` annotation for a non-nullable value, and the lack of processing of default values for `@Serializable` classes, in this case the `@Required` annotation actually doesn't change the outcome of the schema.

I have replaced this property with a `requiredNullableValue` property instead, where the `@Required` annotation correctly alters the required-ness of the property (that would otherwise be optional).

Azquelt (approved):

PR #2372, opened 2025-09-30
Use jdk11 profile for CI with older dependencies, update build matrices

MikeEdgar:
- Update JDK 11 Maven profile to use older Java 11 compatible library versions. As libraries are upgraded to new versions supporting Java 17 and beyond, they will be added here so that we can still run the TCK in a Java 11 environment.
- Run each supported version of the TCK with Java 11, 17, and 21 and attach/publish reports to releases.
- Delete plugin management for unused `asciidoctor-maven-plugin`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,6 +105,9 @@ jobs:
         - 11
         - 17
         - 21
+        tck-version:
+        - "4.0.2"
+        - "4.1"
 
         include:
         # Disable older TCK jobs prior to 4.0 GA release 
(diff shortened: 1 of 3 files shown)


phillip-kruger (approved):

MikeEdgar:
@Azquelt let me know if anything with this change could be a problem. As mentioned, you would need to make sure that `smallrye-common-classloader` is kept at version 2.8.0 or earlier if running on Java 11.

Azquelt:
All looks good to me :+1:

PR #2375, opened 2025-10-01
Enable YAML alias expansion for Jackson IO

### LINKED ISSUE #1237: Support YAML anchors for static files
OpenApiProcessor uses Jackson, which does not support YAML anchors.

([OpenAPITools sample](https://github.com/OpenAPITools/openapi-generator/issues/1593), [Jackson issue (stale)](https://github.com/FasterXML/jackson-dataformats-text/issues/98))

It would be possible to support YAML anchors directly in smallrye-open-api by "preprocessing" a YAML file with snakeyaml.

I'm not sure how much appetite there is to have this kind of change. IMO if it's added to smallrye-open-api, then it should be hidden behind a toggle at least, or better using another `io.smallrye.openapi.runtime.io.Format` type.

The cleaner solution would probably be to add support for YAML anchors to Jackson.

MikeEdgar:
Fixes #1237 

TBD if this should be either opt-in (default disabled) or opt-out (default enabled). Currently, it is enabled and cannot be disabled.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -123,3 +123,11 @@ Specify a strategy to be used globally for generating operation IDs when not spe
 ** `CLASS_METHOD` (may result in duplicate operation IDs in an application where REST endpoint simple class name + method names are not unique)
 ** `PACKAGE_CLASS_METHOD`
 ** A fully-qualified class name of an implementation of `io.smallrye.openapi.api.OperationIdGenerator` (as of 4.1.2, https://javadoc.io/doc/io.smallrye/smallrye-open-api-core/latest/io/smallrye/openapi/api/OperationIdGenerator.html[see JavaDoc])
+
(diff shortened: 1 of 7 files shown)


phillip-kruger (approved):

phillip-kruger:
Maybe opt-out (default enabled) ?

MikeEdgar:
> Maybe opt-out (default enabled) ?

I agree that's the better option. This change is closer to a bug fix since using aliases properly is expected when they are present. In the unlikely case users have aliases already that are not being expanded, the fix can be disabled.

MikeEdgar:
@phillip-kruger please take a look at the second commit and if you're good with this I'll go ahead and merge. Thanks!

phillip-kruger (approved):

PR #2385, opened 2025-10-14
Adding a test to validate processing rules (@ExternalDocumentation ATM)

fabiobrz:
This test aims at validating processing rules, see https://download.eclipse.org/microprofile/microprofile-open-api-4.1/microprofile-openapi-spec-4.1.html#_processing_rules
It is about an application which includes a resources that is annotated with `@ExternalDocumentation`.

According to the MicroProfile OpenAPI Javadoc, the generated OpenAPI documentation should include an `externalDocs` element at root level, but it does not, and the test is failing.

_Note_: 
The test logic uses a custom `OASModelReader` instance in order to generate the OpenAPI model with an initial value for the `externalDocs` element, which should be overridden by the annotation. The test would fail even when such initial step is not executed, i.e. when line 39 is commented out.

### DIFF
--- a/testsuite/extra/src/test/java/io/smallrye/openapi/tck/extra/ProcessingRulesTest.java
+++ b/testsuite/extra/src/test/java/io/smallrye/openapi/tck/extra/ProcessingRulesTest.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


MikeEdgar (commented):
Since using `@ExternalDocumentation` on a class is [deprecated in MP OpenAPI 4.2](https://github.com/microprofile/microprofile-open-api/pull/735) and the test for method-level use was [added](https://github.com/microprofile/microprofile-open-api/pull/714), let's go ahead and close this one.

Issue #2401, opened 2025-11-05
Openapi generation somewhat slow


Postremus:
I noticed that openapi generation takes about half the total build time of the quarkus application - at least for this reproducer.
 Can this be improved?

Reproducer: 
The model of this reproducer is genarated based on the openapi definition of one of my (public) real world applications, where I noticed this initially. 

https://github.com/Postremus/smallrye-openapi-slow-repr

1. mvn clean install - just to have all dependencies ready
2. mvn quarkus:build. Remember the time noted in the log line 
```[INFO] [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 13063```
4. remove quarkus-smallrye-openapi from the pom.xml
5. mvn quarkus:build. Take note of the time noted in the log 
```[INFO] [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 6714```
6. Compare both times. Without openapi build will be much faster.

Flamegraphs:

[wall-1-reverse.html](https://github.com/user-attachments/files/23372805/wall-1-reverse.html)
[wall-1.html](https://github.com/user-attachments/files/23372804/wall-1.html)
[cpu-1-reverse.html](https://github.com/user-attachments/files/23372806/cpu-1-reverse.html)
[cpu-1.html](https://github.com/user-attachments/files/23372802/cpu-1.html)
[alloc-1.html](https://github.com/user-attachments/files/23372803/alloc-1.html)

No reverse graphs for the allocation profiling - process hit the memory limit..

I can provide jfr recordings over zulip / mail if desired (github does not like a 200mb file).

**Comment by MikeEdgar** [comment]
Thanks for reporting this, I'll take a look.

**Comment by MikeEdgar** [comment]
@Postremus if you can give it a try, PR #2404 has some changes that seem to improve this.

**Comment by Postremus** [comment]
@MikeEdgar 

Tests done on a different vm compared to yesterday. Each is the fastest of 3 runs of mvn quarkus:build.

Without smallrye-openapi:
```
[INFO] [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 5002ms
```

With Smallrye-openapi (current version from quarkus):
```
 [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 10337ms
```

With PR 2404:
```
[io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 6896ms
```

Already way better! Thanks.

[...]

PR #2404, opened 2025-11-06
Performance improvements, replace several streams w/loops

MikeEdgar:
For #2401

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/OpenApiDataObjectScanner.java
@@ -34,6 +34,7 @@ import io.smallrye.openapi.runtime.io.schema.SchemaFactory;
 import io.smallrye.openapi.runtime.scanner.dataobject.AnnotationTargetProcessor;
 import io.smallrye.openapi.runtime.scanner.dataobject.AugmentedIndexView;
 import io.smallrye.openapi.runtime.scanner.dataobject.DataObjectDeque;
+import io.smallrye.openapi.runtime.scanner.dataobject.DataObjectDeque.PathEntry;
 import io.smallrye.openapi.runtime.scanner.dataobject.IgnoreResolver;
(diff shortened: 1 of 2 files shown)


phillip-kruger (approved):

gsmet:
@MikeEdgar so only changing streams to for loops makes such a huge difference?

MikeEdgar:
> @MikeEdgar so only changing streams to for loops makes such a huge difference?

@gsmet also the change in `AugmentedIndexView` is quite significant. Instead of asking the Jandex class for all methods and filtering the stream, it now asks for the method by name + parameter types. Most of the other changes to switch from streams to loops help with reducing the call depth and also make debugging/stepping through the scanner easier.

I think there is very likely to be some more room to improve how we use Jandex and possibly propose some features there to help out further.

gsmet:
@mikeEdgar FYI I will push the new Jandex to next 3.29.z and 3.30.0.

MikeEdgar:
I think this should be ready to go now. It can be seen in the flame grapns how significant the reduction in parameter processing is with these changes. Thank you @Postremus for bringing up the issue.

Release 4.2.1: [flamegraph-cpu-sropenapi-4.2.1.html](https://github.com/user-attachments/files/23506178/flamegraph-cpu-sropenapi-4.2.1.html) (filtered to io.smallrye.openapi)
image

Changes in this PR: [flamegraph-cpu-sropenapi-pr2404.html](https://github.com/user-attachments/files/23506179/flamegraph-cpu-sropenapi-pr2404.html) (filtered to io.smallrye.openapi)
image

phillip-kruger (approved):

Postremus:
@gsmet 
Is it possible to include this with Quarkus 3.30 / 3.30.1? Seems to be a bit late, but.. hope never dies :)

gsmet:
@MikeEdgar do you plan to release today? If so, we could target Quarkus 3.30.

MikeEdgar:
> @MikeEdgar do you plan to release today? If so, we could target Quarkus 3.30.

Yes, I'll release it soon.

PR #2422, opened 2025-11-21
Set `multipleOf` for numbers with `@Digits`, stop generating `pattern`

### LINKED ISSUE #2384: Automatically determine multipleOf from `Digits` constraint
Right now to document and validate a bigdecimal field with a specific count of fractions I have to write something like this:

```
@Schema(multipleOf = 0.001)
@Digits(integer = 5, fraction = 3)
@Min(0)
@DecimalMax("99999.999")
public BigDecimal width;
```

Could maybe the `@Schema(multipleOf = 0.001)` automaticaly be determined if not specified otherwise?
E.g. `@Digits(integer = 5, fraction = 3)` -> `@Schema(multipleOf = 0.001)`

This would, in my opinion, at least cover the most common case where multipleOf is simply used to explicitily state in the doc the fraction count.

MikeEdgar:
Remove incorrect mapping of number `@Digits` to `pattern` schema constraint and replace with `multipleOf` when not already set in the schema. Uses `multipleOf: 1` for integers (i.e. `Digits#fraction()` <= 0) and `multipleOf` based on `fraction` value when > 0.

Fixes #2384

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/DataObjectLogging.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/DataObjectLogging.java
@@ -15,7 +15,8 @@ import org.jboss.logging.annotations.MessageLogger;
 
 @MessageLogger(projectCode = "SROAP", length = 5)
 interface DataObjectLogging extends BasicLogger {
-    DataObjectLogging logger = Logger.getMessageLogger(lookup(), DataObjectLogging.class, DataObjectLogging.class.getPackage().getName());
+    DataObjectLogging logger = Logger.getMessageLogger(lookup(), DataObjectLogging.class,
+            DataObjectLogging.class.getPackage().getName());
 
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 6000, value = "Processing @Schema annotation %s on a field %s")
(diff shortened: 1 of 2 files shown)


MikeEdgar:
@Azquelt please take a look if you have a chance and let me know if anything here appears to you that it may conflict with the spirit of https://github.com/microprofile/microprofile-open-api/issues/717.

MikeEdgar:
@Postremus FYI. Please take a look and give feedback if this looks amiss.

Postremus:
@MikeEdgar looks good to me.

MikeEdgar:
> We could infer minimum and maximum values based on `Digits.integer` if it's not otherwise constrained.

I considered this one too, but didn't do it here mainly for the reason you mentioned about being inaccurate. I suppose if `@Digits` is not scanned until after all of the possible min/max constraints, it should be fine.

> We could infer `"multipleOf": 1` for integer types

You mean when no `@Digits` is present at all? I think for the integer Java types we should be generating `type: [ integer ]` so it may be redundant.

Azquelt:
> I think for the integer Java types we should be generating `type: [ integer ]` so it may be redundant.

Ah that's true, it would then be redundant.

MikeEdgar:
I've opened #2423 for the additional handling to set min/max constraints so we can think about that in the future and not block this fix/change for now.

PR #2466, opened 2026-01-28
MicroProfile OpenAPI 4.2 updates

MikeEdgar:
This PR includes the changes for updating from MicroProfile OpenAPI 4.1.x to 4.2

Related upstream PRs
- https://github.com/microprofile/microprofile-open-api/pull/714
- https://github.com/microprofile/microprofile-open-api/pull/728
- https://github.com/microprofile/microprofile-open-api/pull/733
- https://github.com/microprofile/microprofile-open-api/pull/736

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
@@ -365,29 +365,29 @@ public class BeanValidationScanner {
             // Both attributes are required - safe to use primitives.
             final int integerPart = context.annotations().value(constraint, "integer");
             final int fractionPart = context.annotations().value(constraint, "fraction");
+
+            if (integerPart < 1 && fractionPart < 1) {
+                // Bogus constraint does not allow any digits, bail out
+                return;
+            }
+
             final StringBuilder pattern = new StringBuilder(50);
 
             pattern.append('^');
+            pattern.append("[+-]?"); // optional sign
+            pattern.append("(?=.*\\d)"); // look-ahead to ensure some digits present
 
             if (integerPart > 0) {
-                pattern.append("\\d");
-
-                if (integerPart > 1) {
-                    pattern.append("{1,").append(integerPart).append('}');
-                }
+                pattern.append("\\d{0,").append(integerPart).append('}');
             }
+
+            pattern.append("(?:\\.");
 
             if (fractionPart > 0) {
-                pattern.append("([.]\\d");
-
-                if (fractionPart > 1) {
-                    pattern.append("{1,").append(fractionPart).append("}");
-                }
-
-                pattern.append(")?");
+                pattern.append("\\d{0,").append(fractionPart).append('}');
             }
 
-            pattern.append('$');
+            pattern.append(")?$");
             schema.setPattern(pattern.toString());
         }
     }
@@ -398,11 +398,12 @@ public class BeanValidationScanner {
         if (constraint != null && schema.getMultipleOf() == null) {
             // `fraction` attribute is required - safe to use primitive.
             final int fractionPart = context.annotations().value(constraint, "fraction");
-            BigDecimal multipleOf;
+            BigDecimal multipleOf = null;
 
             if (fractionPart > 0) {
                 multipleOf = new BigDecimal(BigInteger.ONE, fractionPart);
-            } else {
+            } else if (SchemaSupport.getNonNullType(schema) != Schema.SchemaType.INTEGER) {
+                // multipleOf=1 is redundant for integers, apply only for generic numbers
                 multipleOf = BigDecimal.ONE;
             }
 
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
@@ -375,10 +375,12 @@ public class BeanValidationScanner {
 
             pattern.append('^');
             pattern.append("[+-]?"); // optional sign
-            pattern.append("(?=.*\\d)"); // look-ahead to ensure some digits present
+            pattern.append("(?=.*?\\d)"); // look-ahead to ensure some digits present
 
             if (integerPart > 0) {
                 pattern.append("\\d{0,").append(integerPart).append('}');
+            } else {
+                pattern.append("0?");
             }
 
             pattern.append("(?:\\.");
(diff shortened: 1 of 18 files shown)


MikeEdgar (commented):
@Azquelt these are the SmallRye changes for MP OpenAPI 4.2. Once this is in, I'll do an alpha release of SmallRye OpenAPI 4.4. There will probably be a couple additional features in the final release, including Jackson 3 support (#2615) and unused tag pruning (#2590).

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
```diff
@@ -365,29 +365,29 @@ void digitsString(AnnotationTarget target, Schema schema) {
...
+            pattern.append("(?:\\.");
 
-                pattern.append(")?");
+            if (fractionPart > 0) {
+                pattern.append("\\d{0,").append(fractionPart).append('}');
             }
 
-            pattern.append('$');
+            pattern.append(")?$");
             schema.setPattern(pattern.toString());
```
Azquelt, line 393: This whole pattern is interesting. There's nothing in Jakarta Validation I can see which says what's accepted as a number for `@Digits` when applied to a `CharSequence`.

Is any whitespace accepted? Is exponent notation accepted? Is a lone decimal point accepted? Are leading zeroes counted? Who knows!

Allowing numbers like `-.3` and `5.` seems reasonable to me.

I note that with `integer = 0`, you would not be allowed `0.2`, but would be allowed `.2` which seems weird.

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
```diff
@@ -365,29 +365,29 @@ void digitsString(AnnotationTarget target, Schema schema) {
...
+            if (integerPart < 1 && fractionPart < 1) {
+                // Bogus constraint does not allow any digits, bail out
+                return;
+            }
+
             final StringBuilder pattern = new StringBuilder(50);
 
             pattern.append('^');
+            pattern.append("[+-]?"); // optional sign
+            pattern.append("(?=.*\\d)"); // look-ahead to ensure some digits present
```
Azquelt:
```suggestion
            pattern.append("(?=.*?\\d)"); // look-ahead to ensure some digits present
```
To be non-greedy. Unsure if that matters in a lookahead like this.

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
```diff
@@ -365,29 +365,29 @@ void digitsString(AnnotationTarget target, Schema schema) {
...
 
             if (integerPart > 0) {
-                pattern.append("\\d");
-
-                if (integerPart > 1) {
-                    pattern.append("{1,").append(integerPart).append('}');
-                }
+                pattern.append("\\d{0,").append(integerPart).append('}');
             }
```
Azquelt, line 384: Maybe allow optional leading zero when `integerPart <= 0`?

model/src/main/java/io/smallrye/openapi/model/BaseExtensibleModel.java
```diff
@@ -70,102 +86,48 @@ public void setAllProperties(Map<String, ?> allProperties) {
...
 
     @Override
     protected <V> void addListPropertyEntry(String name, V value) {
-        if (extensionNames.contains(name)) {
-            // Property replaces the extension entry.
-            extensionNames.remove(name);
-        }
+        assertNotExtension(name, "Property %s has already been set as an extension and cannot be modified as a List");
         super.addListPropertyEntry(name, value);
     }
```
Azquelt, line 107: If you call `setListProperty("x-foo", List.of("bar"))`, it will add `x-foo` as an extension

If you call `addListPropertyEntry("x-foo", "baz")`, and `x-foo` is already an extension, it will reject it.

If you call `addListPropertyEntry("x-foo", "baz")`, and `x-foo` does not exist, it will add `x-foo` as a _non-extension_, which seems weird and unexpected.

Possibly the change needed here is that `assertNotExtension` should call `isExtension` rather than checking `extensionNames`.

model/src/main/java/io/smallrye/openapi/model/BaseModel.java
```diff
@@ -48,7 +48,7 @@ public int getModCount() {
         return modCount;
     }
 
-    Map<String, Object> getProperties() {
+    Map<String, Object> getModelProperties() {
         return properties;
     }
```
Azquelt, line 53: Was this just renamed to make it more obviously different from `getAllProperties()` and `getProperties(Class)`?

If so that makes sense, if not then was there another reason for renaming it?

Azquelt (commented):
Sorry this took a bit longer than I'd hoped.

model/src/main/java/io/smallrye/openapi/model/BaseModel.java
MikeEdgar: > Was this just renamed to make it more obviously different from getAllProperties() and getProperties(Class)?

Yes, exactly. It became even worse because `Schema` has `getProperties` which is public, the implementation of which doesn't actually override this package-private method.

core/src/main/java/io/smallrye/openapi/runtime/scanner/dataobject/BeanValidationScanner.java
MikeEdgar: > This whole pattern is interesting. There's nothing in Jakarta Validation I can see which says what's accepted as a number for @Digits when applied to a CharSequence.
>
> Is any whitespace accepted? Is exponent notation accepted? Is a lone decimal point accepted? Are leading zeroes counted? Who knows!

Yeah, it's pretty vague. I check Hibernate validator's `DigitsValidatorForCharSequence` and it basically just creates a `BigDecimal` from the string and performs the validation with this check:

```java
int integerPartLength = bigDec.precision() - bigDec.scale();
int fractionPartLength = bigDec.scale() < 0 ? 0 : bigDec.scale();
boolean valid = (annotation.integer() >= integerPartLength && annotation.fraction() >= fractionPartLength)
```

> I note that with integer = 0, you would not be allowed 0.2, but would be allowed .2 which seems weird.

I'll fix this.

model/src/main/java/io/smallrye/openapi/model/BaseExtensibleModel.java
MikeEdgar: I'll check into this. I'd like to make these sealed once we move to Java 17, they shouldn't be callable to anyone using the interfaces.

MikeEdgar: I tightened up `assertNotExtension` to also check `isExtension` as you suggested. There shouldn't be a way for these methods to get called unless a user builds their own implementation, but it's good to have the extra check.

Azquelt (approved):
Thanks

phillip-kruger (approved):
Thanks @MikeEdgar !

PR #2548, opened 2026-04-24
Process profiles in filter phase

MikeEdgar:
For quarkusio/quarkus#53479

This change moves the handling of operation profiles to the filter phase, allowing operations contributed to the model from a static file or `OASModelReader` to be included/excluded according to any given configurations.

### DIFF
--- a/model/src/main/java/io/smallrye/openapi/model/Extensions.java
+++ b/model/src/main/java/io/smallrye/openapi/model/Extensions.java
@@ -100,15 +100,24 @@ public final class Extensions {
     public static boolean includedProfile(Extensible<?> extensible, Set<String> included, Set<String> excluded) {
         Set<String> profiles = getProfiles(extensible);
 
-        if (!excluded.isEmpty()) {
-            return excluded.stream().noneMatch(profiles::contains);
+        if (profiles.isEmpty()) {
+            // Include an extensible without any profiles attached when none
+            // are explicitly included via configuration.
+            return included.isEmpty();
         }
 
-        if (included.isEmpty()) {
+        if (included.stream().anyMatch(profiles::contains)) {
             return true;
         }
 
-        return included.stream().anyMatch(profiles::contains);
+        if (excluded.stream().anyMatch(profiles::contains)) {
+            // Exclude the extensible if it has configured exclusions and also
+            // does not have any configured inclusions. This also means that
+            // inclusion overrides exclusion if a profile is configured for both.
+            return false;
+        }
+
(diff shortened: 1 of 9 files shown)


phillip-kruger (dismissed):
LGTM

MikeEdgar:
@Azquelt I don't recall if you were curious to look at this one. I'll be looking to do a release in the next day or two, so please let me know if you'd like some additional time to go over it.

model/src/main/java/io/smallrye/openapi/model/Extensions.java
```diff
@@ -95,6 +97,20 @@ public static Set<String> getProfiles(Extensible<?> extensible) {
...
+            return excluded.stream().noneMatch(profiles::contains);
+        }
+
+        if (included.isEmpty()) {
+            return true;
+        }
+
+        return included.stream().anyMatch(profiles::contains);
+    }
+
```
Azquelt, line 122: This looks a bit sketchy. Can the user specify both and include list and an exclude list?

It's odd that the include list is ignored completely if the exclude list is specified.

I realise this logic has not changed.

MikeEdgar: Good point, I agree on that. Maybe something like this would be better (plus updating the test). Both include and exclude could be specified.

```java
public static boolean includedProfile(Extensible<?> extensible, Set<String> included, Set<String> excluded) {
    Set<String> profiles = getProfiles(extensible);

    if (profiles.isEmpty()) {
        // Include an extensible without any profiles attached when none
        // are explicitly included via configuration.
        return included.isEmpty();
    }

    if (included.stream().anyMatch(profiles::contains)) {
        return true;
    }

    if (excluded.stream().anyMatch(profiles::contains)) {
        // Exclude the extensible if it has configured exclusions and also
        // does not have any configured inclusions. This also means that
        // inclusion overrides exclusion if a profile is configured for both.
        return false;
    }

    return included.isEmpty();
}
```

MikeEdgar: I applied the change from my previous comment and updated the test case.

Azquelt (approved):

model/src/main/java/io/smallrye/openapi/model/Extensions.java
MikeEdgar: I had to revert this logic change due to some existing test cases and documentation around it already in Quarkus. We can revisit in the future if it comes up perhaps.

PR #2586, opened 2026-06-10
Update to Jandex 3.6.0 and implement new method

gsmet:
I plan to use this new method in Quarkus so it would be great if we could support it.

### DIFF
--- a/core/src/test/java/io/smallrye/openapi/runtime/scanner/FilteredIndexViewTest.java
+++ b/core/src/test/java/io/smallrye/openapi/runtime/scanner/FilteredIndexViewTest.java
@@ -272,4 +272,31 @@ class FilteredIndexViewTest {
                 .getAnnotationsWithRepeatable(DotName.createSimple(Parameter.class.getName()), index);
         assertEquals(4, params.size());
     }
+
+    @Test
+    void testGetAnnotationsWithRepeatableByContainerName() {
+        class Target {
+            @APIResponses({
+                    @APIResponse(),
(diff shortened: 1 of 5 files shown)


core/src/test/java/io/smallrye/openapi/runtime/scanner/FilteredIndexViewTest.java
```diff
@@ -272,4 +272,32 @@ void test2(@Parameter() String value1, @Parameter() String value2) {
...
+                APIResponses.class,
+                Parameter.class,
+                Target.class));
+
+        OpenApiConfig config = IndexScannerTestBase.emptyConfig();
+        FilteredIndexView view = new FilteredIndexView(index, config);
+
+        Collection<AnnotationInstance> responses = view
+                .getAnnotationsWithRepeatable(DotName.createSimple(APIResponse.class.getName()),
+                        DotName.createSimple(APIResponses.class.getName()));
```
Ladicek: BTW, the `IndexView` interface declares overloads that accepts `String`s and even `Class`es, so you can just write `view.getAnnotationsWithRepeatable(APIResponse.class, APIResponses.class)`.

gsmet: Thanks, I simplified it.

MikeEdgar (approved):
Thanks @gsmet

MikeEdgar:
Let me deal with the open PRs and I'll do a release.

PR #2587, opened 2026-06-10
Prevent conflation of unnamed @Parameter annotations with matching style

### LINKED ISSUE #2585: @BeanParam fields with identical @Parameter style are merged
When a `@BeanParam` class has two fields with `@Parameter` annotations sharing the same style (e.g. `ParameterStyle.FORM`) and neither explicitly sets `name`, SmallRye conflates the two parameters during spec generation. The second parameter's description overwrites the first, and the first parameter loses its schema metadata.

I would expect each `@QueryParam` field generating its own independent parameter entry in the OpenAPI spec with its own description and schema.

Example:
```
# Expected:
- description: "Field(s) to sort by..."    # sortBy's description
  name: sortBy
  schema: { type: array, items: { $ref: "#/components/schemas/MyEnum" } }
- description: "Set of fields to include..."    # fields' description
  name: fields
  schema: { type: array, items: { $ref: "#/components/schemas/MyEnum" } }

# Actual:
- name: fields                              # lost description and array schema
  schema: { type: string }
- description: "Set of fields to include..."  # fields' description on sortBy!
  name: sortBy
  schema: { type: array, items: { $ref: "#/components/schemas/MyEnum" } }
```

This may be working as designed - SmallRye's own tests always set `name` explicitly on `@Parameter` when used on `@BeanParam` fields. However, the failure mode is silent, instead of warning about ambiguous parameters, it produces a spec with swapped descriptions and lost schema.

Pepo48:
Closes: #2585

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
@@ -111,6 +111,33 @@ public abstract class AbstractParameterProcessor {
     private Set<String> processedMatrixSegments = new HashSet<>();
     private List<Parameter> preferredOrder;
 
+    /**
+     * Comparator to order framework parameter annotations before MP OpenAPI Parameter annotations.
+     *
+     * Processing the framework annotations first is preferred because they generally allow for
+     * parameter `name`, `location`, and `style` to be derived whereas those values may be null
+     * in the MP OpenAPI Parameter annotation.
+     */
+    private final class FrameworkParamsFirstComparator implements Comparator<AnnotationInstance> {
+        @Override
+        public int compare(AnnotationInstance o1, AnnotationInstance o2) {
+            boolean frameworkParam1 = !Names.PARAMETER.equals(o1.name()) && isParameter(o1.name());
+            boolean frameworkParam2 = !Names.PARAMETER.equals(o2.name()) && isParameter(o2.name());
+
+            if (frameworkParam1) {
+                if (!frameworkParam2) {
+                    return -1;
+                }
+            } else if (frameworkParam2) {
+                return 1;
+            }
+
+            return 0;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
```diff
@@ -1692,6 +1692,10 @@ public boolean equals(Object obj) {
                     return ref.equals(other.ref);
                 }
 
+                if (this.name == null && other.name == null) {
+                    return this == other;
+                }
+
```
MikeEdgar: I'm not entirely sure this won't introduce some unexpected issues somewhere. I've been looking at this locally and I think if we process the framework-specific annotations (e.g. JAX-RS) for any given target before the MP OpenAPI `@Parameter` annotation, the parameter context key will have obtained the name from the framework annotation already and it should work.

I have a working version that I can push to your PR unless you disagree. The test changes in this PR look good.

Pepo48: Sure, no worries. Go ahead either with pushing to this one or by surpassing with the new one. Thanks, Michael. 😊

MikeEdgar: I pushed the changes in the last commit. We can go ahead with merging this one unless you spot any issues with the change.

cc: @phillip-kruger @Azquelt if either of you would like to look at this one.

Pepo48: Thanks, Michael, LGTM.

MikeEdgar (approved):
This is ready to merge from my perspective, but I'll leave it open for a few days in case any others would like to review it or chime in on the change.

PR #2588, opened 2026-06-11
feat: support spring controllers that don't use the @RestController annotation

skmichaeldavis:
I was integrating smallrye into my project and found that virtually no endpoints were showing up.  Discovered it's because our codebase doesn't use the `@RestController` helper utility.  This change will allow detecting any beans that use `@RequestMapping`, which I believe is the true common denominator for spring-web-mvc.

In theory I think the RestController detection is redundant with this, but didn't want to fully swap it out just in case.  I changed to using a LinkedHashSet to avoid double counting beans with both annotations.

### DIFF
--- a/extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
+++ b/extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
@@ -1,13 +1,6 @@
 package io.smallrye.openapi.spring;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -152,13 +145,25 @@ public class SpringAnnotationScanner extends AbstractAnnotationScanner {
         // Get all Spring controllers and convert them to OpenAPI models (and merge them into a single one)
         Collection<AnnotationInstance> controllerAnnotations = context.getIndex()
                 .getAnnotations(SpringConstants.REST_CONTROLLER);
-        List<ClassInfo> applications = new ArrayList<>();
+        Collection<AnnotationInstance> requestMappingAnnotations = context.getIndex()
+                .getAnnotations(SpringConstants.REQUEST_MAPPING);
+
+        Set<ClassInfo> applications = new LinkedHashSet<>();
         for (AnnotationInstance annotationInstance : controllerAnnotations) {
             if (annotationInstance.target().kind().equals(AnnotationTarget.Kind.CLASS)) {
                 ClassInfo classInfo = annotationInstance.target().asClass();
                 applications.add(classInfo);
             } else {
                 SpringLogging.log.ignoringAnnotation(SpringConstants.REST_CONTROLLER.withoutPackagePrefix());
+            }
+        }
+
+        for (AnnotationInstance annotationInstance : requestMappingAnnotations) {
+            if (annotationInstance.target().kind().equals(AnnotationTarget.Kind.CLASS)) {
+                ClassInfo classInfo = annotationInstance.target().asClass();
+                applications.add(classInfo);
+            } else {
+                SpringLogging.log.ignoringAnnotation(SpringConstants.REQUEST_MAPPING.withoutPackagePrefix());
             }
         }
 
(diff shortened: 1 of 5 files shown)


extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
```diff
@@ -1,13 +1,6 @@
...
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
MikeEdgar: Can you unroll the wildcard imports? I think the CI will complain about them too.

extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
```diff
@@ -162,6 +158,15 @@ private void processControllerClasses(OpenAPI openApi) {
             }
         }
 
+        for (AnnotationInstance annotationInstance : requestMappingAnnotations) {
+            if (annotationInstance.target().kind().equals(AnnotationTarget.Kind.CLASS)) {
+                ClassInfo classInfo = annotationInstance.target().asClass();
+                applications.add(classInfo);
+            } else {
+                SpringLogging.log.ignoringAnnotation(SpringConstants.REQUEST_MAPPING.withoutPackagePrefix());
```
MikeEdgar: This will log at a `WARN` level. I suspect most of these annotations will be on methods, so it might be worth a new log method in `SpringLogging` at `DEBUG` level to reduce noise/confusion on them.

MikeEdgar (changes requested):
Thanks for the PR. I left a few comments. I also wonder whether simply having `@Controller` on the class is enough to trigger a scan of the methods. I see several Spring examples where there is no `@RequestMapping` at the class level. That can probably be left for a future PR if someone uses it that way.

extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
skmichaeldavis: Fixed, my IDE trying to be helpful

extension-spring/src/main/java/io/smallrye/openapi/spring/SpringAnnotationScanner.java
skmichaeldavis: Fixed.  Think I did it right, haven't used jboss-logging before

skmichaeldavis:
> Thanks for the PR. I left a few comments. I also wonder whether simply having `@Controller` on the class is enough to trigger a scan of the methods. I see several Spring examples where there is no `@RequestMapping` at the class level. That can probably be left for a future PR if someone uses it that way.

Controller might be a bit much since it doesn't have any special programmatic meaning. You could use `@Controller` for pretty much anything, it'd just be semantically confusing. My codebase doesn't do that, but I could see some large codebases having bad performance impacts from that.

extension-spring/src/test/java/io/smallrye/openapi/runtime/scanner/SpringAnnotationScannerTest.java
```diff
@@ -8,17 +8,7 @@
...
-import test.io.smallrye.openapi.runtime.scanner.resources.GreetingGetController;
-import test.io.smallrye.openapi.runtime.scanner.resources.GreetingGetControllerAlt;
-import test.io.smallrye.openapi.runtime.scanner.resources.GreetingGetControllerAlt2;
-import test.io.smallrye.openapi.runtime.scanner.resources.GreetingPostController;
-import test.io.smallrye.openapi.runtime.scanner.resources.GreetingPostControllerAlt;
-import test.io.smallrye.openapi.runtime.scanner.resources.GreetingPostControllerMultiplePaths;
-import test.io.smallrye.openapi.runtime.scanner.resources.GreetingPutController;
-import test.io.smallrye.openapi.runtime.scanner.resources.GreetingPutControllerAlt;
-import test.io.smallrye.openapi.runtime.scanner.resources.GreetingPutControllerMultiplePaths;
+import test.io.smallrye.openapi.runtime.scanner.resources.*;
```
MikeEdgar: One more wildcard import :)

MikeEdgar (changes requested):
One more import to fix, otherwise LGTM. Thanks!

MikeEdgar (approved):
Thanks for the PR @skmichaeldavis !

PR #2591, opened 2026-06-12
feat: don't error out with duplicate key when there's no actual conflict

skmichaeldavis:
A small tweak to be more forgiving.  Ran into this when using meta-annotations to add `x-smallrye-profile-blah` `@Extension`. If you end up with multiple of them on the same endpoint, currently your build errors out.  Given there's no actual problem with that, and its convenient to allow overlaps when you have multiple annotations at play to control documentation builds, I made this tweak

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/ModelIO.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/ModelIO.java
@@ -7,6 +7,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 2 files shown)


MikeEdgar:
Thanks for the PR @skmichaeldavis. What is the use case for duplicates? Do I understand correctly that you want to have this?

```java
@Extension(name = "x-smallrye-profile-blah" value = "")
@Extension(name = "x-smallrye-profile-blah" value = "")
```

skmichaeldavis:
@MikeEdgar Indirectly, but yes. I have multiple builds of documentation for different audiences and I pre-existing helper annotations (think like `@PublicAccess`) for controlling access and who can see the docs.  I'm working on moving from our old system to smallrye so  I have added the `@Extension(blahblah)` to those annotations. That works great, but sometimes there can be overlap and multiple annotations give the same profile Extension.

MikeEdgar (approved):
Ok, thanks. The change seems generally harmless so I'll go ahead and merge this.

PR #2596, opened 2026-06-16
Augment parameter schema with type's schema when no type given

### LINKED ISSUE #2595: @Schema(defaultValue) inside @Parameter on @BeanParam field loses type and enum
When a `@BeanParam` field has `@Parameter(schema = @Schema(defaultValue = "ASC"))` on an enum-typed field, SmallRye produces a schema with only the default value, type and enum are lost.

I would expect SmallRye to pick up the type and enum values from the Java field type and combine them with the explicitly provided default value.

Example:

```
public static class QueryOptions {
    @QueryParam("sortOrder")
    @Parameter(description = "Sort direction", schema = @Schema(defaultValue = "ASC"))
    SortOrder sortOrder;  // enum with ASC, DESC
}

# Expected:
schema:
  type: string
  enum: [ASC, DESC]
  default: ASC

# Actual:
schema:
  default: ASC
```

Pepo48:
… is set

* merge type, enum and format from the Java field type into the existing partial schema when it has no type set

Closes: #2595


MikeEdgar (commented):
Thanks for the PR @Pepo48 . I plan to review later today/tomorrow. I'm wondering if this issue affects schema handling generally and not just for parameters. E.g. does it happen for a property within another schema as well?

Pepo48:
@MikeEdgar I also wondered about this earlier and at least based on what I see from my research is that it doesn't affect schema properties. The POJO property pipeline handles this correctly via `TypeProcessor.processType()`, which resolves enums before `@Schema` is read.

The bug is in `mapParameterSchema()` in `AbstractParameterProcessor`, which handles all parameter types, not just `@BeanParam` fields. Verified on `@QueryParam` and `@HeaderParam`, both as `@BeanParam` fields and direct method arguments. In all cases, `parameterHasSchema()` returns early when any schema object exists, skipping type resolution entirely.

But, please, recheck this thoroughly, just to be on the safe side. 😊

MikeEdgar:
@Pepo48 pardon the delay, I am still thinking through this one. Your change works well for the narrow set of attributes in the new logic (type, enumeration, format), but really the issue impacts any type that is currently not being introspected. Granted, this can easily be worked around by adding `implementation = SomeClass.class` to the `@Schema` annotation nested within the `@Parameter`, but I think it might be best to handle it generally for more than just enumerations.

Pepo48:
Hey @MikeEdgar, I found some time to look and it again. I verified one by one with all 7 types (enum, Integer, Double, Long, Boolean, LocalDate, UUID) that they all lose their type info when `@Schema(defaultValue)` is present.

I tried to replace the manual property merge with a simpler approach. For ref types (UUID, LocalDate, enum), the `$ref` is now set on the existing schema, so enums are no longer a special case, they just behave like any other ref type. For non-ref types (Integer, Boolean, etc.), `mergeObjects` fills in `type`/`format` while preserving `defaultValue`. 

Is this roughly roughly the approach you had in mind? (I pushed the changes as a separate commit for now)

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
```diff
@@ -688,7 +688,22 @@ void mapParameterStyle(Parameter param, ParameterContext context) {
...
+            if (existingSchema != null && existingSchema.getRef() == null && existingSchema.getType() == null) {
+                Schema typeSchema = SchemaFactory.typeToSchema(scannerContext, context.targetType, null);
+                if (typeSchema != null) {
+                    if (typeSchema.getRef() != null) {
+                        existingSchema.setRef(typeSchema.getRef());
+                    } else {
+                        mergeObjects(existingSchema, typeSchema);
+                    }
+                }
+            }
```
MikeEdgar: A couple issues I see.
- The param schema could have multiple schemas (various media types). In that case, `param.getSchema` will be `null`, even though `ModelUtil.parameterHasSchema` is true.
- `mergeObjects` will override the entries on the first argument with those from the second. So in this case, if the type/enum has a default, it will overlay the one in the `@Parameter.schema`.

I am playing around with adding a few additional test parameters to your test case to show it.

MikeEdgar:
@Pepo48 I pushed a commit with some more test scenarios in the same test case and handling for multi-variant parameter content (i.e., >1 schema per parameter).

MikeEdgar:
@Pepo48 are you OK with the current state of this after my last commit? I'd like to get this in to the next patch release.

Pepo48:
@MikeEdgar of course, the tests prove that our side of concerns is fully fixed and covered and at the same time, I'm totally fine with whatever you propose on top of that. No need to wait for my approval 😊 

Thanks for the help, I really appreciate it!

MikeEdgar (approved):

PR #2615, opened 2026-07-10
Add support for Jackson v3 and JSON provider selection via builder

MikeEdgar:
- Add JSON provider for Jackson v3
- Enable provider selection using the SmallRyeOpenAPI.Builder
- Auto-select JSON provider when not chosen with the builder. Order of preference is Jackson3, Jackson2, Jakarta JSON.
- Adjust CI to test Jackson 3 separately and collect coverage statistics

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/io/JsonIO.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/io/JsonIO.java
@@ -17,6 +17,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import io.smallrye.openapi.api.OpenApiConfig;
+import io.smallrye.openapi.api.SmallRyeOpenAPI;
 import io.smallrye.openapi.model.BaseModel;
 
 /**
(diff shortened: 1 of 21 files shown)


MikeEdgar:
@Azquelt please let me know if you spot anything that could impact the OpenLiberty integration.

core/src/main/java/io/smallrye/openapi/runtime/io/JsonIO.java
```diff
@@ -61,10 +62,32 @@ default void mapObject(BaseModel<?> object, OB nodeBuilder) {
         }
     }
 
-    public static <V, A extends V, O extends V, AB, OB> JsonIO<V, A, O, AB, OB> newInstance(OpenApiConfig config) {
-        @SuppressWarnings("unchecked")
-        JsonIO<V, A, O, AB, OB> jackson = (JsonIO<V, A, O, AB, OB>) new JacksonJsonIO(config);
-        return jackson;
+    @SuppressWarnings("unchecked")
+    public static <V, A extends V, O extends V, AB, OB> JsonIO<V, A, O, AB, OB> newInstance(
```
Azquelt, line 66: Was this method API?

I assume not. We were using it in one place in Open Liberty to parse an `Info` object from config, but we can adapt that code when we update.

Azquelt (approved):
This looks good and I don't expect any issues for Liberty consuming it.

core/src/main/java/io/smallrye/openapi/runtime/io/JsonIO.java
MikeEdgar: I wouldn't say it's intended to be a public API, but nonetheless it's accessible. We can keep the original method and use the AUTO or JACKSON2 provider. Would that work?

Azquelt: That would keep our existing code working. For me it's not an issue, I'm happy to update Open Liberty when we consume this.

phillip-kruger (approved):

Issue #2630, opened 2026-08-15
Openapi build time performance could be improved still


Postremus:
While #2401 reduced the time it takes to build an openapi document, we still could go further.

Reproducer:
The model of this reproducer is genarated based on the openapi definition of one of my (public) real world applications, where I noticed this initially.
I updated the reproducer - compared to #2401 - with an (somewhat) up to date version of our openapi definition.

https://github.com/Postremus/smallrye-openapi-slow-repr

1. mvn clean install - just to have all dependencies ready
2. mvn quarkus:build. Remember the time noted in the log line
[INFO] [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 5870ms
3. remove quarkus-smallrye-openapi from the pom.xml
4. mvn quarkus:build. Take note of the time noted in the log
[INFO] [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 3984ms

Compare both times. Without openapi the build will be much faster.

Following screenshot was taken from a profiling session. 
- The concrete numbers are not representative (profiler overhead is massive..).
- Taken from a quarkus hot reload, easier right now to profile for me. Any savings here should translate to quarkus:build as well.

<img width="2519" height="1112" alt="Image" src="https://github.com/user-attachments/assets/442da67d-2d40-48ea-91dd-39f5412e10a6" />

Some initial points I noticed:
* `io.smallrye.openapi.jaxrs.JaxRsParameterProcessor#isResourceMethod` might not need to build the ancestry list, if its `method` parameter has a http method already. I am testing a changeset on this right now.
* `io.smallrye.openapi.runtime.scanner.spi.AbstractParameterProcessor#readParametersInherited` is called repeatedly with the same `clazz` parameter. Resulting in 258 calls to readparameters, finding 583 parameters in the whole class, resulting in ~15k calls to isresourcemethod in the end. Maybe this could be improved.

Postremus:
First PR (for isResourcemethod) is provided.

Next I am looking into Annotations.java.

PR #2631, opened 2026-08-15
perf: Reduce ancestry calls in isResourceMethod

Postremus:
ancestry determines the parent methods the given method is overriden. Which in turn are checked if they are a resource method.

However, often the resource methods are not overriding any parent methods. so the ancestry call would be wasted.

ancestry() is slow (relativly speaking) because of the interfaces() calls.

Introduce a fast path for first checking if the given method is already resource method, before falling back to checking ancestry.

This saves about 100ms when doing a quarkus:build.

Related to #2630


MikeEdgar (dismissed):
Looks good. I'm fine to merge this as-is, but it might be nice to merge the `if` statements as suggested by Sonar to avoid accumulating more issues there. Let me know if you're planning any other updates to this PR.

Postremus:
@MikeEdgar done.

I don't plan any more changes to this PR.

MikeEdgar (approved):

PR #2632, opened 2026-08-17
perf: reduce declaredAnnotations calls in Annotations.java

Postremus:
Annotations.getDeclaredAnnotation returns the list of annotations, matching a given name. A list is returned because composable annotations are also search through a match. The composable annotations where searched through unconditionally, meaning that even if no composable annotation is annotated with the given name, they are still considered. target.declaredAnnotations() (without filter by name) was used, and manually searched through for the initial name match, and then later for the composable case. -> Meaning jandex created a new ArrayList for each getDeclaredAnnotation call.

We now build an initial lookup set of annotations which are part of a composed annotation, which is used to circumvent declaredAnnotations() calls, and the whole search for composable annotations. A direct call to target.declaredAnnotation(name) is now preferred, since no new ArrayList is created in the background.

This saves about 200ms when doing a quarkus:build.

related to #2630

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/util/Annotations.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/util/Annotations.java
@@ -57,13 +57,52 @@ public final class Annotations {
     private final AnnotationScannerContext context;
     private final Collection<DotName> excludedPackages;
 
+    /**
+     * Names of annotations which are part of a composable annotation; composed annotations need not be searched for when the
+     * initially searched for annotation name is not present in this set.
+     */
+    private final Set<DotName> composedSearchCandidates;
+
     public Annotations(AnnotationScannerContext context) {
         this.context = context;
         this.excludedPackages = Names.componentize(context.getConfig().getScanCompositionExcludePackages())
                 .values();
+
+        composedSearchCandidates = new HashSet<>();
+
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/smallrye/openapi/runtime/util/Annotations.java
```diff
@@ -57,13 +57,52 @@ public final class Annotations {
...
+
+        for (AnnotationInstance retention : context.getAugmentedIndex().getAnnotations(DotName.RETENTION_NAME)) {
+            ClassInfo annotatedType = retention.target().asClass();
+            AnnotationValue policy = retention.value();
+
+            if (!composable(annotatedType.name())
+                    || policy == null
+                    || !"RUNTIME".equals(policy.asEnum())) {
+                continue;
+            }
```
MikeEdgar: Maybe slightly cheaper to check `!composable` last.

MikeEdgar (dismissed):
This looks good, thanks @Postremus . I just have one comment on the short-circuit code.

core/src/main/java/io/smallrye/openapi/runtime/util/Annotations.java
Postremus: Makes sense, thanks for the review.

MikeEdgar (approved):

PR #2636, opened 2026-08-24
feat: add APPEND_NUMBER behavior for duplicate operationIds

adancus:
Implement a new `DuplicateOperationIdBehavior.APPEND_NUMBER` strategy that auto-resolves
duplicate operationIds by appending `_1`, `_2`, etc. instead of failing or warning. Enables
viable use of `METHOD/CLASS_METHOD` strategies even when overloaded methods produce identical
operation IDs.

- Add `APPEND_NUMBER` enum value to `DuplicateOperationIdBehavior`
- Modify `saveOperationId()` to return the resolved ID
- Loop through `_1`, `_2`, ... until `putIfAbsent()` succeeds
- Update `processOperation()` to apply the renumbered ID
- Add test with inline resource containing duplicate method names

Fixes duplicate operationId conflicts without requiring explicit `@Operation(operationId="...")`
annotations on every overload.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -113,7 +113,7 @@ Configures handling of schema inheritance. Allowed values include `NONE` (defaul
 ----
 mp.openapi.extensions.smallrye.duplicateOperationIdBehavior
 ----
-Set to `FAIL` to abort in case of duplicate operationIds, set to `WARN` to log warnings when the build encounters duplicate operationIds. Default value is `WARN`.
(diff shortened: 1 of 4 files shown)


MikeEdgar (approved):
This looks reasonable to me. @phillip-kruger any concerns from your perspective?

phillip-kruger:
@MikeEdgar lgtm

adancus:
thanks for the quick review! I have amended the commit to be signed

PR #2644, opened 2026-08-26
fix: preserve unresolved parameter references

### LINKED ISSUE #2643: Multiple unresolved parameter references are treated as duplicates
### Description

When an operation declares multiple `@Parameter` annotations using different `ref` values, only the first parameter is included in the generated OpenAPI document if the referenced components are added later by an `OASFilter`.

This was originally reported in quarkusio/quarkus#56241:
https://github.com/quarkusio/quarkus/issues/56241

The same behavior occurs when the parameters are declared inside an `@Parameters` container.

### Example

```java
@Parameter(ref = "#/components/parameters/XHeader2")
@Parameter(ref = "#/components/parameters/XHeader1")
```

Expected:

```yaml
parameters:
  - $ref: "#/components/parameters/XHeader2"
  - $ref: "#/components/parameters/XHeader1"
```

Actual:

```yaml
parameters:
  - $ref: "#/components/parameters/XHeader2"
```

### Analysis

When the referenced components are supplied by a filter, they are not available while annotations are being scanned. The parameters therefore still have null `name`, `in`, and `style` values.

`AbstractParameterProcessor#getParameters` performs its final duplicate check using only those fields:

```java
Arrays.asList(param.getName(), param.getIn(), param.getStyle())
```

As a result, every unresolved reference has the same `[null, null, null]` key, and all but the first are removed.

Referenced parameters should use their `$ref` as the duplicate key. Inline parameters should continue to use `name`, `in`, and `style`.

A regression test should cover both directly repeated `@Parameter` annotations and the `@Parameters` container, with the referenced components supplied after annotation scanning.

arnabnandy7:
## Description

Fixes #2643.

Multiple unresolved parameter references were treated as duplicates because their `name`, `in`, and `style` values were all `null` during annotation scanning.

Referenced parameters now use their `$ref` as the duplicate key. Inline parameters continue to use `name`, `in`, and `style`.

## Changes

- Preserve distinct unresolved parameter references.
- Retain existing duplicate handling for inline parameters.
- Add regression coverage for:
  - Repeated `@Parameter` annotations.
  - Explicit `@Parameters` containers.
  - Referenced components supplied after scanning by a filter.

## Testing

```bash
./mvnw -pl extension-jaxrs -am -Dtest=ParameterScanTests#testMultipleUnresolvedParameterRefs test
```

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
@@ -565,12 +565,18 @@ public abstract class AbstractParameterProcessor {
                      * post-processed later and renamed with a "Matrix" suffix, if needed.
                      *
                      * https://spec.openapis.org/oas/v3.1.0.html#parameter-object
-                     */
-                    ScannerSPILogging.log.duplicateParameter(
-                            param.getName(),
-                            String.valueOf(param.getIn()),
-                            String.valueOf(param.getStyle()),
-                            String.valueOf(context.target));
+                    */
+                    if (param.getRef() != null) {
+                        ScannerSPILogging.log.duplicateParameterReference(
+                                param.getRef(),
+                                String.valueOf(context.target));
+                    } else {
+                        ScannerSPILogging.log.duplicateParameter(
+                                param.getName(),
+                                String.valueOf(param.getIn()),
+                                String.valueOf(param.getStyle()),
+                                String.valueOf(context.target));
+                    }
                 }
             }
         }
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
@@ -549,12 +549,17 @@ public abstract class AbstractParameterProcessor {
             Parameter param = mapParameter(resourceMethod, context);
 
             if (param != null) {
-                if (paramKeys.add(Arrays.asList(param.getName(), param.getIn(), param.getStyle()))) {
+                List<?> paramKey = param.getRef() != null
+                        ? Collections.singletonList(param.getRef())
+                        : Arrays.asList(param.getName(), param.getIn(), param.getStyle());
+
+                if (paramKeys.add(paramKey)) {
                     parameters.add(param);
                 } else {
                     /*
-                     * Parameters are unique by name and location - filter out any
-                     * duplicates seen earlier in the stream. We also consider the parameter's
+                     * Referenced parameters are unique by reference. Inline parameters are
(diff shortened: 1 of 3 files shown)


extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/ParameterScanTests.java
```diff
@@ -549,6 +552,52 @@ void testJakartaParameterRefOnly() throws IOException, JSONException {
...
+            public void filterOpenAPI(OpenAPI openAPI) {
+                openAPI.setComponents(OASFactory.createComponents()
+                        .addParameter("XHeader1", OASFactory.createParameter()
+                                .name("XHeader1")
+                                .in(org.eclipse.microprofile.openapi.models.parameters.Parameter.In.HEADER))
+                        .addParameter("XHeader2", OASFactory.createParameter()
+                                .name("XHeader2")
+                                .in(org.eclipse.microprofile.openapi.models.parameters.Parameter.In.HEADER)));
+            }
+        }.filterOpenAPI(result);
```
MikeEdgar: This part doesn't seem necessary. The subsequent assertions only care about `paths`. It would be needed for a technically correct OpenAPI in a real world API, but it probably just complicates things for the purposes of the test.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
MikeEdgar: This needs some adjustment for when `param.getRef()` is non-null. I.e. an alternate log statement.

MikeEdgar (changes requested):

extension-jaxrs/src/test/java/io/smallrye/openapi/runtime/scanner/ParameterScanTests.java
arnabnandy7: @MikeEdgar I removed the filter and component setup since the test only needs to verify that both references survive annotation scanning. This keeps the regression test focused on the deduplication behavior.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
arnabnandy7: Added a dedicated debug message for duplicate parameter references, so referenced parameters now log their `$ref` instead of reporting null `name`, `in`, and `style` values. 

The existing message is still used for inline parameter duplicates.

MikeEdgar:
@arnabnandy7 the CI is failing due to formatting. Please take a look at the error message for details. Basically, just run the build locally and commit the formatting changes.

arnabnandy7:
> @arnabnandy7 the CI is failing due to formatting. Please take a look at the error message for details. Basically, just run the build locally and commit the formatting changes.

sure I missed running second time @MikeEdgar

MikeEdgar (approved):

PR #2645, opened 2026-08-29
feature: ensure stable order for sorted-parameters.enable=false

Postremus:
`sorted-parameters.enable=false` allows to preserve definition order of parameters into the final openapi document.
Jandex first scans for annotations (our source of parameters) on fields, then on methods.
The output was therefore never in definition order.
Compare `org.jboss.jandex.Indexer.indexWithSummary`, which calls processFieldInfo before processMethodInfo.
(matchin jvms, which already sorts fields before methods https://docs.oracle.com/javase/specs/jvms/se25/html/jvms-4.html)

For openapi this means, we can add a test which makes sure this order stays stable with `sorted-parameters.enable=false`.

One hiccup was though, that the keys of ClassInfo.annotationsmap are not sorted in any particular way. The order changed mostly when adding a new kind of parameter type, e.g. adding header params on a resource where previously only query and path params where present

I therefore added an explicit sort on the In parameter in the `sorted-parameters.enable=false` case. This means though that we might break the currently existing order for existing apis.

Related to #2630. The next changeset basically requires that we agree upon a stable ordering.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -140,7 +140,7 @@ Set this boolean value to disable the merging of the deprecated `@Schema` `examp
 ----
 mp.openapi.extensions.smallrye.sorted-parameters.enable
 ----
-Set this boolean value to enable or disable the sorting of parameter array entries during annotation scanning. When enabled (set to `true`), parameters will be order either by their order within a `@Parameters` annotation on an operation method or (in the absence of that annotation) by their `$ref`, `in`, and `name` attributes. When disabled (set to `false`), parameters will be in the order they are encountered in the Java code. If not set, it will default to `true`.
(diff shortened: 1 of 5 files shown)


Postremus:
Draft; I might want to expand the tests further.
Might be I interesting to explicitly add parameters sourced from method annotations to the test.

Postremus:
Ok I expanded the tests a bit more. Should be ready now.

Postremus:
Thinking some more. This approach might not be that great.

Before, at least the method parameters (of the resource method) where sorted in declaration order.
Only parameters from the class or from beanparams where grouped by in.

Now however everything is grouped. Within each group the parameters are in declaration order.

Not sure if that is a nicer behaviour.

I will try out if I can get this PR to emit every parameter in declaration order.


Ok I just checked. I will drop this PR. My next changeset ensures declaration order for free. I will pull the test changes into the next PR.

PR #2646, opened 2026-08-30
perf: optimize AbstractParameterProcessor.readParameters

Postremus:
readParameters checked each annotation on the class, if it is parameter. Then checked if the annotation instances target is FIELD|METHOD|METHOD_PARAMETER, and then checked that target if it has any parameter annotations. I turned it around. We overall only check fields and methods now for parameter annotations. For fields we only require one isParameter check for the specific annotation instance. For methods, we check if the method could even be a bean property method first. Short circuit for parameterless methods (constructors, static initializers). This saves around 2/3 of the calls towards isResourceMethod (150k -> 50k). We also save on retrieving the annotations of each target again and again, if the target has multiple parameter annotations (resource methods, constructors).

Optimize JaxRsParameterProcessor.hasMethodHTTPMethodAnnotation further to save on the ArrayList allocation inside MethodInfo.declaredAnnotations(). Also compare by name directly instead of AnnotationInstance.

sorted-parameters.enable=false had the problem, that parameters picked up from beanparams where not sorted in declaration order. They where instead grouped by annotation (i.e. path param, query param, etc), and in each group in declaration order. The behaviour is now so that beanparams also follow declaration order; And their parameters are inserted right between the other parameters from the "parents" (e.g. the resource methods) parameters. This is also ensured for parameters declared on constructors. Jandex sorts method parameters alphabetically, so we have to resort based on parameter position.

Add a test to ensure the declaration order sorting stays consistent.

Saves about 80ms of quarkus:build.

related to #2630
replaces #2645

### DIFF
--- a/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
+++ b/core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
@@ -43,7 +43,6 @@ import org.jboss.jandex.DotName;
 import org.jboss.jandex.FieldInfo;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
-import org.jboss.jandex.MethodParameterInfo;
 import org.jboss.jandex.PrimitiveType.Primitive;
 import org.jboss.jandex.Type;
 
@@ -138,6 +137,20 @@ public abstract class AbstractParameterProcessor {
     }
 
     private final Comparator<AnnotationInstance> frameworkParametersFirst = new FrameworkParamsFirstComparator();
+
+    /**
+     * Jandex sorts method arguments alphabetically. This comparator ensures they are sorted by their declaration order instead.
+     */
+    private static final Comparator<AnnotationInstance> COMPARATOR_METHOD_PARAMETER_ORDER = Comparator
+            .comparingInt(an -> {
+                AnnotationTarget target = an.target();
+                if (target.kind() != Kind.METHOD_PARAMETER) {
+                    // order method level parameters before method argument level parameters
+                    return -1;
+                }
+
+                return target.asMethodParameter().position();
+            });
 
     protected AbstractParameterProcessor(AnnotationScannerContext scannerContext,
             String contextPath,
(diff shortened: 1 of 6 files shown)


extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsConstants.java
```diff
@@ -125,12 +123,6 @@ public class JaxRsConstants {
     public static final Set<DotName> HTTP_METHODS = Collections
             .unmodifiableSet(methods);
 
-    private static final AnnotationValue[] EMPTY_VALUES = new AnnotationValue[0];
-    public static final Set<AnnotationInstance> HTTP_METHOD_INSTANCES = methods
-            .stream()
-            .map(name -> AnnotationInstance.create(name, null, EMPTY_VALUES))
-            .collect(Collectors.toUnmodifiableSet());
-
```
Postremus, line 133: Do you consider the fields in this class as public API?
I.e. can I just remove them?

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
```diff
@@ -1685,18 +1676,6 @@ boolean nonSyntheticParameterMethod(MethodInfo method, Collection<AnnotationInst
      */
     protected abstract boolean isResourceMethod(MethodInfo method);
 
-    /**
-     * Check for the existence relevant parameter annotations in the collection.
-     *
-     * @param annotations collection of annotations
-     * @return true if any of the annotations is a relevant parameter annotation.
-     */
-    protected boolean hasParameters(Collection<AnnotationInstance> annotations) {
```
Postremus, line 1694: Another "API" method that I removed.

Found no (non smallrye-openapi) implementation of AbstractParameterProcessor on github. Safe enough to remove?

extension-jaxrs/src/main/java/io/smallrye/openapi/jaxrs/JaxRsConstants.java
MikeEdgar: From my perspective, it's fine to remove these. I don't consider this package/module as part of the public API.

core/src/main/java/io/smallrye/openapi/runtime/scanner/spi/AbstractParameterProcessor.java
MikeEdgar: Same here, `AbstractParameterProcessor` isn't intended for use outside of the project.

MikeEdgar (approved):
Nice work, thanks @Postremus !

MikeEdgar:
@Postremus I'll go ahead and merge this unless you're still working on more changes.

Postremus:
@MikeEdgar
No thats it for now.

Thank you.

PR #2650, opened 2026-09-07
perf: reduce calls to FilteredIndexView.accepts

Postremus:
calls to FilteredIndexView.accepts are a bit expensive in my reproducer. This method does multiple checks on the class and package name. Most of them short circuit, since the config for these checks is not set by quarkus by default.

However, when reducing the invocation count of accepts from 42k -> 13k, there is still a noticeable improvement in document generation time.

Removed the possible double wrapping of the Index inside a FilterIndexView. Quarkus already gives us an FilterIndexView. Rewrote consecutive containsClass / getClass calls to a simple getClass with null check (where easily possible). 
Check the jdkIndex first in TypeUtil. This local index only contains some of the jdk classes, and is therefore quite small (avoids CompositeIndex scans / FilteredIndexView.accepts).

This saves about 30ms when doing a quarkus:build.

related to #2630


MikeEdgar (approved):
Looks good to me, thanks @Postremus . The Sonar warning looks bogus. Good to merge from your perspective?

Postremus:
Thank you for the review @MikeEdgar 
Yes this PR is ready.

PR #2667, opened 2026-09-19
perf: Existence check before annotation search

Postremus:
Most often, Annotations are defined in pars like List(javax.ws.rs.QueryParam, jakarta.ws.rs.QueryParam).
Applications generally either use javax, or the jakarta annotations. Mixing both is (afaik) also not supported by quarkus.

So smallrye openapi checks a lot of annotations unnecessarily.
In practice, 232k calls to getDeclaredAnnotations.

Jandex has a quick (ish) way to detect if an annotation is used anywhere in the application, IndexView.getAnnotations.

Based on that, we can first check if an annotation is even used anywhere, before searching the annotations on an AnnotationTarget for it.

In practice, this reduced the calls to the (inner) getDeclaredAnnotations from 232k to around 50k.

While at it, I defined constants for some more Lists of annotations.
And added non-varargs version of the hasAnnotation, getAnnotation, getAnnotationValue methods in Annotations (to avoid allocations for the varargs arrays).

However, IndexView.getAnnotations is performing allocations, and CompositeIndex.getAnnotations even more so. So we have to cache the annotation presence check.

(I also checked, an optimized, non allocating, containsAnnotation method in jandex helps, but still slower than our local cache. Most likely because multiple indexes annotation Hashmaps need to be checked).

This saves about 30ms for quarkus:build

related to #2630.


Postremus:
Setting this back to draft for now.

I might as well try to add the index.containsAnnotation method to jandex and use it.
In practice this will likely only safe a few allocations though; no noticable improvements from doing so.

PR #2668, opened 2026-09-21
perf: Optimize deepcopy

Postremus:
This PR intends to optimize the OpenAPI.deepcopy method, by avoiding reflection and reducing allocations.

I also added basic benchmark module (jmh) based on the on in jandex.

This saves about 3 ms for a quarkus:build.
related to #2630


testsuite/benchmarks/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.smallrye</groupId>
+        <artifactId>smallrye-open-api-parent</artifactId>
+        <version>4.4.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>smallrye-open-api-benchmarks</artifactId>
```
MikeEdgar, line 13: I'm thinking it might be better to put this new module under `testsuite/` and also make sure the Java packages are something like `io.smallrye.openapi.test.benchmarks`, etc. Eventually when we finally do JPMS modules that will avoid split packages.

model/src/main/java/io/smallrye/openapi/model/BaseModel.java
```diff
@@ -95,6 +106,8 @@ public C constructible() {
         return (C) this;
     }
 
+    public abstract Class<C> getConstructibleClass();
```
MikeEdgar, line 109: Good idea 👍

MikeEdgar (commented):
I haven't spotted anything major, just a few comments on the new benchmark module. Also, I think the CI won't be runnable without resolving the conflicts.

testsuite/benchmarks/pom.xml
Postremus: done

testsuite/benchmarks/src/main/java/io/smallrye/openapi/test/benchmarks/chart/ChartGenerator.java
```diff
@@ -0,0 +1,105 @@
...
+import org.knowm.xchart.CategoryChart;
+import org.knowm.xchart.CategoryChartBuilder;
+import org.openjdk.jmh.annotations.Mode;
+
+import com.google.gson.Gson;
+
+import io.smallrye.openapi.test.benchmarks.json.BenchmarkDto;
+
+public class ChartGenerator {
+    private static final Gson GSON = new Gson();
```
MikeEdgar, line 27: I meant to comment on it yesterday but passed it by. Does this have to be Gson or could it be Jackson 2 or 3 instead? Just to keep dependency bloat to a minimum.

testsuite/benchmarks/run-benchmarks.sh
MikeEdgar, line 1: It might be nice to see this running in a new GHA workflow (single Java version is probably fine). It could attach the chart to the result build for review. What do you think? Maybe a separate PR for that part.

MikeEdgar (commented):
Looking good, just one more question about the JSON binding that I forgot to bring up yesterday.

testsuite/benchmarks/src/main/java/io/smallrye/openapi/test/benchmarks/chart/ChartGenerator.java
Postremus: done, switched to jackson.

testsuite/benchmarks/run-benchmarks.sh
Postremus: Lets do this as follow up.

Problem is, that github actions is, obviously, not a lab setup. Meaning, that results will vary wildly between runs.
So we should note something like: "Charts are for review purposes only, and don't indicate real world performance.".

Then I believe that it would be fine to look at the charts as part of the review.

We could compare the branch head to the main commit the branch was created from.

MikeEdgar: Right, it wouldn't need to be anything precise. It should mainly be a way to catch drastic changes in performance or the introduction of some bug in the benchmark code itself. A follow-up sometime would be great, thank you.

MikeEdgar (changes requested):
Hopefully this is the last thing... Running this locally with JDK 25, I need to add the JMH annotation processor to the Maven compiler to get this working. Otherwise, there aren't any benchmarks found. We may as well just add it here and it should work for any recent JDK version then.

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <configuration>
        <annotationProcessorPaths>
            <path>
                <groupId>org.openjdk.jmh</groupId>
                <artifactId>jmh-generator-annprocess</artifactId>
                <version>${version.jmh}</version>
            </path>
        </annotationProcessorPaths>
    </configuration>
</plugin>
```

Postremus:
Good catch @MikeEdgar , I only tested with jdk 21 locally.

Postremus:
done. I also had to add test-compile to the run-benchmarks.sh script.
The jmh plugin ignores annotation processor while compiling sources.

Postremus:
I am going to disable sonar for the benchmarks module.
(if I can figure out how). Don't see a point in requiring tests for it.

MikeEdgar:
That's fine - you can add it to `<sonar.coverage.exclusions>` in the root POM.

MikeEdgar (approved):
LGTM, the Sonar warnings about the lambdas seem fine to me. Reworking that would probably make the assertions less readable.

